PR #23, opened 2019-06-18
Add ILGPU.Tests

mikhail-khalizev:
Add ILGPU.Tests project.


m4rs-mt:
@mikhail-khalizev I will publish all internal tests in the next days in a separate repository called `ILGPU.Tests`. I suggest that we move your new tests to this repository and connect both via a git submodule. What do you think?

m4rs-mt:
May be we can port your test cases to the existing test framework at [https://github.com/m4rs-mt/ILGPU.Tests](url)?

mikhail-khalizev:
Hi. That's good. Only one question, why you want to use separate repository? Usually library and tests for it placed in same repository.

mikhail-khalizev:
> May be we can port your test cases to the existing test framework at [https://github.com/m4rs-mt/ILGPU.Tests](url)?

Yes,  I will look.

m4rs-mt:
Yes, you're absolutely right. The reason for this is that I have received many requests to create a separate repository for the test cases to make it easier (for them?) to distinguish between actual compiler commits and test updates. But maybe we want to change that in the future :)

m4rs-mt:
I published all internal test cases (merged into the main repository).

mikhail-khalizev:
Add #24.

PR #25, opened 2019-07-03
Fixed issue with PTX compatibility on new RTX cards

MoFtZ:
- PTX 6.0 only supports up to SM_70

### DIFF
--- a/Src/ILGPU/Backends/PTX/PTXArchitecture.cs
+++ b/Src/ILGPU/Backends/PTX/PTXArchitecture.cs
@@ -76,6 +76,11 @@ namespace ILGPU.Backends
         SM_70,
 
         /// <summary>
+        /// The 7.2 architecture.
+        /// </summary>
+        SM_72,
+
+        /// <summary>
         /// The 7.5 architecture.
         /// </summary>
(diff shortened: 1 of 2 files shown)


m4rs-mt:
@MoFtZ you are absolute right. PTX ISA 6.0 does not support SM_72 or higher. I think upgrading the PTX version in the backend is a better alternative than mapping a modern architecture to an older one. What do you think?

MoFtZ:
@m4rs-mt I was afraid you'd say that =)

yes, that's a reasonable request. I've updated to PTX v6.4 - it really only needed v6.3 to get SM_75, so happy to downgrade to v6.3 for broader compatibility with drivers.

From reading the PTX release notes, there were some deprecated/removed instructions, but it should not affect any existing code.

m4rs-mt:
I also checked all deprecated instructions for possible issues.
However, I guess this shouldn't affect existing code, as you mentioned earlier.

I guess v6.4 should be fine as we want to add support for additional PTX features soon =)

RickHH:
It seems that the PTX compatibility issue appears on the newest RTX cards(RTX 3080)?

PR #33, opened 2019-09-10
Added breaking unit test when using if statement with multiple clauses

MoFtZ:
hi @m4rs-mt, just ran into this code generation issue, and was able to create a simplified unit test.

Not really sure how to fix at the moment, but will dig around.

### DIFF
--- a/Src/ILGPU.Tests/BasicIfs.cs
+++ b/Src/ILGPU.Tests/BasicIfs.cs
@@ -240,6 +240,33 @@ namespace ILGPU.Tests
                 }
             }
         }
+
+        internal static void IfAndOrKernel(
+            Index index,
+            ArrayView<int> data)
+        {
+            int value;
+
+            if ((index.X == 0 || index.X == 1) && index.X <= 2)
+                value = 42;
+            else
+                value = 0;
+
+            data[index] = value;
+        }
+
+        [Fact]
+        [KernelMethod(nameof(IfAndOrKernel))]
+        public void IfAndOr()
+        {
+            using (var buffer = Accelerator.Allocate<int>(Length))
+            {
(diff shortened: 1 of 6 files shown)


MoFtZ:
Phew.. I'm exhausted.
On the plus side, I worked out the problem.

In IR\Analyses\CFG.cs:450, we add successors/predecessors to each of the blocks. However, because the code iterates over a dictionary, the order of adding successors is non-deterministic.

Later, in IR\Construction\IRRebuilder.cs:77, we copy the blocks in Scope, again in an unordered way.

Finally, in Backends\PTX\PTXCodeGenerator.cs:406, when generating the kernel code, we perform a mapping of Phi arguments from predecessor nodes. However, by this stage, the order of predecessors/successors has changed since the original Phi node was created.

The crash occurred because the Phi arguments had swapped (because the order of adding successors/predecessors was different).

As a fix, I have ordered the Blocks by ID.
Can you see any issues with this fix? or do you have a suggestion for a better solution?

MoFtZ:
Damn, forgot to check the CUDA unit tests - they failed because something else must be unordered.

Will have to try again.

MoFtZ:
OK, changed PhiNode to be immutable again. Hopefully I replaced the PhiValue correctly.

m4rs-mt:
I guess it's not enough to revert the commit mentioned. However, it was intended to remove explicit tracking of predecessors to simplify upcoming transformations.

In the case of ILGPU, it may not be necessary to use different dialects for SPIR-V and CUDA.
A number of additional nodes (such as a `Loop` node) may be advantageous, as they allow high-level constructs to be expressed within the IR.

Block arguments will definitely simplify many transformations and make the integration of lambda functions much easier, I guess.

PR #35, opened 2019-10-10
Updated CudaMemoryBuffer to support MemSetToZero using alternate streams

MoFtZ:
hi @m4rs-mt , was doing some profiling and noticed that MemSetToZero would always use the default accelerator stream.


m4rs-mt:
Yeah, you're right. I can confirm that the `memset` operation always used the default stream.

Issue #7, opened 2018-11-22
Why .NET Standard dropped?


kant2002:
As part of 0.4.0-beta release, I see that .NET Standard was dropped, and replaced to .NET Core App 2.0 which is more limiting in my opinion. What are the rationale why .NET Core App is strictly needed for this library?

m4rs-mt:
The reason ILGPU does not support .NET Standard 2.0 is because ILGPU relies on the System.Reflection.Emit.ILGeneration package, which is currently not (correctly) supported on all .NET Standard compatible platforms. For more information, visit [https://github.com/dotnet/corefx/issues/29365](https://github.com/dotnet/corefx/issues/29365) and [https://github.com/dotnet/corefx/issues/30654](https://github.com/dotnet/corefx/issues/30654)

There are basically two possibilities:
1) Create an ILGPU package that pretends to be .NET Standard 2.0 compatible, but one of its dependencies is not.
2) Wait until the dependency is fully compatible before claiming full compatibility with the .NET standard.

kant2002:
I want to ask one more question. I just do not aware about ability to target `netcoreapp2.0` libraries from `netcoreapp2.1`. If this is possible, I personally would not be affected in any way, if not, then I would ask for targeting of multiple frameworks. Even if it is targeted, maybe it would be practical to include into samples, that this library could be used from `netcoreap2.1` project.

m4rs-mt:
I don't see any problems using `.netcoreapp2.0` libs from `netcoreapp2.1` apps or libraries. This should work without any further problems (in theory).

scherenhaenden:
Hello. 
I got almost the same issue but the other way around. I get from nuget only the .netstandard version, not the .netframework version

kant2002:
@scherenhaenden 0.3.0 version was `netstandard2.0` which allow it to run under `net47`
0.4.0-beta would be `netcoreapp2.0` (which is different from netstandard) and that make it run only on `netcoreapp2.0`, `netcoreapp2.1` and `netcoreapp3.0` (when it will be published). So your application should target at least `netcoreapp2.0` now.

Given the MS stance on full .NET, it is reasonable to move all applications on .NET Core instead of creating them on .NET framework.

samy2:
@scherenhaenden , @kant2002 
[...]

PR #37, opened 2019-10-15
Added support for .Net Standard 2.1

MoFtZ:
With the recent release of VS2019 v16.3, .NET Core 3.0 and .NET Standard 2.1 are now available.

In particular, .NET Standard 2.1 supports the the Reflection.Emit functionality required by #7 
https://devblogs.microsoft.com/dotnet/announcing-net-standard-2-1/

Unfortunately, .NET Standard 2.1 libraries cannot be used by either .NET Core 2.x nor .NET Framework.

### DIFF
--- a/.nuget/ILGPU.nuspec
+++ b/.nuget/ILGPU.nuspec
@@ -21,7 +21,6 @@
     <tags>ilgpu gpu msil il cil cpu ptx nvidia amd amp opencl cuda compiler jit</tags>
     <dependencies>
       <dependency id="System.Collections.Immutable" version="1.5.0" />
-      <dependency id="System.Reflection.Emit.ILGeneration" version="4.3.0" />
       <dependency id="System.Reflection.Metadata" version="1.6.0" />
       <dependency id="System.Runtime.CompilerServices.Unsafe" version="4.5.2" />
     </dependencies>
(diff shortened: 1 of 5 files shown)


m4rs-mt:
@MoFtZ nice work 👍 I was wondering whether we have to udpate the `dependencies` section in the `.nuspec` file as well?
```
<dependencies>
  <dependency id="System.Collections.Immutable" version="1.5.0" />
  <dependency id="System.Reflection.Emit.ILGeneration" version="4.3.0" />
  <dependency id="System.Reflection.Metadata" version="1.6.0" />
  <dependency id="System.Runtime.CompilerServices.Unsafe" version="4.5.2" />
</dependencies>
```

MoFtZ:
To be honest, I'm not sure.

Each of those packages have an updated version in Nuget, so sure, we can update all of them.

However, the blog post about .Net Standard 2.1 says that they could not implement reflection emit using nuget packages alone. It sounds like the .Net Standard 2.1 "framework" is the one that provides the required functionality, similar to the full .Net 4.7 framework. In which case, it is not necessary to update the nuget packages.

What do you think?

m4rs-mt:
Right, you're right 👍 This should still work on the "older" .Net framework 4.7 and .Net Core 2.0 compatible frameworks.

Issue #39, opened 2019-10-18
A PTX JIT compilation failed


AlexAba:
I have an exception: "A PTX JIT compilation failed" at 

`var myKernel = accelerator.LoadAutoGroupedStreamKernel<Index, ArrayView<int>>(MyKernel);`.

I have windows 10, nVidia GeForce 940MX, Visual Studio 2017

            using (var context = new Context())
            {
                using (var accelerator = new CudaAccelerator(context))
                {
                    var myKernel = accelerator.LoadAutoGroupedStreamKernel<Index, ArrayView<int>>(MyKernel);

                    using (var buffer2 = accelerator.Allocate<int>(1000))
                    using (var buffer = accelerator.Allocate<int>(1000))
                    {

                        s1.Start();

                        myKernel(buffer.Length, buffer.View);

                        accelerator.Synchronize();
                        s1.Stop();

                        var data = buffer.GetAsArray();

                        for (var i = 0; i < 100; i++)
                        {
                            Console.Write(" " + data[i]);
                        }
                    }
                }
            }

m4rs-mt:
@AlexAba this sounds like a driver issue. Have you tried updating your graphics drivers (as also proposed in #36)?

AlexAba:
I have update my drivers to last version. And when I use 0.5.1 version it's all ok.

m4rs-mt:
Hmm.... Can you provide the kernel code in such a way that we can investigate the problem more closely?

AlexAba:
`const int n = 10;`
`static void MyKernel(Index index, ArrayView<int> dataView)`
`        {`
`           int z = 0;`
`           for (int t = 0; t < n; t++)`
`            {`
`                z += 5;`
`           }`
`            dataView[index] = z;`
`        }`

Issue #36, opened 2019-10-10
"SimpleKernel" example throws exception


AlexBatsev:
I've tried to run SimpleKernel example and has got 
_ILGPU.Runtime.Cuda.CudaException: 'a PTX JIT compilation failed'_

Inside CudaKernel constructor, the function `CudaAPI.Current.LoadModule(...)` returned
`CudaError cudaStatus = CUDA_ERROR_INVALID_PTX;`
```
string errorLog = "
_ptxas application ptx input, line 5; fatal   : Unsupported .version 6.0; current version is '5.0'
ptxas fatal   : Ptx assembly aborted due to errors_
"

```
Indeed, on line 5 one can see "_.version 6.0_":

`kernel.PTXAssembly =` 

	//
	// Generated by ILGPU v0.5.1
	//
	
	.version 6.0
	.target sm_50
	.address_size 64

	.visible .entry ILGPUKernel(
	.param .align 4 .b8 _index_1318[4],
	.param .align 8 .b8 _dataView_1319[16],
	.param .b32 _constant_1320
	)
	{
		.reg .pred	%p<2>;
		.reg .b32	%r<8>;
		.reg .b64	%rd<4>;
	
		ld.param.b32	%r2, [_index_1318];
		ld.param.b64	%rd1, [_dataView_1319];
		ld.param.b32	%r3, [_dataView_1319+8];
		ld.param.b32	%r4, [_constant_1320];
		mov.b32	%r5, %ctaid.x;
		mov.b32	%r6, %ntid.x;
		mov.b32	%r7, %tid.x;
		mad.lo.s32	%r1, %r5, %r6, %r7;
		setp.ge.s32	%p1, %r1, %r2;
		@%p1 ret;
	
		L\_13170:
		mul.wide.u32	%rd3, %r1, 4;
		add.u64	%rd2, %rd1, %rd3;
		add.s32	%r7, %r1, %r4;
		st.b32	[%rd2], %r7;
		ret;
	}
If it is important, my graphic card is  Quadro M2000M.

Is this a bug or do I misunderstand something?
Thank you.

MoFtZ:
hi @AlexBatsev, have you tried updating your CUDA drivers? This appears to be a similar problem to #29.

AlexBatsev:
After I've updated drivers the problem's gone.
After updating ILGPU to 0.6.0, the .version changed to "6.4".
[...]

PR #40, opened 2019-10-21
Added support for flexible PTX ISA detection

### LINKED ISSUE #29: PTX JIT compilation failed (Ver. 0.6.0)
Hi @m4rs-mt,

this will return an PTX JIT compilation error on Linux but works on Windows.
> _kernelRGB888PlanarToBGR888xPacked = _accelerator.LoadAutoGroupedStreamKernel<Index, int, int, int, ArrayView<byte>, ArrayView<byte>>(RGB888PlanarToBGR888xPacked);

Kernel:
```
private static void RGB888PlanarToBGR888xPacked(Index pixelIndex, int width, int height, int bpp, ArrayView<byte> source, ArrayView<byte> destination)
        {
            var offset = width * height;
            var r = source[pixelIndex];
            var g = source[pixelIndex + offset];
            var b = source[pixelIndex + offset * 2];

            destination[pixelIndex * bpp] = b;
            destination[pixelIndex * bpp + 1] = g;
            destination[pixelIndex * bpp + 2] = r;
        }
```

Only in version 0.6.0. The same code works with 0.5.1.

MoFtZ:
hi @m4rs-mt, here are some changes to try to fix #29, #36 and #39.

This should give a nicer error message, prompting users to update their CUDA drivers.

Not sure if the flexible PTX ISA detection is the best approach - let me know what you think.

### DIFF
--- a/Src/ILGPU/Backends/PTX/PTXArchitecture.cs
+++ b/Src/ILGPU/Backends/PTX/PTXArchitecture.cs
@@ -3,7 +3,7 @@
 //                     Copyright (c) 2016-2019 Marcel Koester
 //                                www.ilgpu.net
 //
-// File: PTXBackend.cs
+// File: PTXArchitecture.cs
 //
 // This file is part of ILGPU and is distributed under the University of
 // Illinois Open Source License. See LICENSE.txt for details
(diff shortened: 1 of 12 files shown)


m4rs-mt:
I like the idea of an automated PTX-ISA detection that is fully implemented by ILGPU 🥇

Issue #41, opened 2019-10-22
Rework static analysis


m4rs-mt:
As we have just switched to VS2019 (312280c90a65d4334a820bc9780663ea7418c6d6) we should definitely rework the internal static-program analysis pipeline. The current ILGPU project relies on an (now) obsolete version of the well known Post-build Code Analysis. We should definitely upgrade the analysis to a newer version.

m4rs-mt:
After updating to the latest Visual Studio version, the FXCop analyzers seem to stop running after a successful build. Also, I usually don't get any warnings.
@MoFtZ Are there known solutions to this problem? I couldn't find any solutions to this issue on the Web so far.

MoFtZ:
hi @m4rs-mt, I've noticed that the FXCop analyzers appear to run in the background, and in some cases only when a file is open.

I personally turn on "Enable full solution analysis", though this may have a performance impact for very large projects.
![image](https://user-images.githubusercontent.com/27792737/68083375-24e68e00-fe7c-11e9-8035-d7b2ab05a213.png)

In rare situations when there are errors/warnings/messages that should not be there, I have found a "Rebuild All" or restarting Visual Studio itself, will fix the problem.

m4rs-mt:
@MoFtZ Thank you very much. The activation of the "full solution analysis" seems to solve the problem. Rebuilding and restarting Visual Studio seems to work from time to time. I also noticed the background tasks, but unfortunately the errors/warnings/messages don't appear after some time.

PR #42, opened 2019-10-27
Static analysis

MoFtZ:
hi @m4rs-mt, I have started working on updating the static analysis as per #41.

I had some questions for you regarding a few of the issues that came up.

- [x] Disposable-related warnings

  CA1001: Types that own disposable fields should be disposable
  CA2000: Dispose objects before losing scope
  CA2213: Disposable fields should be disposed

  It looks like a number of these have been ignored/suppressed in the past. Should I continue to suppress these warnings?

- [x] CA1060: Move pinvokes to native methods class

  This was part of your recent changes to support OpenCL. I was not planning to make any changes in case they caused conflicts with your pending commits.

- [x] [CA2225: Operator overloads have named alternates](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca2225?view=vs-2019)

  It looks like it is recommended that operators have a matching named method. So:
  ```
  public static implicit operator ArrayView<T, Index>(ArrayView<T> view)
  {
      return new ArrayView<T, Index>(view, view.Length);
  }
  ```
  would have a new method:
  ```
  public ArrayView<T, Index> ToArrayView();
  ```
  What are your thoughts on this recommendation? Implement or Suppress?


MoFtZ:
There are also a number of code styling issues, which are subjective. Do you prefer to use the new style, or keep the existing style?

- [ ] IDE0019: Use pattern matching
    ```
    switch (value)
    {
        case 0:
            value = 7;
            break;
        case 1:
            value = 42;
            break;
        case 2:
            value = 1337;
            break;
        default:
            value = -1;
            break;
    }
    ```
   can be changed to:
    ```
    value = value switch
    {
        0 => 7,
        1 => 42,
        2 => 1337,
        _ => -1,
    };
    ```
- [ ] IDE0034: 'default' expression can be simplified

    ```
   Current = default(LocalVariable);
    ```
   can be changed to:
    ```
   Current = default;
    ```
- [ ] IDE0063: 'using' statement can be simplified
 
    ```
    using (var buffer = Accelerator.Allocate<int>(1))
    {
        Execute<Index>(specializedMethod, buffer.Length, buffer.View);

        var size = Marshal.SizeOf(type);
        var expected = new int[] { size };
        Verify(buffer, expected);
    }
    ```
   can be changed to:
    ```
    using var buffer = Accelerator.Allocate<int>(1);
    Execute<Index>(specializedMethod, buffer.Length, buffer.View);

    var size = Marshal.SizeOf(type);
    var expected = new int[] { size };
    Verify(buffer, expected);
    ```

m4rs-mt:
@MoFtZ First of all: thank you very much for diving into this topic :)

1. Regarding the `IDisposable` question
Some warnings have been suppressed in the past because static analysis cannot check whether the `DisposeBase.Dispose(ref object)` method actually disposes of the object and sets the referenced field to `null`. Should we simply change this, since it is no longer necessary to set fields to zero, since multiple calls to the `Dispose` method are "intercepted" by the `DisposeBase` anyway. What do you think?
2. I already expected something like this to happen. I ignore some of these warnings in the past to simplify the integration of the `CudaAPI` method calls due to native imports from different Cuda libraries. I am going to update the `OpenCL` part of the project after you have updated the static analysis part.
3. I guess it makes sense to implement them in most of the cases. However, since we usually target the "core" .NET world it might not be required in all cases. What do you think?

m4rs-mt:
Regarding the guidelines on the "new coding style": I guess we should follow them. For example, the new `using` statements and simplified `default` keywords make it much easier to understand and structure the code. What do you think?

I can also apply the new coding styles to the source code after you've done the updated analysis. Of course you can also apply the changes to all source files, but I'd like to ask you to omit the `IR` and `Backend` packages as I'm working on them anyway.

MoFtZ:
@m4rs-mt, it sounds like the implementation of `IDisposable` and `DisposeBase` should be reviewed. For now, I'm going to suppress the warnings - it can be handled in separate enhancement.

For `CA2225: Operator overloads have named alternates`, I'm going to suppress this too. Again, it can be handled in a separate enhancement, and as long as there are no issues with the current API, there is no urgency to implement the named alternatives.

I have also updated the source code to remove all warnings/messages except for in `IR` and `Backend`.

PR #46, opened 2019-11-06
Rework implementation of dispose pattern

MoFtZ:
hi @m4rs-mt. OK, I've gone through the implementation of `IDisposable` in ILGPU and made some changes to remove all the warnings.

A number of the changes will break the existing ILGPU v0.6.0 API.

- [ ] Removed DisposeBase
  The `DisposeBase` utility class was removed because in the `DisposeBase:Dispose(ref object)` static method was not understood by the FXCop analyzers. In addition, there were only a handful of places that needed the `object?.Dispose()` and `object = null` functionality.

  `IDisposable.Dispose` should be safe to call multiple times. Added an `alreadyDisposed` flag to many classes. This did not necessarily have to use `Interlocked.CompareExchange` everywhere, but I could not be sure which needed to be thread-safe. Also ensured that native handles are reset to `IntPtr.Zero`.

- [ ] Removed unnecessary IDisposable interface from AcceleratorId
  Never disposed anyway

- [x] Removed unnecessary IDisposable interface from AssemblyDebugInformation
  Do not need to hold onto the MetadataReaderProvider

- [ ] Added finalizer only to classes with unmanaged resources
  Finalizers are expensive, and should only be implemented by classes that require them.

- [x] Use auto-implemented property (instead of private member variables)
  `DisposeBase.Dispose(ref object)` required the private member variable to be writable. After removing `DisposeBase`, some member variables did not need to be changed (only set in the constructor).

### DIFF
--- a/Src/ILGPU.Tests/Generic/TestBase.cs
+++ b/Src/ILGPU.Tests/Generic/TestBase.cs
@@ -205,6 +205,7 @@ namespace ILGPU.Tests
                 Accelerator.Dispose();
                 Context.Dispose();
             }
+            base.Dispose(disposing);
         }
 
         #endregion
--- a/Src/ILGPU.Tests/Generic/TestBase.cs
+++ b/Src/ILGPU.Tests/Generic/TestBase.cs
@@ -32,9 +32,6 @@ namespace ILGPU.Tests
 
         #endregion
 
-        private Context context;
-        private Accelerator accelerator;
-
         /// <summary>
         /// Constructs a new test base.
         /// </summary>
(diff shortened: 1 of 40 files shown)


m4rs-mt:
@MoFtZ thanks a lot for diving into this topic.

1.I'm not sure if we should really remove the `DisposeBase` class, as in many cases this leads to code duplication. However, I totally agree that we should definitely remove the `DisposeBase::Dispose(ref object)` method. What do you think?
2. We should keep the `IDisposable` functionality of the `AcceleratorId' objects, since the `OpenCL' version assigns some native handles that should definitely be released by a user/automatically freed by the finalizer.
3. 👍 
4. This makes sense and avoids unnecessary overhead; which in turn means, that the `DisposeBase` class should be removed.
5. 👍

MoFtZ:
OK, I've restored `DisposeBase`. 

It's OK for `DisposeBase` to implement the finalizer - `GC.SuppressFinalize(this)` will deal with the normal use case, and the penalty for finalizers will not occur. However, the finalizer penalty will affect users of the ILGPU library that do not properly dispose of ILGPU objects.

I have also restored `IDisposable` on `AcceleratorId` objects - I missed the `Dispose` in `CLAcceleratorId`.

- [ ] What do you think about changing the method `DisposeBase:Dispose(bool)` from `abstract` to `virtual`? This will mean that all implementations of `Dispose` can consistently call `base.Dispose(disposing)`.

m4rs-mt:
Makes perfect sense to me. This standardizes the code and avoids programming errors.

MoFtZ:
Changed DisposeBase.Dispose(bool) from abstract to virtual.

PR #55, opened 2019-12-23
Fix opencl code generator for float values that are assign integers v…

12345swordy:
…alues.


Src/ILGPU/Backends/OpenCL/CLCodeGenerator.Emitter.cs
```diff
@@ -395,7 +395,10 @@ public void AppendConstant(float value)
             {
                 AppendArgument();
                 stringBuilder.Append(value);
-                stringBuilder.Append('f');
+                if (value % 1 == 0)
```
m4rs-mt: Please change the constants `1` and `0` to `1.0f` and `0.0f`. After that this PR should be ready to go 👍

PR #57, opened 2019-12-29
Increase the LogSize

12345swordy:
I have encounter open-cl bugs with AMD when running the test suits. I need logSize to be way bigger then usual, as it crashes when the LogSize is not big enough.

### DIFF
--- a/Src/ILGPU/Runtime/OpenCL/API/CLAPI.cs
+++ b/Src/ILGPU/Runtime/OpenCL/API/CLAPI.cs
@@ -463,7 +463,7 @@ namespace ILGPU.Runtime.OpenCL.API
             IntPtr device,
             out string buildLog)
         {
-            const int LogSize = 4096;
+            const int LogSize = 32_000;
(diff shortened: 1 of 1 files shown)


Src/ILGPU/Runtime/OpenCL/API/CLAPI.cs
```diff
@@ -463,7 +463,7 @@ public static CLError CreateContext(IntPtr device, out IntPtr context)
             IntPtr device,
             out string buildLog)
         {
-            const int LogSize = 4096;
+            const int LogSize = 32_000;
```
m4rs-mt: Since you have increased the log size to about 32Kb, it might be a good idea to allocate it on the heap instead. Can you adjust the following line to use a heap-allocated array?
After that, this PR should be ready.

m4rs-mt:
Unfortunately, I made a mistake by merging this PR. The code does not compile any more since the heap-allocated version does not provide a pointer to the other functions any more.

PR #67, opened 2020-02-09
Fix the error by having manual casting the value to genetic when pass…

12345swordy:
…ing to the parameter

### DIFF
--- a/Src/ILGPU/Backends/OpenCL/CLCodeGenerator.Values.cs
+++ b/Src/ILGPU/Backends/OpenCL/CLCodeGenerator.Values.cs
@@ -296,7 +296,7 @@ namespace ILGPU.Backends.OpenCL
                 statement.BeginArguments();
                 statement.AppendAtomicCast(atomicCAS.ArithmeticBasicValueType);
                 statement.AppendArgument(target);
-                statement.AppendArgumentAddressWithCast(value, "__generic " + CLTypeGenerator.GetBasicValueType(atomicCAS.ArithmeticBasicValueType) + " *");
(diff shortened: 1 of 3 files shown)


12345swordy:
This fix only works for debug mode btw.

Src/ILGPU/Backends/OpenCL/CLCodeGenerator.Values.cs
```diff
@@ -296,7 +296,7 @@ public void Visit(AtomicCAS atomicCAS)
                 statement.BeginArguments();
                 statement.AppendAtomicCast(atomicCAS.ArithmeticBasicValueType);
                 statement.AppendArgument(target);
-                statement.AppendArgumentAddress(value);
+                statement.AppendArgumentAddressWithCast(value, "__generic " + CLTypeGenerator.GetBasicValueType(atomicCAS.ArithmeticBasicValueType) + " *");
```
m4rs-mt: nit-picking: I think it would be a bit nicer to shorten this line, as it has significantly more than 100 characters and is not a comment line. Also, it would be great if we could use the `CLInstructions.DereferenceOperation` instead of the hard-coded `*`. How about creating a more generic function that generates this cast expression within the `CLInstructions` class?

12345swordy:
> Furthermore, I am wondering why this fix does not solve the problem in release mode?
Because it currently gives me this error message:
``` error: initializing 'int *' with an expression of type '__generic int *' changes address space of pointer
private int* var29 = (int*) var12; 
```



I don't know where to begin with that that.

m4rs-mt:
After diving deeper into this issue it turns out that all pointer types should be annotated with the proper "__generic" address space. However, I don't known whether all drivers will accept this code.

PR #95, opened 2020-05-06
Fixed unordered floating point comparisons

### LINKED ISSUE #93: Cuda accelerator incorrectly handles NaN comparison
When using a simple kernel, and supplying `NaN` as an input value, the Cuda accelerator incorrectly compares `NaN` and returns `2.0` rather than `4.0`.

NB: The comparison `value == 0.0` can be changed to another number with the same/incorrect result on Cuda.
```
internal static void TestKernel(Index1 index, ArrayView<double> input, ArrayView<double> output)
{
    var value = input[index];
    if (value == 0.0)
        output[index] = 2.0;
    else
        output[index] = 4.0;
}
```

Full sample:
[ZeroOneNaN.zip](https://github.com/m4rs-mt/ILGPU/files/4572667/ZeroOneNaN.zip)

MoFtZ:
Fixes #93

### DIFF
--- a/Src/ILGPU.Tests/Generic/ConfigurationBase.tt
+++ b/Src/ILGPU.Tests/Generic/ConfigurationBase.tt
@@ -36,7 +36,8 @@ public static IEnumerable<string> GetFloatRanges(
     bool maxRanges,
     bool useZero = true,
     bool useInfinity = true,
-    bool useMultiplesOfPi = true)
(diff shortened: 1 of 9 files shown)


m4rs-mt:
@MoFtZ Just in time 🔢  I just wanted to solve the problem in a similar way.

PR #96, opened 2020-05-06
Fixed OpenCL unary logical not operation on boolean

MoFtZ:
`ILGPU.Tests.OpenCL.CLBasicIfs.IfAndOr` fails because bitwise not `~` does not always work on a boolean - depends on the internal representation. Changed to logical not `!`.

### DIFF
--- a/Src/ILGPU/Backends/OpenCL/CLInstructions.Data.cs
+++ b/Src/ILGPU/Backends/OpenCL/CLInstructions.Data.cs
@@ -238,51 +238,92 @@ namespace ILGPU.Backends.OpenCL
             ">="
         };
 
+        /// <summary>
+        /// Identifies the permitted value types for unary arithmetic operations.
+        /// </summary>
+        private enum CLUnaryCategory
+        {
+            /// <summary>
+            /// Unary operation on booleans.
+            /// </summary>
+            Boolean,
+
+            /// <summary>
+            /// Unary operation on signed or unsigned integers.
+            /// </summary>
+            Int,
+
+            /// <summary>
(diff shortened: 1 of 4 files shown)


Src/ILGPU/Backends/OpenCL/CLInstructions.Data.cs
```diff
@@ -238,51 +238,59 @@ partial class CLInstructions
             ">="
         };
 
+        private enum CLUnaryCategory
```
m4rs-mt, line 241: Can you add a small comment here why we need this enumeration? (It is absolutely clear but a little comment would be helpful in this case, I guess)

Src/ILGPU/Backends/OpenCL/CLInstructions.cs
```diff
@@ -54,19 +55,22 @@ public static partial class CLInstructions
...
         public static string GetArithmeticOperation(
             UnaryArithmeticKind kind,
-            bool isFloat,
+            ArithmeticBasicValueType basicValueType,
             out bool isFunction)
         {
-            if (!UnaryArithmeticOperations.TryGetValue(
-                (kind, isFloat),
-                out var operation))
+            var category = basicValueType.IsFloat()
```
m4rs-mt, line 68: Can we create a mapping that immediately maps `ArtihmeticBasicValueType` values to their appropriate `CLUnaryCategory` values? I guess this makes it easier to spot where and how those value instances are being created 🔍

MoFtZ:
PR updated

m4rs-mt:
Awesome work 👍

Issue #90, opened 2020-05-02
Group reduction hangs for CPU accelerator


AlexBatsev:
I was writing a group reduction algorithm and noticed that it hangs for the CPU accelerator while all others (Cuda, OpenCL) execute it correctly. Source file is attached [Program.zip](https://github.com/m4rs-mt/ILGPU/files/4568504/Program.zip). Minimal kernel on which it is reproduced is as follows:

        private static void GroupReductionHanging(ArrayView<float> dataView)
        {
            var threadNo = Group.IdxX;
            InitSharedMemory();
            Group.Barrier();
            for (var nLivingThreads = Group.DimX >> 1; nLivingThreads > 0; nLivingThreads >>= 1)
            {
                if (threadNo >= nLivingThreads)
                {
                    WriteToSharedMemory();
                    return;
                }

                Group.Barrier();
                ReadFromSharedMemory();
            }
        }
        }
All the functions in the kernel like `InitSharedMemory()` do not matter and can be empty for simplicity. 
On each `for` loop iteration the number of active threads reduces twice, and only half of them reach `Group.Barrier()`, and I was suspecting that waiting inside `Barrier` for inactive ones (which already have hit `return`) causes the problem. That is not a case, though, as the kernel does not hang inside the `Barrier()`, but upon return from the kernel, when the last surviving thread finishes. It hangs in the 
`System.Threading.ManualResetEventSlim.Wait(int millisecondsTimeout, CancellationToken cancellationToken)`.

Still, I’ve verified an idea of fixing the problem by making all threads, including inactive ones, to hit the `Barrier()`. And it worked! The “fixed” kernel which is equivalent to original one is given by `GroupReductionOk` in the attached file.

I believe this is a bug.

m4rs-mt:
@AlexBatsev Thanks for your issue report. Please note that this is the intended behavior, as it is always considered to be dangerous to use block synchronization in divergent control flow in general. It works on certain devices but you should definitely be very careful when using these instructions. A look at the latest [Cuda documentation](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#synchronization-functions) says the following:
> __syncthreads is allowed in conditional code but only if the conditional evaluates identically across the entire thread block, otherwise the code execution is likely to hang or produce unintended side effects.

However, I know that this code works on many devices out there on the one hand. I have seen other devices which simply hang or even crash, on the other hand. We might want to relax this constraint for the CPU accelerator in the future.

[...]

PR #106, opened 2020-05-11
Fixed Group.Barrier compatibility with CPU accelerator when some threads terminate early

MoFtZ:
Fixes https://github.com/m4rs-mt/ILGPU/issues/90


m4rs-mt:
I double checked all test cases and it turned out that the new test `GroupDivergentControlFlow` for OpenCL fails. @MoFtZ Can you take a look at this?

PR #112, opened 2020-05-13
Fixed inlining OpenCL constants on if statements and function arguments

MoFtZ:
Fixed regression from commit 0609ea8b0a7be3dae746992c6c1e5dcaad54662e

`ILGPU.Tests.OpenCL.CLArrayViews_Debug.ArrayViewGetSubVariableView` kernel compilation failed because it generated code like `if (var20)` - `var20` was a constant, and never declared.

`ILGPU.Tests.OpenCL.CLGroupOperations_Debug.GroupDivergentControlFlow` kernel compilation failed because it generated code like `atomic_fetch_add ((atomic_int*)var12, , 1);` - appending the constant added a second/invalid comma.

PR #113, opened 2020-05-13
Fixed initialisation of GroupDivergentControlFlow unit test

MoFtZ:
Related to #106

NOTE: Includes PR #112

NOTE: My OpenCL accelerator appears to deadlock on these unit tests when `i >= 16`
Intel HD Graphics 630 on OpenCL v1.2

### DIFF
--- a/Src/ILGPU.Tests/GroupOperations.cs
+++ b/Src/ILGPU.Tests/GroupOperations.cs
@@ -259,6 +259,7 @@ namespace ILGPU.Tests
             {
                 using var buffer = Accelerator.Allocate<int>(length * i);
                 buffer.MemSetToZero();
+                Accelerator.Synchronize();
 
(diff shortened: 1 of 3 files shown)


m4rs-mt:
I've got kind of the same feeling that the tests appears to deadlock on an AMD 5700XT when i >= 32

MoFtZ:
Should we rollback the CPU group barrier changes and the unit tests?

m4rs-mt:
A closer look revealed that it "seems to work" on all test machines using the latest AMD drivers. I would consider it to be "okay" for now. May be we should limit the test to the maximum warp size available on these devices? What about the potential warp size of your Intel HD Graphics device?

MoFtZ:
OK, updating the graphic drivers allows running `i <= 16`, previous it was only `i <= 8`. The warp size on the device is 16, so that works out well 👍

Issue #116, opened 2020-05-20
InvalidOperationException getting type member offset on Unix platforms using .Net Core 2.1.X or higher


adamreeve:
When running ILGPU.Tests.CPU on Linux (Ubuntu 20.04 in WSL with dotnet SDK 3.1.202 and runtime 2.1.18) I get the following error:

```
The active test run was aborted. Reason: Test host process crashed : Unhandled Exception: System.InvalidOperationException: Failed to compare two elements in the array. ---> System.ArgumentException: Type 'ILGPU.ArrayView`2[System.Int32,ILGPU.Index1]' cannot be marshaled as an unmanaged structure; no meaningful size or offset can be computed.
   at System.Runtime.InteropServices.Marshal.OffsetOfHelper(IRuntimeFieldInfo f)
       at ILGPU.IR.Types.TypeInformationManager.<>c__DisplayClass9_0.<CreateCompoundTypeInfo>b__0(FieldInfo left, FieldInfo right) in /mnt/c/Users/adree/gr/ILGPU/Src/ILGPU/IR/Types/TypeInformationManager.cs:line 325
          at System.Collections.Generic.ArraySortHelper`1.SwapIfGreater(T[] keys, Comparison`1 comparer, Int32 a, Int32 b)
       at System.Collections.Generic.ArraySortHelper`1.Sort(T[] keys, Int32 index, Int32 length, Comparison`1 comparer)
       --- End of inner exception stack trace ---
          at System.Collections.Generic.ArraySortHelper`1.Sort(T[] keys, Int32 index, Int32 length, Comparison`1 comparer)
       at ILGPU.IR.Types.TypeInformationManager.CreateCompoundTypeInfo(Type type) in /mnt/c/Users/adree/gr/ILGPU/Src/ILGPU/IR/Types/TypeInformationManager.cs:line 323
          at ILGPU.IR.Types.TypeInformationManager.CreateTypeInfo(Type type) in /mnt/c/Users/adree/gr/ILGPU/Src/ILGPU/IR/Types/TypeInformationManager.cs:line 307
             at ILGPU.IR.Types.TypeInformationManager.GetTypeInfo(Type type) in /mnt/c/Users/adree/gr/ILGPU/Src/ILGPU/IR/Types/TypeInformationManager.cs:line 229
[...]

PR #117, opened 2020-05-20
Add continuous integration pipeline using GitHub actions

adamreeve:
This adds a CI pipeline using GitHub actions that runs on all pushes to master and for any pull requests. The jobs will build with Debug and Release configurations and run the CPU tests, then create a NuGet package artifact.

In order to make this work I've added targets to the csproj files that allow building all T4 templates automatically, although I haven't made these run as part of the normal build as they make it take a lot longer.

You can see an example run on my fork: https://github.com/adamreeve/ILGPU/actions/runs/110085962

~~This only runs on Windows currently due to the tests failing on Linux (#116).~~ I've added Linux tests but have skipped the problematic tests when running under Linux for now.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,23 +22,6 @@ jobs:
         run: dotnet build --configuration=Release -p:TreatWarningsAsErrors=true Src
       - name: CPU Tests
         run: dotnet test --configuration=Release ./Src/ILGPU.Tests.CPU
-      - name: Upload bin directory
-        uses: actions/upload-artifact@v2
-        with:
-          name: ${{ runner.os }}-bin
-          path: Bin
-
-  nuget-package:
-    runs-on: windows-latest
-    needs: build-and-test
-    steps:
-      - name: Checkout
-        uses: actions/checkout@v2
-      - name: Download Windows binaries
-        uses: actions/download-artifact@v2
-        with:
-          name: Windows-bin
-          path: Bin
       - name: Set up NuGet
         uses: nuget/setup-nuget@v1
         with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,17 +15,13 @@ jobs:
         working-directory: ./Src
       - name: Build templates
         # TargetFramework specified so this only runs once rather than per framework
-        run: dotnet msbuild -t:TextTemplateTransform -p:TargetFramework=NA
-        working-directory: ./Src
+        run: dotnet msbuild -t:TextTemplateTransform -p:TargetFramework=NA Src
       - name: Debug Build
-        run: dotnet build --configuration=Debug -p:TreatWarningsAsErrors=true
-        working-directory: ./Src
+        run: dotnet build --configuration=Debug -p:TreatWarningsAsErrors=true Src
       - name: Release Build
-        run: dotnet build --configuration=Release -p:TreatWarningsAsErrors=true
-        working-directory: ./Src
+        run: dotnet build --configuration=Release -p:TreatWarningsAsErrors=true Src
       - name: CPU Tests
-        run: dotnet test --configuration=Release
-        working-directory: ./Src/ILGPU.Tests.CPU
+        run: dotnet test --configuration=Release ./Src/ILGPU.Tests.CPU
       - name: Upload bin directory
         uses: actions/upload-artifact@v2
         with:
@@ -48,10 +44,9 @@ jobs:
         with:
           nuget-version: '5.x'
       - name: Create NuGet package
-        run: nuget pack ILGPU.nuspec
-        working-directory: .nuget
+        run: nuget pack .nuget/ILGPU.nuspec
       - name: Upload NuGet package
         uses: actions/upload-artifact@v2
         with:
           name: nuget-package
-          path: .nuget/ILGPU.*.nupkg
+          path: ILGPU.*.nupkg
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,8 +2,6 @@ name: CI
 
 on:
   push:
-    branches:
-      - master
   pull_request:
 
 jobs:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,27 @@
+name: CI
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+
+jobs:
+  build-and-test:
+    runs-on: windows-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Restore tools
+        run: dotnet tool restore
+        working-directory: ./Src
+      - name: Build templates
+        # TargetFramework specified so this only runs once rather than per framework
+        run: dotnet msbuild -t:TextTemplateTransform -p:TargetFramework=NA
+        working-directory: ./Src
+      - name: Build
+        run: dotnet build --configuration=Release -p:TreatWarningsAsErrors=true
+        working-directory: ./Src
+      - name: CPU Tests
+        run: dotnet test --configuration=Release
+        working-directory: ./Src/ILGPU.Tests.CPU
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,9 +19,40 @@ jobs:
         # TargetFramework specified so this only runs once rather than per framework
         run: dotnet msbuild -t:TextTemplateTransform -p:TargetFramework=NA
         working-directory: ./Src
-      - name: Build
+      - name: Debug Build
+        run: dotnet build --configuration=Debug -p:TreatWarningsAsErrors=true
+        working-directory: ./Src
+      - name: Release Build
         run: dotnet build --configuration=Release -p:TreatWarningsAsErrors=true
         working-directory: ./Src
       - name: CPU Tests
         run: dotnet test --configuration=Release
         working-directory: ./Src/ILGPU.Tests.CPU
+      - name: Upload bin directory
+        uses: actions/upload-artifact@v2
+        with:
+          name: ${{ runner.os }}-bin
+          path: Bin
+
+  nuget-package:
+    runs-on: windows-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Download Windows binaries
+        uses: actions/download-artifact@v2
+        with:
+          name: Windows-bin
+          path: Bin
+      - name: Set up NuGet
+        uses: nuget/setup-nuget@v1
+        with:
+          nuget-version: '5.x'
+      - name: Create NuGet package
+        run: nuget pack ILGPU.nuspec
+        working-directory: .nuget
+      - name: Upload NuGet package
+        uses: actions/upload-artifact@v2
+        with:
+          name: nuget-package
+          path: .nuget/ILGPU.*.nupkg
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,6 +36,7 @@ jobs:
 
   nuget-package:
     runs-on: windows-latest
+    needs: build-and-test
     steps:
       - name: Checkout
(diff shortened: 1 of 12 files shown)


m4rs-mt:
@adamreeve A great job, thank you very much for submitting the PR and adding CI build features to this project.

Src/ILGPU.Tests.CPU/ILGPU.Tests.CPU.csproj
```diff
@@ -34,19 +39,16 @@
   </ItemGroup>
 
   <ItemGroup>
-    <ProjectReference Include="..\..\Src\ILGPU\ILGPU.csproj" />
-    <ProjectReference Include="..\ILGPU.Tests\ILGPU.Tests.csproj" />
-  </ItemGroup>
-
-  <ItemGroup>
-    <None Update="Configurations.tt">
+    <None Remove="Configurations.tt" />
```
m4rs-mt, line 42: I wonder about these `Remove` and `Include` operations. I suppose this is required to create the T4 templates automatically, right?

Src/ILGPU.Tests.Cuda/ILGPU.Tests.Cuda.csproj
```diff
@@ -66,4 +68,11 @@
     </Compile>
   </ItemGroup>
 
+  <Target Name="TextTemplateTransform">
+    <ItemGroup>
+      <TextTemplate Include="@(Content)" Condition="'%(Content.Extension)' == '.tt'">
+    </TextTemplate>
+  </ItemGroup>
+    <Exec WorkingDirectory="$(ProjectDir)" Command="dotnet tool run t4 %(TextTemplate.Identity) -I $([MSBuild]::ValueOrDefault('%(TextTemplate.RelativeDir)', '.')) -o %(TextTemplate.RelativeDir)%(TextTemplate.Filename).cs" />
```
m4rs-mt, line 76: Nice extension to automatically transform all T4 templates 🤓

Src/ILGPU/Frontend/Intrinsic/RemappedIntrinsics.Generated.tt
```diff
@@ -96,16 +96,16 @@ namespace ILGPU.Frontend.Intrinsic
 #if NETCORE
             var systemMathFType = typeof(MathF);
 
-<# foreach (var (method, type, _, className) in #>
-<#      unaryMathFunctions.Where(t => t.Item2 != null)) { #>
+<# foreach (var (method, type, _, className) in unaryMathFunctions
```
m4rs-mt, line 99: This is interesting; it looks like the T4-based parser may fail when using the previous code.

Src/ILGPU.Tests.OpenCL/ILGPU.Tests.OpenCL.csproj
```diff
@@ -66,4 +68,11 @@
     </Compile>
   </ItemGroup>
 
+  <Target Name="TextTemplateTransform">
```
GPSnoopy, line 71: It's probably worth consolidating all of these behind a single target file that you can import inside your csproj.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,59 @@
...
+      - name: Release Build
+        run: dotnet build --configuration=Release -p:TreatWarningsAsErrors=true
+        working-directory: ./Src
+      - name: CPU Tests
+        run: dotnet test --configuration=Release
+        working-directory: ./Src/ILGPU.Tests.CPU
+      - name: Upload bin directory
+        uses: actions/upload-artifact@v2
+        with:
+          name: ${{ runner.os }}-bin
```
jgiannuzzi, line 34: Is templating with `${{ runner.os }}` necessary?
I guess the idea is to have this job run on multiple platforms once possible, but we would only upload the Windows build, right?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,59 @@
...
+        working-directory: ./Src
+      - name: Debug Build
+        run: dotnet build --configuration=Debug -p:TreatWarningsAsErrors=true
+        working-directory: ./Src
+      - name: Release Build
+        run: dotnet build --configuration=Release -p:TreatWarningsAsErrors=true
+        working-directory: ./Src
+      - name: CPU Tests
+        run: dotnet test --configuration=Release
+        working-directory: ./Src/ILGPU.Tests.CPU
```
jgiannuzzi, line 30: Can we use `dotnet test --configuration=Release Src/ILGPU.Tests.CPU` instead of setting `working-directory`? Same remark for all the other `working-directory` instances except for `dotnet tool restore`.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,59 @@
...
+        with:
+          name: Windows-bin
+          path: Bin
+      - name: Set up NuGet
+        uses: nuget/setup-nuget@v1
+        with:
+          nuget-version: '5.x'
+      - name: Create NuGet package
+        run: nuget pack ILGPU.nuspec
+        working-directory: .nuget
```
jgiannuzzi, line 54: I wonder whether we should not create the nuget package directly inside the previous job. In that case, there would be no need to upload the bin directory as an artifact. The resulting artifact would only be the nuget package.

In the future, once we can run tests on multiple platforms, we might want to have a separate job for publishing the nuget package when all tests succeed and the CI is running for a tag like `v*`.

jgiannuzzi:
Awesome job, @adamreeve!

.github/workflows/ci.yml
m4rs-mt: Sounds reasonable; furthermore, this should simplify the whole internal deployment/test process.

Src/ILGPU/Frontend/Intrinsic/RemappedIntrinsics.Generated.tt
adamreeve: Yes the template wouldn't build for me without this change, I wasn't sure whether that was due to a difference in Visual Studio versions or whether it was broken for everyone but hadn't been noticed.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,59 @@
...
+      - name: Restore tools
+        run: dotnet tool restore
+        working-directory: ./Src
+      - name: Build templates
+        # TargetFramework specified so this only runs once rather than per framework
+        run: dotnet msbuild -t:TextTemplateTransform -p:TargetFramework=NA
+        working-directory: ./Src
+      - name: Debug Build
+        run: dotnet build --configuration=Debug -p:TreatWarningsAsErrors=true
+        working-directory: ./Src
```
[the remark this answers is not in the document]
adamreeve: :+1: That would simplify things, we could then use dotnet pack to build a package straight from the project and not need to install nuget separately.

.github/workflows/ci.yml
adamreeve: Yeah I separated them out thinking it would be cleaner if we built on multiple OSs and then created a NuGet package just from the Windows build, but we could still do that by conditionally running the NuGet package step and it would be simpler.

Src/ILGPU.Tests.CPU/ILGPU.Tests.CPU.csproj
adamreeve: Ah so this was needed when I first used Microsoft.TextTemplating.targets to run the transforms as that only worked with Content items, but now I've switched to using dotnet-t4 (because the first approach doesn't work with the dotnet commands) I realise I can revert this change and keep these as None items, so I'll do that to keep things simpler.

The tt files are automatically included as "None" items so this was removing them from the None items and then adding them as Content items.

adamreeve:
Thanks for the feedback, I've addressed all of the comments. I'll start adding automatic publishing of tagged releases to NuGet but think that might be best as a separate pull request.

Src/.config/dotnet-tools.json
```diff
@@ -0,0 +1,12 @@
+{
+  "version": 1,
+  "isRoot": true,
+  "tools": {
+    "dotnet-t4-fix": {
+      "version": "2.3.0-preview-0000-gb1ee03d9bc",
```
adamreeve, line 6: This should be updated to use a stable release of dotnet-t4 once a new release is available. I ran into errors with the most recent stable release so had to switch to this pre-release version.

adamreeve:
I've added some additional changes to run the build and tests on Linux too, but have set up the problematic tests to be skipped when running on Linux for now.

.github/workflows/ci.yml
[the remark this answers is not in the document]
jgiannuzzi: `dotnet pack` would be awesome, yes. I understand from @m4rs-mt that release-only builds are coming soon! 🥳

Src/ILGPU/Frontend/Intrinsic/RemappedIntrinsics.Generated.tt
m4rs-mt: I've spend some time investigating these issues. It seems to be a problem with different SDK versions.

.github/workflows/ci.yml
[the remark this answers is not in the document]
m4rs-mt: Exactly; however, this "feature" requires some updates to the internal compiler architecture. We are currently working on it for v0.8.1.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,40 @@
...
+      - name: Release Build
+        run: dotnet build --configuration=Release -p:TreatWarningsAsErrors=true Src
+      - name: CPU Tests
+        run: dotnet test --configuration=Release ./Src/ILGPU.Tests.CPU
+      - name: Set up NuGet
+        uses: nuget/setup-nuget@v1
+        with:
+          nuget-version: '5.x'
+      - name: Create NuGet package
+        run: nuget pack .nuget/ILGPU.nuspec
```
jgiannuzzi, line 34: I think those 2 steps should run only on Windows

.github/workflows/ci.yml
```diff
@@ -0,0 +1,40 @@
...
+
+on:
+  push:
+  pull_request:
+
+jobs:
+  build-and-test:
+    strategy:
+      matrix:
+        os: [ubuntu-18.04, windows-latest]
```
jgiannuzzi, line 11: I think we should add macos-latest in the test matrix

Src/ILGPU/Frontend/Intrinsic/RemappedIntrinsics.Generated.tt
m4rs-mt: I have encountered similar problems in the past regarding the "linefeed" characters appended to the end of the file, which had to be removed to have build support for Linux.

Src/ILGPU.Tests/ArrayViews.cs
```diff
@@ -381,7 +381,7 @@ public void ArrayViewGetVariableView(int length)
             data[index] = otherKernel[index];
         }
 
-        [Theory]
+        [WindowsOnlyTheory("Fails to create type information for ArrayView on Linux")]
```
jgiannuzzi, line 384: Comment should say "Unix" instead of "Linux" as this test also fails on macOS

Src/ILGPU.Tests/StructureValues.cs
```diff
@@ -124,7 +124,7 @@ public void StructureInterop<T>(T value)
             value.Val2[index] = val2;
         }
 
-        [Theory]
+        [WindowsOnlyTheory("Fails to create type information for ArrayView on Linux")]
```
jgiannuzzi, line 127: Comment should say "Unix" instead of "Linux" as this test also fails on macOS

jgiannuzzi:
Looks like I'm a bit late to the party with this review 🤷🏻‍♂️
@m4rs-mt @adamreeve Should I create a PR to address my comments above?

adamreeve:
@jgiannuzzi I'm happy to create another PR to address those

PR #120, opened 2020-05-21
Add MacOS to CI pipeline

adamreeve:
Also addresses other related comments from @jgiannuzzi  on #117.

I wonder what people think about using -latest vs specific OS versions? I had copied what Parquet sharp used with ubuntu-18.04 and windows-latest but seems a bit inconsistent to fix the Ubuntu version and not Windows and Mac.

Am making this as a draft PR for now as the Mac build seems to be hanging at compiling the T4 templates so need to investigate what's happening there (https://github.com/adamreeve/ILGPU/runs/695667380?check_suite_focus=true).

Edit: Also looks like the MacOS hosts will need the 2.1 runtime installed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,7 +8,7 @@ jobs:
   build-and-test:
     strategy:
       matrix:
-        os: [ubuntu-18.04, windows-latest, macos-latest]
+        os: [ubuntu-latest, windows-latest, macos-latest]
       fail-fast: false
     runs-on: ${{ matrix.os }}
     steps:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,10 +27,12 @@ jobs:
       - name: CPU Tests
         run: dotnet test --configuration=Release ./Src/ILGPU.Tests.CPU
       - name: Set up NuGet
+        if: runner.os == 'Windows'
         uses: nuget/setup-nuget@v1
         with:
           nuget-version: '5.x'
       - name: Create NuGet package
+        if: runner.os == 'Windows'
         run: nuget pack .nuget/ILGPU.nuspec
       - name: Upload NuGet package artifact
         if: runner.os == 'Windows'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,7 +8,7 @@ jobs:
   build-and-test:
     strategy:
       matrix:
-        os: [ubuntu-18.04, windows-latest]
+        os: [ubuntu-18.04, windows-latest, macos-latest]
       fail-fast: false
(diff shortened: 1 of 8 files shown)


jgiannuzzi:
Thanks for this PR, @adamreeve!

About using -latest vs specific OS versions: as this project does not depend on any native code, I would suggest we use -latest everywhere.

About T4: that's weird, as it works on my mac with the same OS and .NET Core versions. However I see it worked in the run for this PR.

One more thing: it looks like we should actually setup .NET Core 2.1 on macOS (and ideally on Linux to be consistent). Have a look at https://github.com/G-Research/ParquetSharp/blob/9dc36fad3e4b3b916a84f60248df57eaeff1bb80/.github/workflows/ci.yml#L63 for an example.

adamreeve:
Hmm yeah not sure what happened with the hanging step then, hopefully a random one-off thing...

Regarding the SDK setup, I actually tried that originally with Linux but it failed as I'm using "dotnet tool restore" which is only available since the 3.0 SDK, and that setup-dotnet action doesn't have support for installing different SDK and runtime versions side by side. Ideally we could use the 3.1 SDK with 2.1 runtime. It just works on Windows and Linux as there GitHub has the hosts set up with all the required SDK and runtime versions. I think I should be able to work around that by just installing the T4 tool explicitly each time rather than using "dotnet tool restore"

jgiannuzzi:
Not sure about this, but what about running that action twice, once with 2.1, and then once with 3.1?

m4rs-mt:
@adamreeve Hmm strange; We can also update the .Net Core version of all test cases if this helps to avoid the common problems. This should not hurt us, as ILGPU supports .Net Standard 2.0 and higher. We could even think about upgrading the .Net Core version of ILGPU, as well.

adamreeve:
Installing both SDK versions can work but needs a workaround to copy them into the same location, there's an issue open for this at https://github.com/actions/setup-dotnet/issues/25. Or I could install the 3.1 SDK to set up and run the T4 tool, then switch to the 2.1 SDK to run the tests, but that feels a bit hacky too.

Upgrading the test projects to 3.1 would be the nicest option from my point of view if that won't be too much pain for other developers who might need to then install the new SDK. Otherwise I'm happy to just change the workflow to install T4 as a global tool in the CI build which will work with the 2.1 SDK.

m4rs-mt:
> Installing both SDK versions can work but needs a workaround to copy them into the same location, there's an issue open for this at actions/setup-dotnet#25. Or I could install the 3.1 SDK to set up and run the T4 tool, then switch to the 2.1 SDK to run the tests, but that feels a bit hacky too.

This sounds like workarounds that are not really necessary. Let's take the upgrade approach, which shouldn't hurt many developers too much and makes the test integration process much easier.

Issue #98, opened 2020-05-06
OpenCL Intrinsic Implementations


m4rs-mt:
Since some intrinsics (`Atomics` in particular) are not supported in OpenCL in general, we need to provide software implementations for some of them (see also #94). This can be done in analogy to the `PTXIntrinsics` class, which defines software wrappers around several functions that cannot be expressed directly in `PTX`.

One example is `atomic_fetch_add` for floats and doubles that should be mapped to a software implementation using `Atomics.MakeAtomic` defined [here](https://github.com/m4rs-mt/ILGPU/blob/master/Src/ILGPU/Atomic.cs#L264).

m4rs-mt:
All features missing until now have been implemented. If we come across some missing features in the future, we can easily reopen this issue or create a new one.

PR #125, opened 2020-05-25
Add support for Group.BarrierPopCount on OpenCL

m4rs-mt:
Using `Group.BarrierPopCount` in OpenCL throws an `InternalCompilerError`, since this function cannot be mapped to a 'native' OpenCL instruction directly. This PR adds a new intrinsic software implementation for OpenCL to implement this function (see also #98).

### DIFF
--- a/Src/ILGPU/IR/Transformations/Optimizer.cs
+++ b/Src/ILGPU/IR/Transformations/Optimizer.cs
@@ -124,6 +124,11 @@ namespace ILGPU.IR.Transformations
             // Apply final DCE phase in release mode
             if (level > OptimizationLevel.O0)
                 builder.Add(new DeadCodeElimination());
+
+            // Append further backend specific transformations in release mode
+            builder.Add(new Inliner());
+            if (level > OptimizationLevel.O0)
+                builder.Add(new SimplifyControlFlow());
         }
 
         /// <summary>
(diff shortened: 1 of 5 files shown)


Src/ILGPU/IR/Transformations/Optimizer.cs
```diff
@@ -124,6 +124,11 @@ public static class Optimizer
             // Apply final DCE phase in release mode
             if (level > OptimizationLevel.O0)
                 builder.Add(new DeadCodeElimination());
+
+            // Append further backend specific transformations in release mode
```
MoFtZ, line 128: Is this change intented for this PR? It appears unrelated to Group.PopBarrierCount

m4rs-mt: Yes and no; perhaps this change should have been submitted in aother PR However, the OpenCL intrinsic does not appear to compile if the newly added `Group.BarrierPopCount` implementation is not inlined in the context of several AMD drivers. For this reason, I have modified the default backend transformations to include the `Inliner` phase by default (which was previously only enabled for the `PTXBackend`).

MoFtZ: Ah, I see. Fair enough.

PR #134, opened 2020-06-03
Fixed InvalidOperationException on Unix platforms

m4rs-mt:
The method `TypeInformationManager.CreateCompoundTypeInfo` throws an `InvalidOperationException` on Unix platforms (see #116). This PR solves this problem by using a different approach to sort all fields by their `MetadataToken` (more information about this idea can be found [here](https://github.com/dotnet/runtime/issues/19732)).


jgiannuzzi:
Well done (as always 😉), @m4rs-mt!

I think this PR should also re-enable the 2 tests that were failing on Unix. I can confirm that they now succeed on macOS and Linux.

PR #135, opened 2020-06-03
Fixed invalid placement of allocation nodes in LowerArrays.

m4rs-mt:
In special situations, `LowerArrays` can place allocation nodes in the wrong places. This PR solves this problem, which can occur in `PTX` and `OpenCL` backends.


Src/ILGPU/IR/Transformations/LowerArrays.cs
```diff
@@ -97,6 +97,7 @@ public ArrayTypeLowering(Method.Builder builder)
             var arrayLength = entryBlock.ComputeArrayLength(
                 location,
                 value.Extent);
+            entryBlock.SetupInsertPosition(value);
```
MoFtZ, line 100: QUESTION: Is line 96 still required?
`entryBlock.InsertPosition = 0;`

PR #140, opened 2020-06-05
Fixed invalid access to MethodCall parameters

MoFtZ:
Fixes the `IndexOutOfRangeException` from for https://github.com/m4rs-mt/ILGPU.Algorithms/issues/21

It some cases, it looks like the `Method` parameters are never populated. I'm not sure if this is intentional or an oversight.


m4rs-mt:
@MoFtZ I stumbled about the same problem. Maybe this issue is related to the intrinsic-specialization phase that imports methods from a different module. I am going to have a closer look

m4rs-mt:
I have made a detailed analysis of this problem. It turns out that the analysis tries to infer the address spaces of a `MethodCall` value to an intrinsic method that does not have a valid implementation at this point.

MoFtZ:
@m4rs-mt your PR looks good, I'll abandon this one.

PR #143, opened 2020-06-08
Updated ComputeRelativeSizeOf to calculate size of more than one element

MoFtZ:
First step in reducing the amount of memory required when calculating the size of a temporary buffer for data types smaller than an `int`.

For example, when calling `TempViewManager.Allocate<Int16>(length: 33)`, this will internally call `ComputeRelativeSizeof<int, Int16>()`. The size of `int` is 4 bytes, and the size of `Int16` is 2 bytes. The minimum required bytes to store 33 Int16s is 66 bytes.

Based on the original code, the relative size is 1 int, and we would then allocate 33 ints (132 bytes).

With the modified code, by calling `ComputeRelativeSizeof<int, Int16>(length: 33)`, we only need to allocate 17 ints (68 bytes).

### DIFF
--- a/Src/ILGPU/Interop.cs
+++ b/Src/ILGPU/Interop.cs
@@ -70,9 +70,10 @@ namespace ILGPU
         /// <typeparam name="TSecond">
         /// The base type that should be represented with <typeparamref name="TFirst"/>.
         /// </typeparam>
+        /// <param name="numSecondElements">The number of <typeparamref name="TSecond"/> elements to be stored.</param>
         /// <returns>
-        /// The number of required <typeparamref name="TFirst"/> instances to store an
-        /// instance of type <typeparamref name="TSecond"/>.
+        /// The number of required <typeparamref name="TFirst"/> instances to store <paramref name="numSecondElements"/>
+        /// instances of type <typeparamref name="TSecond"/>.
         /// </returns>
         [SuppressMessage(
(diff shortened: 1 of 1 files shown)


m4rs-mt:
@MoFtZ using this extension allows us to save memory when allocating multiple elements of a certain type. Nice extension 🤓

Src/ILGPU/Interop.cs
```diff
@@ -70,27 +70,28 @@ public static int SizeOf<T>(T structure)
...
+        /// instances of type <typeparamref name="TSecond"/>.
         /// </returns>
         [SuppressMessage(
            "Microsoft.Design",
             "CA1004:GenericMethodsShouldProvideTypeParameter",
             Justification = "The type is required for the computation of the " +
             "field offset")]
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        public static int ComputeRelativeSizeOf<TFirst, TSecond>()
+        public static int ComputeRelativeSizeOf<TFirst, TSecond>(int numSecondElements = 1)
```
m4rs-mt, line 84: I am not sure if the global program analysis accepts optional parameters without warnings...

MoFtZ: Split into two methods - the original without parameters (marked as `Obsolete`), and a new method with required parameter.

PR #151, opened 2020-06-11
Added extension methods to initialize MemoryBuffer instances using managed arrays.

m4rs-mt:
Previously it was necessary to allocate a buffer and copy the initial values in a separate step. This PR adds several extension methods to simplify the allocation of `MemoryBuffer` instances with managed data arrays.

### DIFF
--- a/Src/ILGPU/Runtime/MemoryBufferExtensions.cs
+++ b/Src/ILGPU/Runtime/MemoryBufferExtensions.cs
@@ -1070,4 +1070,9 @@ namespace ILGPU.Runtime
 
         #endregion
     }
+
+    /// <summary>
+    /// Extension methods for the allocation of memory buffers.
+    /// </summary>
+    public static partial class MemoryBufferExtensions { }
 }
--- a/Src/ILGPU/Runtime/MemoryBuffers.tt
+++ b/Src/ILGPU/Runtime/MemoryBuffers.tt
@@ -18,7 +18,10 @@
 using ILGPU.Util;
 using System;
 using System.Diagnostics;
+using System.Diagnostics.CodeAnalysis;
 using System.Runtime.CompilerServices;
+
+// disable: max_line_length
 
 namespace ILGPU.Runtime
 {
@@ -26,6 +29,7 @@ namespace ILGPU.Runtime
 <#      var typeName = i == 1 ? "MemoryBuffer" : $"MemoryBuffer{i}D"; #>
 <#      var arrayViewType = i == 1 ? "ArrayView" : $"ArrayView{i}D"; #>
 <#      var indexType = $"Index{i}"; #>
+<#      var arraySeparator = new string(Enumerable.Repeat(',', i - 1).ToArray()); #>
 <#      bool obsoleteCopyMethods = i > 1; #>
     /// <summary>
     /// Represents a <#= i #>D memory buffer that can be used in the scope
@@ -715,6 +719,54 @@ namespace ILGPU.Runtime
         }
 
         #endregion
+    }
+
+    public partial class MemoryBufferExtensions
+    {
+        /// <summary>
+        /// Allocates a <#= i #>D memory buffer with the given content on the associated
+        /// accelerator.
+        /// </summary>
+        /// <typeparam name="T">The element type.</typeparam>
+        /// <param name="accelerator">The current accelerator.</param>
+        /// <param name="data">The initial data array.</param>
+        /// <returns>The allocated memory buffer.</returns>
+<#      if (i > 1) { #>
+        [SuppressMessage(
+            "Microsoft.Performance",
+            "CA1814: PreferJaggedArraysOverMultidimensional",
+            Target = "data")]
+<#      } #>
+<#      var allocExpression = Enumerable.Range(0, i).Select(t => $"data.GetLength({t})"); #>
+        public static <#= typeName #><T> Allocate<T>(
+            this Accelerator accelerator,
+            T[<#= arraySeparator #>] data)
+            where T : unmanaged
+        {
+            if (data is null)
+                throw new ArgumentNullException(nameof(data));
+            var buffer = accelerator.Allocate<T>(<#= string.Join(", ", allocExpression) #>);
+            buffer.CopyFrom(data, <#= indexType #>.Zero, <#= indexType #>.Zero, buffer.Extent);
+            return buffer;
+        }
+
+        /// <summary>
+        /// Allocates a <#= i #>D memory buffer on the associated accelerator that is
+        /// initialized with 0-byte values.
+        /// </summary>
+        /// <typeparam name="T">The element type.</typeparam>
+        /// <param name="accelerator">The current accelerator.</param>
+        /// <param name="extent">The extent (number of elements to allocate).</param>
+        /// <returns>The allocated memory buffer.</returns>
+        public static <#= typeName #><T> AllocateZero<T>(
+            this Accelerator accelerator,
+            <#= indexType #> extent)
+            where T : unmanaged
+        {
+            var buffer = accelerator.Allocate<T>(extent);
+            buffer.MemSetToZero();
+            return buffer;
+        }
     }
 
 <#  } #>


Src/ILGPU/Runtime/MemoryBufferExtensions.cs
```diff
@@ -1070,4 +1070,89 @@ partial class MemoryBuffer3D<T>
...
+        /// <param name="data">The initial data array.</param>
+        /// <returns>The allocated memory buffer.</returns>
+        public static MemoryBuffer<T> Allocate<T>(
+            this Accelerator accelerator,
+            T[] data)
+            where T : unmanaged
+        {
+            if (data is null)
+                throw new ArgumentNullException(nameof(data));
+            if (data.Length < 1)
```
GPSnoopy, line 1094: IMHO Allocating 0 bytes should not be an error, as it avoid having to catter for the special case all over the user code. It definitely works with malloc(), how does CUDA react?

m4rs-mt: That makes sense to me. However, the current `MemoryBuffer` implementation does [not support](https://github.com/m4rs-mt/ILGPU/blob/master/Src/ILGPU/Runtime/MemoryBuffer.cs#L36) 0-length allocations.

MoFtZ:
I realise this is a draft, but adding `Stream` support would be nice 😅

m4rs-mt:
Makes sense to add 🤓

GPSnoopy:
While I can see why you'd want parity with the `Stream` API in most situations, I wonder if in this case it should be avoided. The underlying `cudaMalloc` and `cudaFree` will cause a GPU wide synchronisation and do not take a stream as an argument, while the memory copy will not happen asynchronously due to the host memory not being pinned.

So perhaps by not providing these extension methods on the `Steam` API it will avoid the user getting a false sense of security and parallelism?

MoFtZ:
@GPSnoopy there are no guarantees with the current PR (as it is) either - it uses the `DefaultStream` for the memory copy, which will also [happen asynchronously](https://docs.nvidia.com/cuda/cuda-driver-api/api-sync-behavior.html#api-sync-behavior).

I'm not sure about you, but my code looks like:
```csharp
var inputValues = /* stuff */;
using var stream = accelerator.CreateStream();

using var input = accelerator.Allocate<int>(1024);
input.CopyFrom(stream, inputValues, 0, 0, inputValues.Length);

using var output = accelerator.Allocate<int>(1024);
output.MemSetToZero(stream);

kernel(stream, input.Length, input.View, output.View);
stream.Synchronize();
```
with this PR it could hopefully become:
```csharp
var inputValues = /* stuff */;
using var stream = accelerator.CreateStream();

using var input = accelerator.Allocate(stream, inputValues);
using var output = accelerator.AllocateZero<int>(stream, 1024);

kernel(stream, input.Length, input.View, output.View);
stream.Synchronize();
```
@m4rs-mt Not important, but can we have a `AllocateZero`/`calloc`?

GPSnoopy:
AFAIK the default stream is always synchronous according to CUDA documentation.

`For code that is compiled using the --default-stream legacy compilation flag, the default stream is a special stream called the NULL stream and each device has a single NULL stream used for all host threads. The NULL stream is special as it causes implicit synchronization as described in Implicit Synchronization.`

Am I wrong to assume ILGPU uses the default behaviour?

On top of that

`3.2.5.5.4. Implicit Synchronization`

`Two commands from different streams cannot run concurrently if any one of the following operations is issued in-between them by the host thread:`
* `a page-locked host memory allocation,`
* `a device memory allocation,`
* `a device memory set,`
* `a memory copy between two addresses to the same device memory,`
* `any CUDA command to the NULL stream,`
* `a switch between the L1/shared memory configurations described in Compute Capability 3.x and Compute Capability 7.x.`

For your example, this is how I interpret the aforementionned behaviour:
```c#
var inputValues = /* stuff */;
using var stream = accelerator.CreateStream();

accelerator.Synchronize(); // Implicit GPU wide synchronisation
using var input = accelerator.Allocate<int>(inputValues.Length);
input.CopyFrom(stream, inputValues, 0, 0, inputValues.Length);

accelerator.Synchronize(); // Implicit GPU wide synchronisation
using var output = accelerator.Allocate<int>(1024);
output.MemSetToZero(stream);

kernel(stream, input.Length, input.View, output.View);
stream.Synchronize();

accelerator.Synchronize(); // Implicit GPU wide synchronisation (due to Dispose() of memory)
```

In a slightly more real-world example, you'd do something like this.
```c#
var inputValues = /* stuff */;
using var streams = Enumerable.Range(0, n).Select(i => accelerator.CreateStream()).ToArray();

using var input = accelerator.Allocate<int>(inputValues.Length);
using var output = accelerator.Allocate<int>(1024);

foreach (var stream in streams)
{
    input.CopyFrom(stream, inputValues, 0, 0, inputValues.Length);
    output.MemSetToZero(stream);

    kernel(stream, input.Length, input.View, output.View);
}

/* more stuff */

foreach (var stream in streams)
{
    stream.Synchronize();
}
```

Where you allocate all your memory and multiple streams up-front and then run several streams concurrently without any implicit synchronisations happening.

Bonus points for putting `inputValues` as pinned host memory first for even more parallelism.

In my experience, there is no reason why you'd want to mix CUDA streams and memory allocations/deallocations, given that the latter will do full GPU synchronisations under the hood.

m4rs-mt:
@GPSnoopy While I see the potential benefit in the future of exploiting an additional level of parallelism, it could give people a false intuition about (implicit) synchronization points (especially in the scope of Cuda programs). However, I can imagine that another GPU API (perhaps even Vulkan?) could benefit from these operations in the future.

@MoFtZ Additional extension methods for allocating "initialized" memory buffers with zeros sometimes make life easier. Especially if you want to make sure that the buffer contents do not consist of random bits 😄 

@GPSnoopy ILGPU uses the "standard" accelerator stream for Cuda, which is initialized with a `NULL` pointer. However, custom streams are automatically created with the flag `CU_STREAM_NON_BLOCKING` to allow concurrent processing. Moreover, ILGPU currently uses the driver function `cuMemAlloc_v2` to allocate memory, which is not associated with any GPU stream.

@MoFtZ @GPSnoopy I normally allocate *all* memory buffers at the beginning of the "application". As @GPSnoopy has shown, this usually only requires several copy operations that have to be performed afterwards which can be conveniently linked to a specific GPU stream. However, in the use case of @MoFtZ, he uses a user-defined `stream` instance which should be associated with all assignment, copy and start operations. However, as @GPSnoopy has explained, we cannot perform the actual array copy operation asynchronously because the array information is not pinned in memory.

MoFtZ:
@GPSnoopy I had assumed (incorrectly it seems), that memory allocation was 'streamless', but I guess it makes sense for it to use the default stream. In my usage, I never explicitly use the default stream which might explain why I never had any bottlenecks with memory allocation.

ILGPU uses the [CU_STREAM_NON_BLOCKING](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__STREAM.html#group__CUDA__STREAM_1ga581f0c5833e21ded8b5a56594e243f4) so that `the created stream should perform no implicit synchronization with stream 0`. My understanding is that this should allow the memory copy to run concurrently with any memory allocation (?).

And while I understand the benefits of pre-allocation of memory, it was never the bottleneck it my application - it wasn't particularly suited/worth the effect. Neither was stream pre-creation, which I created on demand.

@GPSnoopy @m4rs-mt With regards to pinned memory, I thought the [documentation on memory copy](https://docs.nvidia.com/cuda/cuda-driver-api/api-sync-behavior.html#api-sync-behavior) says it's asynchronous:
> Asynchronous
> 1. For transfers from device memory to pageable host memory, the function will return only once the copy has completed.
> 2. For transfers from any host memory to any host memory, the function is fully synchronous with respect to the host.
> 3. For all other transfers, the function is fully asynchronous. <b>If pageable memory must first be staged to pinned memory, this will be handled asynchronously with a worker thread.</b>

@m4rs-mt I don't think any of us are advocating for a `Task`-based API.

Src/ILGPU/Runtime/MemoryBufferExtensions.cs
MoFtZ: It looks like [cuMemAlloc](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__MEM.html#group__CUDA__MEM_1gb82d2a09844a58dd9e744dc31e8aa467) does not like 0 byte allocations:
`If bytesize is 0, cuMemAlloc() returns CUDA_ERROR_INVALID_VALUE.`

m4rs-mt:
> With regards to pinned memory, I thought the documentation on memory copy says it's asynchronous:'

@MoFtZ with respect to the underlying native-API: yes 🤓 However, I am referring to the fact that we have to "pin" the managed array using `fixed (....)` during the copy operation.

GPSnoopy:
FWIW we developed an algorithm that uses 8 concurrent streams (each CUDA stream had its own CPU thread as well) which involves a lot of copy from and to host memory. Using "naive" host memory pretty much gave a serialised timeline in NVIDIA Visual Profiler, while moving to pinned host memory gave the expected parallelism.

This was two years ago, so CUDA behaviour may have changed since then. Although unless observed to be as efficient wrt concurrency, or just better understood, I'm skeptical of the documentation claiming _"If pageable memory must first be staged to pinned memory, this will be handled asynchronously with a worker thread."_ Was this a recent change?

Src/ILGPU/Runtime/MemoryBufferExtensions.cs
m4rs-mt: A possible workaround would be that we always allocate at least one element. This also avoids OpenCL-related problems - since OpenCL doesn't seem to like 0-length allocations 😢

m4rs-mt:
@GPSnoopy I am not 100% sure at this point. However, I suspect that they might have changed this in one of the latest versions.

m4rs-mt:
@MoFtZ @GPSnoopy What is the "final" decision on possible `Stream`-based method overloads?

MoFtZ:
happy to leave them out 👍

GPSnoopy:
I'm also happy to leave out the zero-sized allocation support as well, since low-level CUDA/OpenCL don't like them anyway (even C specs says malloc() behaviour is platform dependent). We can come back to it later if needed.

m4rs-mt:
@GPSnoopy Please note that I have added additional `AllocateZero` extension methods in the current version. Are you suggesting to leave them out for the moment?

GPSnoopy:
@m4rs-mt Yep, those new extension methods look good. I was refering to the discussion around allowing allocating a CUDA memory block of length == 0.

m4rs-mt:
@GPSnoopy I see 😄 Regarding the `length == 0` discussion: I have changed the exception checks and adapted the code to always allocate at least one element.

GPSnoopy:
@m4rs-mt I just had a deeper look at the code. Isn't that going to break user assumptions?

For example:
```c#
using var buffer = accelerator.Allocate<float>(0);

if (buffer.Length != 0) throw new Exception("I would not expect the length not to be equal to the desired length");
```

Is the code above going to throw?

m4rs-mt:
@GPSnoopy Yeah, exactly. Your code will not throw in this particular case. What we could do is to "simulate" a length of 0 by allocating one element under the hood and using views with `length = 0`. However, I would prefer to ignore this for now and skip these changes as proposed.

Src/ILGPU/Runtime/MemoryBuffers.tt
```diff
@@ -717,5 +721,53 @@ namespace ILGPU.Runtime
...
+        /// <param name="accelerator">The current accelerator.</param>
+        /// <param name="data">The initial data array.</param>
+        /// <returns>The allocated memory buffer.</returns>
+<#      if (i > 1) { #>
+        [SuppressMessage(
+            "Microsoft.Performance",
+            "CA1814: PreferJaggedArraysOverMultidimensional",
+            Target = "data")]
+<#      } #>
+<#      var allocExpression = Enumerable.Range(0, i).Select(t => $"data.GetLength({i})"); #>
```
MoFtZ, line 740:
```suggestion
<#      var allocExpression = Enumerable.Range(0, i).Select(t => $"data.GetLength({t})"); #>
```

ISSUE: This is incorrectly generating code like: `var buffer = accelerator.Allocate<T>(data.GetLength(3), data.GetLength(3), data.GetLength(3));`

m4rs-mt: That's why code reviews are important 😄

PR #153, opened 2020-06-14
Multidimensional ExchangeBuffer Support

Yey007:
This is currently still a WIP. The 2D exchange buffer and copy2D test works, but I have no idea if I did everything right. I have not started the 3D exchange buffer and the 1D exchange buffer test is currently broken though. Anyways, it's a start.


Yey007:
Welp looks like I have some tests to pass...

Yey007:
Oh, it seems to be with the XML comments. I was getting warnings for those but had no idea what to do as I didn't touch the ones it's angry about. If anyone can point me in the right direction with those that would be great...

m4rs-mt:
Looks like a good place to start. I have a general remark at this point: ILGPU's coding convention uses the `ExtensionBuffer.cs` file for the `ExtensionBufferXYZ` class(es). Extension methods and all other high-level extension classes go into an `ExtensionBufferExtensions.cs` file. Please revert these changes to keep the original class implementation unless there is a very good reason to delete the existing source code from this file.

m4rs-mt:
The error messages you receive seem to be related to a missing `using ILGPU.Util;` statement in the beginning of the file. Since the `DisposeBase` class is defined in the `ILGPU.Util` namespace, the XML documentation functionality fails to lookup the internal function reference.

Yey007:
Alright, I'll make the changes right away. I'll also fix the XML comments per your suggestion. Thanks!

Yey007:
Oh, and one more question: Should higher level methods be implemented? Like filling a buffer with an array for example? Or do we leave that up to the user?

m4rs-mt:
@Yey007 a good question. There is a similar discussion about these features at the moment in #151.

Yey007:
@m4rs-mt Thank you. I would also like to ask about an idea. Currently, I notice that the `ExchangeBuffer<T>` classes share a lot of methods and even a nested class, mostly due to them all implementing `IMemoryBuffer<T>` in the same fashion and inheriting from `MemoryBuffer`. I'm not sure if it will work, but would it be viable (in terms of convention) to make a new class, `ExchnageBufferBase`, that would contain this shared implementations and the nested class in order to clean things up?

Yey007:
Anyways, I will mark this ready for review for now. This organizational change can probably be implemented in a separate PR if you guys see it fit.

Src/ILGPU/Runtime/ExchangeBuffer.cs
```diff
@@ -14,6 +14,7 @@
 using ILGPU.Util;
 using System;
 using System.Diagnostics;
+using System.Net;
```
m4rs-mt, line 17: I am wondering about this import. Is this required?

Src/ILGPU.Tests/ExchangeBufferOperations.cs
```diff
@@ -0,0 +1,204 @@
...
+
+        private const int Length = 32;
+
+        protected ExchangeBufferOperations(ITestOutputHelper output, TestContext testContext)
+            : base(output, testContext)
+        { }
+
+        internal static void CopyKernel(Index1 index, ArrayView<long> data)
+        {
+            data[index] -= 5;
```
m4rs-mt, line 22: The kernel name suggests that is copies several elements. However, it seems to manipulate the buffer data in-place. Is this intended? Furthermore, I am wondering whether this is really required in this context since such kernels are already tested in the scope of several other tests? Maybe I am missing the point here 😄

Src/ILGPU.Tests/ExchangeBufferOperations.cs
```diff
@@ -0,0 +1,204 @@
...
+using System.Linq;
+using System.Text;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace ILGPU.Tests
+{
+    public abstract class ExchangeBufferOperations : TestBase
+    {
+
```
m4rs-mt, line 13: - empty line

Src/ILGPU.Tests/ExchangeBufferOperations.cs
```diff
@@ -0,0 +1,204 @@
...
+            using var exchangeBuffer = Accelerator.AllocateExchangeBuffer<long>(Length);
+            for (int i = 0; i < Length; i++)
+            {
+                exchangeBuffer[new Index1(i)] = constant;
+            }
+            exchangeBuffer.CopyToAccelerator();
+
+            var expected = new long[Length];
+            for (int i = 0; i < Length; i++)
+            {
```
m4rs-mt, line 41: Singe line statements `inside` for and `if`  are not encapsulated by `{}` in the ILGPU code base 🔢

Src/ILGPU.Tests/ExchangeBufferOperations.cs
```diff
@@ -0,0 +1,204 @@
...
+        {
+            data[index] -= 5;
+        }
+
+        [Theory]
+        [InlineData(10)]
+        [InlineData(int.MaxValue)]
+        [InlineData(int.MinValue)]
+        [KernelMethod(nameof(Copy2DKernel))]
+        public void Copy2D(long constant)
```
m4rs-mt, line 68: Some here. I am wondering whether we can find an alternative way of testing the exchange buffers since their purpose is to transfer data and do not require specific kernel operations?

Src/ILGPU.Tests/ExchangeBufferOperations.cs
```diff
@@ -0,0 +1,204 @@
...
+
+            Accelerator.Synchronize();
+            Execute(exchangeBuffer.Extent, exchangeBuffer.View);
+            exchangeBuffer.CopyFromAccelerator();
+            Accelerator.Synchronize();
+
+            var data = exchangeBuffer.CPUView;
+            Assert.Equal(data.Extent.X, expected.Length);
+            for (int i = 0; i < Length; i++)
+            {
```
m4rs-mt, line 98: Same style comments here 🤓

Src/ILGPU.Tests/ExchangeBufferOperations.cs
```diff
@@ -0,0 +1,204 @@
...
+        }
+
+        [Theory]
+        [InlineData(10)]
+        [InlineData(int.MaxValue)]
+        [InlineData(int.MinValue)]
+        [KernelMethod(nameof(Copy3DKernel))]
+        public void Copy3D(long constant)
+        {
+            using var exchangeBuffer = Accelerator.AllocateExchangeBuffer<long>(new Index3(Length, Length, Length));
```
m4rs-mt, line 156: Try to keep the maximum line of 90 characters. This is not a strict limitations for the tests classes. However, it might look nicer and easier to read 🤓

Src/ILGPU/Runtime/ExchangeBuffer.cs
```diff
@@ -64,6 +65,82 @@ public static class ExchangeBuffer
...
+        /// <param name="extent">The extent (number of elements to allocate).</param>
+        /// <returns>The allocated 2D exchange buffer.</returns>
+        /// <remarks>
+        /// This function uses the default buffer allocation mode
+        /// <see cref="ExchangeBufferMode.PreferPagedLockedMemory"/>
+        /// </remarks>
+        public static ExchangeBuffer2D<T> AllocateExchangeBuffer<T>(
+            this Accelerator accelerator,
+            Index2 extent)
+            where T : unmanaged => accelerator.AllocateExchangeBuffer<T>(extent, ExchangeBufferMode.PreferPagedLockedMemory);
```
m4rs-mt, line 85: Please ensure a maximum line length of 90 characters. There are only a few places which violate these constraints in the ILGPU compiler code base.

Src/ILGPU/Runtime/ExchangeBuffer.cs
```diff
@@ -64,6 +65,82 @@ public static class ExchangeBuffer
...
+        /// <returns>The allocated 3D exchange buffer.</returns>
+        public static ExchangeBuffer3D<T> AllocateExchangeBuffer<T>(
+            this Accelerator accelerator,
+            Index3 extent,
+            ExchangeBufferMode mode)
+            where T : unmanaged
+        {
+            var gpuBuffer = accelerator.Allocate<T>(extent);
+            return new ExchangeBuffer3D<T>(gpuBuffer, mode);
+        }
```
m4rs-mt, line 143: Although these methods seem fine to me I am wondering whether we could come up with a T4 template to generate these classes "automagically"?

Src/ILGPU/Runtime/ExchangeBuffer.cs
```diff
@@ -97,7 +174,7 @@ public sealed unsafe class ExchangeBuffer<T> : MemoryBuffer, IMemoryBuffer<T>
         /// <summary>
         /// Represents the size of an element in bytes.
         /// </summary>
-        public static readonly int ElementSize = MemoryBuffer<T>.ElementSize;
+        public static readonly int ElementSize = MemoryBuffer<T, Index1>.ElementSize;
```
m4rs-mt, line 177: Was this change intended?

Src/ILGPU/Runtime/ExchangeBuffer.cs
```diff
@@ -381,4 +458,682 @@ protected override void Dispose(bool disposing)
...
     }
+
+    /// <summary>
+    /// A 2D buffer that stores a specified amount of elements on the associated accelerator
+    /// instance. Furthermore, it keeps a buffer of the same size in pinned CPU memory
+    /// to enable asynchronous memory transfers between the CPU and the GPU.
+    /// </summary>
+    /// <typeparam name="T">The element type.</typeparam>
+    /// <remarks>Members of this class are not thread safe.</remarks>
+    public sealed unsafe class ExchangeBuffer2D<T> : MemoryBuffer, IMemoryBuffer<T>
```
m4rs-mt, line 469: Maybe we should come up with an OO design that avoids many code duplications like `CudaViewSource`, which is essentially identical in all classes. This makes maintenance of the code bas significantly easier 🤓

Src/ILGPU/Runtime/ExchangeBuffer.cs
```diff
@@ -381,4 +458,682 @@ protected override void Dispose(bool disposing)
...
+        /// <summary>
+        /// Initializes this memory buffer.
+        /// </summary>
+        /// <param name="buffer">The underlying memory buffer.</param>
+        /// <param name="mode">The current buffer allocation mode.</param>
+        internal ExchangeBuffer2D(MemoryBuffer<T, Index2> buffer, ExchangeBufferMode mode)
+            : base(buffer.Accelerator, buffer.Extent.Size)
+        {
+            // Allocate CPU memory
+            cpuMemory = Accelerator is CudaAccelerator &&
```
m4rs-mt, line 560: Same here: please try to remove the amount of duplicated code as much as possible to reduce maintenance and bug-fixing overhead

Src/ILGPU/Runtime/ExchangeBuffer.cs
```diff
@@ -381,4 +458,682 @@ protected override void Dispose(bool disposing)
...
+        /// <param name="index2">The index to access</param>
+        /// <returns>The specified element</returns>
+        /// <remarks>
+        /// Uses the CPU View (<see cref="CPUView"/>), not the GPU View (<see cref="View"/>). <br></br>
+        /// You will need to copy again if you want use the modified version in a kernel.
+        /// </remarks>
+        public T this[Index2 index2]
+        {
+            get => CPUView[index2];
+
```
m4rs-mt, line 618: Nit: no empty line here

Src/ILGPU/Runtime/ExchangeBuffer.cs
```diff
@@ -381,4 +458,682 @@ protected override void Dispose(bool disposing)
...
+        /// Copies the current contents into a new array.
+        /// </summary>
+        /// <param name="stream">The used accelerator stream.</param>
+        /// <returns>A new array holding the requested contents.</returns>
+        public unsafe T[] GetAsArray(AcceleratorStream stream)
+        {
+            CopyFromAccelerator(stream);
+            stream.Synchronize();
+
+            var data = new T[Length];
```
m4rs-mt, line 664: It feels like this functionality already exists in the code base. Maybe you could find another way of expressing this without these code lines?

Src/ILGPU/Runtime/ExchangeBuffer.cs
```diff
@@ -381,4 +458,682 @@ protected override void Dispose(bool disposing)
...
+        /// </summary>
+        /// <param name="stream">The stream to use</param>
+        /// <returns>A jagged array containing all the elements in this exchnage buffer</returns>
+        public unsafe T[][] GetAs2DArray(AcceleratorStream stream)
+        {
+            CopyFromAccelerator(stream);
+            stream.Synchronize();
+
+            var data = new T[Length];
+            fixed (T* ptr = &data[0])
```
m4rs-mt, line 687: Same here

Src/ILGPU/Runtime/ExchangeBuffer.cs
```diff
@@ -381,4 +458,682 @@ protected override void Dispose(bool disposing)
...
+            {
+                System.Buffer.MemoryCopy(
+                    cpuMemoryPointer,
+                    ptr,
+                    LengthInBytes,
+                    LengthInBytes);
+            }
+
+            var data2D = new T[Extent.X][];
```
m4rs-mt, line 696: Same here: this essentially does exactly the same like a `MemoryBuffer2D` instance does

Src/ILGPU/Runtime/ExchangeBuffer.cs
```diff
@@ -381,4 +458,682 @@ protected override void Dispose(bool disposing)
...
+
+        #endregion
+
+        #region Nested Types
+
+        /// <summary>
+        /// Represents a view source that allocates native memory in page-locked CPU
+        /// memory.
+        /// </summary>
+        sealed class CudaViewSource : ViewPointerWrapper
```
m4rs-mt, line 824: Same here 🤓

Src/ILGPU/Runtime/ExchangeBuffer.cs
```diff
@@ -381,4 +458,682 @@ protected override void Dispose(bool disposing)
...
+        /// <param name="index3">The index to access</param>
+        /// <returns>The specified element</returns>
+        /// <remarks>
+        /// Uses the CPU View (<see cref="CPUView"/>), not the GPU View (<see cref="View"/>). <br></br>
+        /// You will need to copy again if you want use the modified version in a kernel.
+        /// </remarks>
+        public T this[Index3 index3]
+        {
+            get => CPUView[index3];
+
```
m4rs-mt, line 955: Same here

Src/ILGPU/Runtime/ExchangeBuffer.cs
```diff
@@ -381,4 +458,682 @@ protected override void Dispose(bool disposing)
...
+        /// </summary>
+        /// <param name="stream">The used accelerator stream.</param>
+        /// <returns>A new array holding the requested contents.</returns>
+        public unsafe T[] GetAsArray(AcceleratorStream stream)
+        {
+            CopyFromAccelerator(stream);
+            stream.Synchronize();
+
+            var data = new T[Length];
+            fixed (T* ptr = &data[0])
```
m4rs-mt, line 1002: And here 😄

Src/ILGPU/Runtime/ExchangeBuffer.cs
Yey007: No, I guess I never removed it.

Src/ILGPU.Tests/ExchangeBufferOperations.cs
Yey007: The point of the said test was to verify that copying to and from the accelerator actually changed the CPUView if the kernel contained an operation of course. I can try to come up with a more suitable name.

Src/ILGPU.Tests/ExchangeBufferOperations.cs
Yey007: I will fix this now.

Src/ILGPU.Tests/ExchangeBufferOperations.cs
Yey007: Sure, I guess we could do no operation at all. As I mentioned above, the point of these tests was to assert that the CPUView would get updated. If you find this unnecessary I can build a much cleaner test just requiring copy.

Src/ILGPU.Tests/ExchangeBufferOperations.cs
Yey007: I'll try to come up with simpler tests

Src/ILGPU/Runtime/ExchangeBuffer.cs
Yey007: I don't have experience with T4 templates, I would have to learn about those.

Src/ILGPU/Runtime/ExchangeBuffer.cs
Yey007: Yeah, that's what I was thinking with my suggestion above, before I marked it as ready for review.

Yey007:
So I refactored the code and cut down the size of ExchangeBuffer.cs to 500 lines. Still a lot to go until 90 lines per file, but it's a start. As you mentioned, this can either be achieved with T4 templates or adding more files and splitting code up accordingly. What's everyone's recommendation on that?

Src/ILGPU.Tests/ExchangeBufferOperations.cs
m4rs-mt: I see; that makes sense to me 🤓 . Otherwise, we cannot verify that the actual buffer content has been updated by the GPU. Let's keep those kernels 👍

Src/ILGPU.Tests/ExchangeBufferOperations.cs
m4rs-mt: Yeah, I got your idea. I just wanted to point out that we try to minimize the required "compiler capabilities" per test. Anyway, I think this makes sense in your case - furthermore, we might want to polish the `[InlineData]` attributes.

m4rs-mt:
@Yey007 Awesome work. I guess you are on the right track 🚀. Just to clarify things: I was talking about the _max line length of 90 characters per line_ and not the _maximum number of lines per file_ (see [here](https://github.com/m4rs-mt/ILGPU/blob/master/Src/.editorconfig#L23)).

You might want to install the `Editor Guidelines` extension for Visual Studio and enable the background code analysis for your project via `Tools->Options->Text Editor->C#->Advanced->Analysis: Background analysis scope: Entire Solution`.

Src/ILGPU/Runtime/ExchangeBuffer.cs
```diff
@@ -381,4 +445,122 @@ protected override void Dispose(bool disposing)
...
+
+        #endregion
+
+        #region Methods
+
+        /// <summary>
+        /// Gets this buffer as a 3D array from the accelerator
+        /// </summary>
+        /// <returns>The array containing all the elemnts in the buffer</returns>
+        public T[,,] GetAs3DArray() => new MemoryBuffer3D<T>(Buffer).GetAs3DArray();
```
m4rs-mt, line 562: Please be very careful about the allocation of temporary `MemoryBufferYD<T>` instances. Once your new `MemoryBuffer3D<T>` instance will be garbage collected, it will invoke its finalizer and dispose the associated `Buffer` instance. In this case, your buffer will be freed although there is still a reference in the scope of the `ExchangeBuffer` object.

Src/ILGPU/Runtime/ExchangeBuffer.cs
```diff
@@ -163,37 +243,23 @@ protected override void Dispose(bool disposing)
...
         [DebuggerBrowsable(DebuggerBrowsableState.Never)]
-        private readonly ViewPointerWrapper cpuMemory;
+        protected readonly ViewPointerWrapper cpuMemory;
 
         /// <summary>
         /// A cached version of the CPU memory pointer.
         /// </summary>
         [DebuggerBrowsable(DebuggerBrowsableState.Never)]
-        private readonly void* cpuMemoryPointer;
+        protected readonly void* cpuMemoryPointer;
```
m4rs-mt, line 252: You can create a `protected` property to access the `cpuMemoryPointer`. Alternatively, you can convert the pointer into an auto-implemented `readonly` property.

Src/ILGPU.Tests/Configurations.txt
```diff
@@ -1,13 +1,14 @@
-﻿ArrayViews: Debug, Release, O2
```
MoFtZ, line 1: ISSUE: Looks like there was an issue here with the merge with the master branch.

You've added `ExchangeBufferOperations: Debug, Release` but at the same time, someone elese added `, O2` to all the other lines.

Src/ILGPU/Runtime/ExchangeBuffer.cs
Yey007: Ok, I will implement it how I did with the 2D exchange Buffer then, I free it there.

Src/ILGPU/Runtime/ExchangeBuffer.cs
Yey007: Having them anything more visible than internal is what yields the "Do not declare visible instance fields" problem. [Microsoft](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1051?view=vs-2019) suggests making them private and using a property, so I guess that's what I'll do. Is that what you were saying? Sorry, I didn't quite understand.

Src/ILGPU.Tests/Configurations.txt
Yey007: Alright, I'll pull from the master branch to resolve these issues.

Yey007:
@m4rs-mt Oh so it needs to be less than 90 characters per line, not 90 lines per file. I was like, _how am I supposed to hit that?_

Src/ILGPU/Runtime/ExchangeBuffer.cs
Yey007: You're right, the functionality is already there. I fixed it.

Yey007:
For some reason, the mac-os test won't progress. Is there any way to restart it?

Src/ILGPU/Runtime/ExchangeBuffer.cs
Yey007: Anyways, I just did it with a protected property.

Src/ILGPU.Tests/ExchangeBufferOperations.cs
Yey007: Anyways, for now I've decided to use MemberData to reduce duplicate code. If you mean something different ou can reopen the conversation I think.

Src/ILGPU/Runtime/ExchangeBuffer.cs
```diff
@@ -165,35 +245,40 @@ protected override void Dispose(bool disposing)
...
+
         /// <summary>
         /// A cached version of the CPU memory pointer.
         /// </summary>
         [DebuggerBrowsable(DebuggerBrowsableState.Never)]
         private readonly void* cpuMemoryPointer;
 
         /// <summary>
-        /// Initializes this memory buffer.
+        /// Property for accessing cpuMemoryPointer
```
m4rs-mt, line 260: The dots again 🤓

Src/ILGPU/Runtime/ExchangeBuffer.cs
```diff
@@ -381,4 +462,200 @@ protected override void Dispose(bool disposing)
 
         #endregion
     }
+
+    /// <summary>
+    /// 1D implementation of <see cref="ExchangeBufferBase{T, TIndex}"/>
+    /// </summary>
+    /// <typeparam name="T">The element type.</typeparam>
+    /// <remarks>Members of this class are not thread safe.</remarks>
+    public sealed unsafe class ExchangeBuffer<T> : ExchangeBufferBase<T, Index1>
```
m4rs-mt, line 471: Looks good to me 🥇 Maybe, we can also generate these classes automatically?

m4rs-mt:
Overall: great work 🥇 👍

Src/ILGPU/Runtime/ExchangeBuffer.cs
```diff
@@ -64,6 +64,86 @@ public static class ExchangeBuffer
...
+        /// <returns>The allocated 3D exchange buffer.</returns>
+        public static ExchangeBuffer3D<T> AllocateExchangeBuffer<T>(
+            this Accelerator accelerator,
+            Index3 extent,
+            ExchangeBufferMode mode)
+            where T : unmanaged
+        {
+            var gpuBuffer = accelerator.Allocate<T>(extent);
+            return new ExchangeBuffer3D<T>(gpuBuffer, mode);
+        }
```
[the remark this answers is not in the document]
Yey007: I will have to work on this, I still haven't looked at T4 templates.

Yey007:
Alright, I _think_ I've sorted out the text template situation correctly, let me know what you think!

Src/ILGPU/Runtime/ExchangeBuffer.cs
Yey007: I've done this with the text template as well

Yey007:
So, I cleaned up the text template a bit, but I'm facing a weird formatting issue. The ///<summary> comments end up in the middle of the page, which is not what I wanted. I tried placing it in all sorts of different places within the text template, but haven't been able to get it to work. I was wondering if you could help me out?

Src/ILGPU.Tests/ExchangeBufferOperations.cs
```diff
@@ -0,0 +1,288 @@
...
+            exchangeBuffer.CopyToAccelerator();
+
+            var expected = new int[Length, Length];
+            for (int i = 0; i < Length; i++)
+                for (int j = 0; j < Length; j++)
+                    expected[i, j] = 10;
+
+            Accelerator.Synchronize();
+
+            //synchronizes on it's own
```
m4rs-mt, line 182: nit: please revise this comment

Src/ILGPU.Tests/ExchangeBufferOperations.cs
```diff
@@ -0,0 +1,288 @@
...
+
+            var expected = new int[Length, Length, Length];
+            for (int i = 0; i < Length; i++)
+                for (int j = 0; j < Length; j++)
+                    for (int k = 0; k < Length; k++)
+                        expected[i, j, k] = 10;
+
+            Accelerator.Synchronize();
+
+            //synchronizes on it's own
```
m4rs-mt, line 220: nit: same here

Src/ILGPU/Runtime/ExchangeBuffer.cs
```diff
@@ -1,61 +1,49 @@
...
 {
     /// <summary>
-    /// A static helper class for the class <see cref="ExchangeBuffer{T}"/>.
+    /// A static helper class for all exchange buffer implementations.
     /// </summary>
     public static class ExchangeBuffer
     {
-        /// <summary>
-        /// Allocates a new exchange buffer that allocates the specified amount of
+                                        /// <summary>
```
m4rs-mt, line 16: nit: fix broken summary indentation

Src/ILGPU/Runtime/ExchangeBuffer.cs
```diff
@@ -64,8 +52,85 @@ public static class ExchangeBuffer
             var gpuBuffer = accelerator.Allocate<T>(extent);
             return new ExchangeBuffer<T>(gpuBuffer, mode);
         }
-    }
+                                        /// <summary>
```
m4rs-mt, line 55: nit: same here; please add an empty line here to separate different method implementations. This makes it easier to read and simplifies navigation through the code using vim 🤓

Src/ILGPU/Runtime/ExchangeBuffer.cs
```diff
@@ -64,8 +52,85 @@ public static class ExchangeBuffer
...
+        public static ExchangeBuffer2D<T> AllocateExchangeBuffer<T> (
+            this Accelerator accelerator,
+            Index2 extent,
+            ExchangeBufferMode mode)
+            where T : unmanaged
+        {
+            var gpuBuffer = accelerator.Allocate<T>(extent);
+            return new ExchangeBuffer2D<T>(gpuBuffer, mode);
+        }
+                                        /// <summary>
```
m4rs-mt, line 94: nit: same here

Src/ILGPU/Runtime/ExchangeBuffer.tt
```diff
@@ -0,0 +1,480 @@
...
+        #region Methods
+
+        /// <summary>
+        /// Gets this buffer as a higher level memory buffer.
+        /// </summary>
+        /// <returns> A <see cref="<#= highLevelMemBuf #>{T}"/> containing the data in this
+        /// exchanage buffer.</returns>
+        public <#= highLevelMemBuf #><T> As<#= highLevelMemBuf #>() => new <#= highLevelMemBuf #><T>(Buffer);
+
+        <#if(i > 1)
```
m4rs-mt, line 443: This is the actual source of the indentation issue: you should start the `<# ...` command at the beginning of the line to avoid unnecessary indentations.

Src/ILGPU/Runtime/ExchangeBuffer.cs
Yey007: This was what I was asking about. I messed with the text template, but couldn't get it to work. I was hoping you could point me in the right direction.

Yey007: Ah, nevermind, you told me at the bottom. Thanks!

Src/ILGPU/Runtime/ExchangeBuffer.tt
Yey007: Hello,
Thanks for helping me out with this. I'm still confused on what I'm supposed to do. I was able to fix the problem by messing with indenting a bit more, but I assume you have a better solution. Could you illustrate exactly what you mean?

MoFtZ, line 443: @Yey007 For T4 templates, any extra whitespace is written to the output file.

You should move the opening `<#` to the start of the line. Then, either move the `if` statement to the start, or add whitespace before the `if` statement.

Note that Whitespace within `<#` and `#>` is fine, as this is not written to the output file.


```suggestion
<#if(i > 1)
```

```suggestion
<#        if(i > 1)
```

Yey007: Oh alright, I got it I think. Thanks.

Yey007:
Hello,
I completed the requested changes (hopefully). I also wanted to apologize for going back and forth so many times like this, this was my first PR and I couldn't get things right the first couple of times around. Thank you for your continued help on this.

Src/ILGPU/Runtime/ExchangeBufferExtensions.cs
```diff
@@ -0,0 +1 @@
+﻿
```
MoFtZ, line 1: ISSUE: `ExchangeBufferExtensions.cs` is a generated file. Please remove it from the commit, and add it to `.gitignore`

Src/ILGPU/Runtime/ExchangeBuffer.cs
```diff
@@ -1,15 +1,4 @@
-﻿// ---------------------------------------------------------------------------------------
```
MoFtZ, line 1: MINOR: Please add the copyright header back

Src/ILGPU/ILGPU.csproj
```diff
@@ -145,6 +150,16 @@
           <AutoGen>True</AutoGen>
           <DependentUpon>IntrinsicMatchers.tt</DependentUpon>
         </Compile>
+        <Compile Update="Runtime\ExchangeBuffer.cs">
```
MoFtZ, line 153: ISSUE: Line 153 to 157 appears to be obsolete, from when `ExchangeBuffer.cs` was changed into `ExchangeBuffer.tt`. Please remove.

Src/ILGPU/Runtime/ExchangeBuffer.cs
Yey007: I don't  remember removing it, that's odd. It must have been an accident. Can I just copy paste it from any other file?

Src/ILGPU/Runtime/ExchangeBufferExtensions.cs
Yey007: I thought I hadn't committed the generated version but oh well, I wasn't aware I had to add it to the .gitignore. I'll do that right away.

Src/ILGPU/Runtime/ExchangeBuffer.cs
Yey007: Also, should the copyright header be included in the text template?

Yey007: Anyways, it appears to be that way in MemoryBuffer.tt, so I'll follow the convention.

Src/ILGPU/ILGPU.csproj
```diff
@@ -145,11 +145,6 @@
           <AutoGen>True</AutoGen>
           <DependentUpon>IntrinsicMatchers.tt</DependentUpon>
         </Compile>
-        <Compile Update="Runtime\ExchangeBufferExtensions.cs">
-          <DesignTime>True</DesignTime>
-          <AutoGen>True</AutoGen>
-          <DependentUpon>ExchangeBufferExtensions.tt</DependentUpon>
-        </Compile>
```
MoFtZ: @Yey007 Looks like you removed too much. These lines should still be included.

In Visual Studio:
1. Right-click on `ExchangeBufferExtensions.tt` in the Solution Explorer.
1. Select `Run Custom Tool`
1. Go to the `File` menu, and select `Save All`.

The above lines should be added back to the project.

Yey007: Oh, I see. Right, I removed the wrong thing.

Src/ILGPU/Runtime/ExchangeBuffer.cs
```diff
@@ -3,7 +3,7 @@
 //                        Copyright (c) 2016-2020 Marcel Koester
 //                                    www.ilgpu.net
 //
-// File: ExchangeBuffer.cs
+// File: ExchnageBuffer.cs
```
m4rs-mt, line 6: nit: Typo

Src/ILGPU/ILGPU.csproj
```diff
@@ -255,6 +264,6 @@
         </EmbeddedResource>
     </ItemGroup>
 
-    <Import Project="Properties\ILGPU.nuspec.targets"/>
+    <Import Project="Properties\ILGPU.nuspec.targets" />
```
m4rs-mt, line 267: nit: was this change intended?

Src/ILGPU/Runtime/ExchangeBuffer.cs
```diff
@@ -165,35 +117,40 @@ protected override void Dispose(bool disposing)
...
             : base(buffer.Accelerator, buffer.Extent.Size)
         {
-            // Allocate CPU memory
-            cpuMemory = Accelerator is CudaAccelerator &&
-                mode == ExchangeBufferMode.PreferPagedLockedMemory
-                ? CudaViewSource.Create(buffer.LengthInBytes)
-                : (ViewPointerWrapper)UnmanagedMemoryViewSource.Create(
-                    buffer.LengthInBytes);
+            cpuMemory = buffer.Accelerator is CudaAccelerator &&
+                    mode == ExchangeBufferMode.PreferPagedLockedMemory
```
m4rs-mt, line 148: nit: Feels like the wrong indentation level; 4 spaces instead?

Src/ILGPU/Runtime/ExchangeBuffer.cs
```diff
@@ -218,24 +175,22 @@ internal ExchangeBuffer(MemoryBuffer<T, Index1> buffer, ExchangeBufferMode mode)
...
-        public ArrayView<T> CPUView { get; }
+        public ArrayView<T, TIndex> CPUView { get; protected set; }
 
         /// <summary>
         /// Returns a reference to the i-th element in CPU memory.
         /// </summary>
         /// <param name="index">The element index to access.</param>
         /// <returns>A reference to the i-th element in CPU memory.</returns>
-        public unsafe ref T this[Index1 index]
+        public T this[TIndex index]
```
m4rs-mt, line 190: nit: shouldn't this accessor return a reference to the i-th element, or is this load-by-value change intended?

Yey007: Oh, should I keep that as a reference? I guess I didn't update the comment, but if it's preferred to return a reference I can do that.

Src/ILGPU/Runtime/ExchangeBuffer.cs
Yey007: I left this how it was in the original ExchangeBuffer.cs if I remember correctly. Should it be changed back?

Yey007: I simply changed this because it looked cleaner to me and I didn't understand the point of returning a reference here.

Src/ILGPU/Runtime/ExchangeBuffer.cs
Yey007: I only changed this because it looked cleaner to me and I didn't understand the point of returning a reference here.

m4rs-mt: I guess we should return a reference to leave further load-store optimizations to the JIT compiler.

Src/ILGPU/Runtime/ExchangeBuffer.cs
m4rs-mt: I guess the indentation of `mode == ...` should be changed in this case because it has a relative indentation of 8 chars in comparison to the line `cpuMemory = ...`

Src/ILGPU/Runtime/ExchangeBuffer.cs
m4rs-mt: But I also agree with this piece of code 🤓

Yey007: Alright, sounds good. I'll return a reference.

Src/ILGPU/Runtime/ExchangeBuffer.cs
Yey007: `mode == ... ` should be 4 spaces ahead of `cpuMemory =...`?

Yey007:
The problems with the style test seem to stem from the text templates. When the template is actually transformed, the file is fine. Is it alright for the end product file to have some (maybe) unnecessary line breaks so the text template does not go over the limit?

Yey007:
For now, I fixed the problem. If it's preferred to just leave the problem for a slightly nice looking end product, I can revert the changes.

Yey007:
Thank you for helping me through the entire process, you've made my first open source contribution very easy and fun!

PR #156, opened 2020-06-15
Enabled ILGPU Release Builds

m4rs-mt:
Converted `Debug.Assert` operations into `Trace.Assert` calls and enabled verification and assertion support in ILGPU release builds.

### DIFF
--- a/Src/ILGPU/ContextFlags.cs
+++ b/Src/ILGPU/ContextFlags.cs
@@ -58,7 +58,12 @@ namespace ILGPU
         /// <summary>
         /// Enables detailed kernel statistics about all compiled kernel functions.
         /// </summary>
-        EnableKernelStatistics = 1 << 3,
+        EnableKernelStatistics = 1 << 4,
+
+        /// <summary>
+        /// Enables the internal IR verifier.
+        /// </summary>
+        EnableVerifier = 1 << 5,
 
         //
         // Code generation settings
--- a/Src/ILGPU/ContextFlags.cs
+++ b/Src/ILGPU/ContextFlags.cs
@@ -33,9 +33,14 @@ namespace ILGPU
         // Debugging settings
 
         /// <summary>
-        /// Enables debug information (if available),
+        /// Enables debug symbols (if available),
         /// </summary>
-        EnableDebugInformation = 1 << 0,
+        EnableDebugSymbols = 1 << 0,
+
+        /// <summary>
+        /// Enables debug information in kernels (if available).
+        /// </summary>
+        EnableKernelDebugInformation = 1 << 1,
 
         /// <summary>
         /// Enables inline source-code annotations when generating kernels.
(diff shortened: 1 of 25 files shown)


Src/ILGPU/ContextFlags.cs
```diff
@@ -37,23 +37,33 @@ public enum ContextFlags : int
         /// </summary>
         EnableDebugInformation = 1 << 0,
 
+        /// <summary>
+        /// Enables debug information in kernels (if available).
+        /// </summary>
+        EnableKernelDebugInformation = 1 << 1,
```
MoFtZ, line 43: SUGGESTION: I'm not sure if this is a breaking change, but could we rename this flag to something more descriptive?

QUESTION: When would I want to use `EnableKernelDebugInformation` rather than `EnableInlineSourceAnnotations`?

`ContextFlags` currently has:
1. EnableDebugInformation
    - enables `DebugInformationManager` for reading PDBs

1. EnableKernelDebugInformation
    - implicitly turns on `EnableDebugInformation`, enabling PDBs
    - enables `PTXDebugLineInfoGenerator`

1. EnableInlineSourceAnnotations
    - implicitly turns on `EnableKernelDebugInformation` (and therefore `EnableDebugInformation`), enabling PDBs
    - enables `PTXDebugSourceLineInfoGenerator`, replacing `PTXDebugLineInfoGenerator`

1. EnableKernelInformation
    - enables `KernelInfo`
    - unrelated to the others

m4rs-mt: The intention behind `EnableKernelDebugInformation` was to put a generated kernel module into `DEBUG` mode. This automatically disables all driver optimizations (to enable the NVIDIA Nsight debugger, for example). The difference to `EnableInlineSourceAnnotations` is that this flag inserts the corresponding source lines into a PTX module to simplify manual code analysis. However, this does not improve Cuda's debugging experience.

The problem in v0.8.0 was that `EnableDebugInformation` automatically enabled debug symbols for the kernel module *and* the ILGPU compiler error messages. This had a significant impact with respect to the kernel runtime of you application in debug mode.

m4rs-mt: More information about these flags:

- `EnableKernelDebugInformation` corresponds to the `-G` NVCC option
- `EnableInlineSourceAnnotations` corresponds to the `--source-in-ptx` NVCC option

What do you think of these flags? Should we somehow unify/rename them?

MoFtZ: what do you think about this:
1. keep `EnableKernelDebugInformation`
    - The Cuda documentation for `-G` says `Generate debug information for device code`, so it makes sense to follow the 'debug information' naming
1. rename `EnableKernelInformation` to `EnableKernelStatistics`
    - or to something other than 'Information'
1. keep `EnableDebugInformation`
    - or maybe rename to `EnableDebugSymbols`?
1. keep `EnableInlineSourceAnnotations`

m4rs-mt: @MoFtZ 1., 2. and 4. sound good to be. I also like the proposal of `EnableDebugSymbols`, since it actually enables debug symbols 😄

PR #163, opened 2020-06-20
Updated nuget package dependencies in .nuspec file.

MoFtZ:


m4rs-mt:
I had the same idea several minutes ago 😄

m4rs-mt:
I am going to update the version number and the .nuspec file structure a bit after merging

MoFtZ:
With the simplified Release-only nuget package, it might be possible to drop the `.nuspec` file and move the fuctionality into the `.csproj` file. That would simplify having to maintain the list of dependencies in multiple files.

m4rs-mt:
We can even think about removing the `.targets` file, since Nuget should be able to wire the framework assemblies automatically - in theory.

m4rs-mt:
You can also do this in the current PR if you like 🤓

PR #167, opened 2020-06-23
Fixed disassembly of generic methods

MoFtZ:
`Stelem_Ref` [does not have an argument](https://docs.microsoft.com/en-us/dotnet/api/system.reflection.emit.opcodes.stelem_ref?view=netframework-4.8).

`Ldtoken` may not be able to resolve generics, unless we pass the generic parameters to provide context.


MoFtZ:
I've added a unit test for `Ldtoken` and `Stelem_Ref`, and also fixed up `Ldelem_Ref`, `Ldind_Ref` and `Stind_Ref`.

However, I do not know how to write C# code that will use those new opcodes. Any suggestions?

Src/ILGPU.Tests/DisassemblerTests.cs
```diff
@@ -0,0 +1,46 @@
...
+        [InlineData(typeof(short))]
+        [InlineData(typeof(uint))]
+        [InlineData(typeof(int))]
+        [InlineData(typeof(ulong))]
+        [InlineData(typeof(long))]
+        [InlineData(typeof(float))]
+        [InlineData(typeof(double))]
+        public void StelemLdtoken(Type genericType)
+        {
+            var methodInfo = typeof(DisassemblerTests).GetMethod(nameof(StelemLdtokenKernel), BindingFlags.NonPublic | BindingFlags.Static);
```
m4rs-mt, line 39: nit: can we break this line?

m4rs-mt:
I have tried different ways to emit these instructions using the C# compiler. As you have already stated, it seems to be a bit tricky. I would recommend that we can skip these tests for now since I have verified your changes manually 🤓

Src/ILGPU.Tests/DisassemblerTests.cs
MoFtZ: Done

MoFtZ:
OK, I've fixed up the long lines in this PR.

OK, will skip those tests.

PR #169, opened 2020-06-24
Fixed misspelt license file

MoFtZ:


MoFtZ:
Yup, it only affected the Visual Studio solution explorer tree.

PR #170, opened 2020-06-26
Fixed emitting OpenCL floating point constants with suffix

MoFtZ:
When the constant `float.MaxValue` is appended into the source of an OpenCL kernel, it incorrectly becomes `3.40282347E+38.0f`. The check `if (value % 1.0f == 0.0f)` appears to be insufficient to detect that the scientific notation is being used.

In addition, floating point values have up to 9 digits of precision, but the `Single.ToString` method defaults to 7 digits. [The documentation](https://docs.microsoft.com/en-us/dotnet/api/system.single.tostring) recommends using the G9 format.

### DIFF
--- a/Src/ILGPU/Backends/OpenCL/CLCodeGenerator.Emitter.cs
+++ b/Src/ILGPU/Backends/OpenCL/CLCodeGenerator.Emitter.cs
@@ -28,6 +28,20 @@ namespace ILGPU.Backends.OpenCL
         /// </summary>
         public struct StatementEmitter : IDisposable
         {
+            #region Static
+
+            /// <summary>
+            /// Indicates char tokens in a formatted floating-point literal that
+            /// do not require a ".0f" suffix.
+            /// </summary>
+            private static readonly char[] FormattedFloatLiteralTokens =
+            {
+                '.',
+                'E'
+            };
+
+            #endregion
+
             #region Instance
 
             private readonly StringBuilder stringBuilder;
@@ -456,9 +470,17 @@ namespace ILGPU.Backends.OpenCL
                 }
                 else
                 {
-                    stringBuilder.Append(
-                        value.ToString(CultureInfo.InvariantCulture));
-                    if (value % 1.0f == 0.0f)
+                    // In C#, the floating point value "1.0f" can be shortened to "1f".
+                    // However, in the C programming language, it is necessary to include
+                    // the ".0f" suffix. However, if the stringified value already
+                    // contains a decimal point, or the exponent notation, appending the
+                    // "f" suffix is sufficient.
+                    var formattedValue =
+                        value.ToString("G9", CultureInfo.InvariantCulture);
+                    stringBuilder.Append(formattedValue);
+
+                    var idx = formattedValue.IndexOfAny(FormattedFloatLiteralTokens);
+                    if (idx == -1)
                         stringBuilder.Append(".0f");
                     else
                         stringBuilder.Append('f');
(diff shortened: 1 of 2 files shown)


Src/ILGPU/Backends/OpenCL/CLCodeGenerator.Emitter.cs
```diff
@@ -28,6 +28,15 @@ partial class CLCodeGenerator
         /// </summary>
         public struct StatementEmitter : IDisposable
         {
+            #region Static
+
+            /// <summary>
+            /// Indicates char tokens in a formatted floating-point literal that do not require a ".0f" suffix
```
m4rs-mt, line 34: nit: '.' dot ist missing

Src/ILGPU/Backends/OpenCL/CLCodeGenerator.Emitter.cs
```diff
@@ -28,6 +28,15 @@ partial class CLCodeGenerator
...
+            /// </summary>
+            private static readonly char[] FormattedFloatLiteralTokens = new char[] { '.', 'E', 'e' };
```
m4rs-mt, line 36: nit: please ensure max 90 chars per line

Src/ILGPU/Backends/OpenCL/CLCodeGenerator.Emitter.cs
```diff
@@ -456,9 +465,15 @@ public void AppendConstant(float value)
                 }
                 else
                 {
-                    stringBuilder.Append(
-                        value.ToString(CultureInfo.InvariantCulture));
-                    if (value % 1.0f == 0.0f)
+                    // In C#, the floating point value "1.0f" can be shortened to "1f". However,
```
m4rs-mt, line 468: nit: same here 🔢

Src/ILGPU/Backends/OpenCL/CLCodeGenerator.Emitter.cs
MoFtZ: Done

Src/ILGPU/Backends/OpenCL/CLCodeGenerator.Emitter.cs
MoFtZ: Done

Src/ILGPU/Backends/OpenCL/CLCodeGenerator.Emitter.cs
MoFtZ: Done

MoFtZ:
Updated formatting, and also removed `e` from the list of tokens because `G9` will only ever produce `E`.

Src/ILGPU/Backends/OpenCL/CLCodeGenerator.Emitter.cs
m4rs-mt: nit: please reformat to not exceed 90 chars per line

m4rs-mt:
Btw: you have change the `float` implementation. This problem also affects `doubles`, doesn't it?

Src/ILGPU/Backends/OpenCL/CLCodeGenerator.Emitter.cs
MoFtZ: Done

MoFtZ:
Yes, you're right. I've updated the case for `double` to use `G17`.

PR #171, opened 2020-06-26
Check style during CI run

jgiannuzzi:
This PR adds a CI job to check the coding style. This job runs in parallel with the other build-and-test jobs, and the publish-nuget job depends on it too.

Currently this is only checking the length of lines (< 90 characters). In the future, we might want to use [StyleCop](https://github.com/StyleCop/StyleCop) to enforce the style in the IDE and during build.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,17 +17,18 @@ jobs:
             Foreach ($file in Get-ChildItem -Path Src -Filter $pattern -Recurse -File) {
               If (-Not (Select-String -Path $file -Pattern "^// disable: max_line_length" -Quiet)) {
                 $index = 1
-            	  Foreach($line in Get-Content $file)
-            	  {
-            		  If ($line.Length -gt 90) {
+                Foreach ($line in Get-Content $file)
+                {
+                  If ($line.Length -gt 90) {
                     Write-Host "${file}:${index}: line too long ($($line.Length) > 90 characters)"
                     $found = $true
                   }
                   $index++
-            		}
-            	}
+                }
+              }
             }
           }
+          
           If ($found) {
             Exit 1
           }
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ jobs:
           $found = $false
           Foreach ($pattern in "*.cs","*.tt") {
             Foreach ($file in Get-ChildItem -Path Src -Filter $pattern -Recurse -File) {
-              If (-Not (Select-String -Path $file -Pattern "^// disable: max_line_length" -Quiet)) {
+              If (-Not (($file.Directory.Name -Eq "Resources") -Or (Select-String -Path $file -Pattern "^// disable: max_line_length" -Quiet))) {
                 $index = 1
                 Foreach ($line in Get-Content $file)
                 {
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,6 +5,33 @@ on:
   pull_request:
 
 jobs:
+  check-style:
+    runs-on: windows-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Check line length
+        run: |
+          $found = $false
+          Foreach ($pattern in "*.cs","*.tt") {
+            Foreach ($file in Get-ChildItem -Path Src -Filter $pattern -Recurse -File) {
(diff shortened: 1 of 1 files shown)


jgiannuzzi:
Once the style errors have been fixed in the main branch, the `check-style` job should be made a required check.

m4rs-mt:
I am going to fix the remaining issues; after that this PR should be ready to land.

MoFtZ:
Is there a way to add an ignore rule for `ErrorMessages.Designer.cs`?

MoFtZ:
I've put up a [PR to fix]( https://github.com/m4rs-mt/ILGPU/pull/172) the line lengths - except `ErrorMessages.Designer.cs`

.github/workflows/ci.yml
```diff
@@ -5,6 +5,33 @@ on:
...
 jobs:
+  check-style:
+    runs-on: windows-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Check line length
+        run: |
+          $found = $false
+          Foreach ($pattern in "*.cs","*.tt") {
```
MoFtZ, line 16: QUESTION: Should we remove the `*.tt` pattern?

jgiannuzzi:
> Is there a way to add an ignore rule for `ErrorMessages.Designer.cs`?

You can add a line starting with `// disable: max_line_length` anywhere in the file (ideally somewhere at the beginning).

.github/workflows/ci.yml
jgiannuzzi: @m4rs-mt only asked for `*.cs` originally, but I figured we should check the template files as well. Maybe we change the default line length for those? Or we add a way to make it configurable per file (like having a line starting with  `// max_line_length: xxx`)?

MoFtZ:
> 
> 
> > Is there a way to add an ignore rule for `ErrorMessages.Designer.cs`?
> 
> You can add a line starting with `// disable: max_line_length` anywhere in the file (ideally somewhere at the beginning).

`ErrorMessages.Designer.cs` is a file generated from `ErrorMessages.resx`. The RESX output is traditionally checked into source control because it does not get generated at build time - it only gets updated when the RESX file is updated.

So, the two solutions I can think of are:
1. add a special case into the `Check line length` script to ignore these files.
1. modify the build process to generate the RESX output at the PreBuild stage, then we can remove the `.Designer.cs` from source control.

What do you think?

m4rs-mt:
As @MoFtZ said: the `ErrorMessages.Designer.cs` and `RuntimeErrorMessages.Designer.cs` will be automatically generated by Visual Studio and should not contain custom annotations.

The problem related to the PreBuild stage is that people cannot compile ILGPU without using Visual Studio any more. Or is there an alternative solution using the `dotnet` pipeline?

MoFtZ:
@m4rs-mt @jgiannuzzi What are your thoughts on replacing this PR with [Menees.Analyzers](https://www.nuget.org/packages/Menees.Analyzers/)?

This would then allow integration into VS2019 to detect long line lengths.

m4rs-mt:
This package looks interesting. I just tested it, and it doesn't seem to check the `.tt` files, as you mentioned 😟 Maybe there is a hidden configuration file to change this behavior?

MoFtZ:
@m4rs-mt Ah, of course, it's a Roslyn analyzer, so it doesn't know about `.tt` files.

jgiannuzzi:
What about simply excluding the Resources directory from the script? After all, it is just a stop gap solution until we use StyleCop, right?

MoFtZ:
@jgiannuzzi StyleCop does not appear to have a line length rule. According to [Menees.Analyzers](https://github.com/menees/Analyzers), the line length rule was part of StyleCop+.

And the replacement [StyleCopAnalyzers](https://github.com/DotNetAnalyzers/StyleCopAnalyzers), which is Roslyn-based, has [an open PR for line length](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/1808) that is very old 😢

m4rs-mt:
@MoFtZ Yeah, it looks like it.
@jgiannuzzi I think this solution is okay for now. Let's exclude the `Resources` directory.

jgiannuzzi:
Thank you @MoFtZ for investigating the various options!

@m4rs-mt I have updated the CI job to exclude the `Resources` directory.

jgiannuzzi:
Rebased on latest master, which now includes #172 and #179.

PR #172, opened 2020-06-27
Fixed line lengths and general clean up

MoFtZ:


m4rs-mt:
I am currently refactoring all T4 templates. Consequently, we should not modify the templates in any way at the moment. However, we have to make sure that the line-length checks will not cause any build issues.

PR #177, opened 2020-07-01
Enabled code generation support for NVIDIA Ampere GPUs.

m4rs-mt:


Src/ILGPU/Backends/PTX/PTXInstructionSet.cs
```diff
@@ -89,6 +89,11 @@ namespace ILGPU.Backends
         /// </summary>
         public static readonly PTXInstructionSet ISA_64 = new PTXInstructionSet(6, 4);
 
+        /// <summary>
+        /// The 7.0 ISA.
+        /// </summary>
+        public static readonly PTXInstructionSet ISA_70 = new PTXInstructionSet(7, 0);
```
MoFtZ, line 95: QUESTION: Should we also add the 6.5 instruction set?

m4rs-mt: I was thinking the same thing. We can add the 6.5 instruction set for completeness, although we are not emitting instructions from that set at the moment.

PR #178, opened 2020-07-01
Fixed invalid intrinsic provider initialization in backend.

m4rs-mt:
The current `Backend` constructor initializes the associated `IntrinsicProvider` instance in an improper way. This results in all intrinsics being mapped invalidly to the wrong implementations.

### DIFF
--- a/Src/ILGPU/Backends/Backend.cs
+++ b/Src/ILGPU/Backends/Backend.cs
@@ -741,9 +741,7 @@ namespace ILGPU.Backends
             BackendFlags backendFlags,
             ArgumentMapper argumentMapper)
             : base(context, backendType, backendFlags, argumentMapper)
-        {
-            IntrinsicProvider = context.IntrinsicManager.CreateProvider<TDelegate>(this);
-        }
+        { }
 
         #endregion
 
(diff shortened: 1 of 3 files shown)


Src/ILGPU/Backends/Backend.cs
```diff
@@ -752,12 +750,25 @@ public abstract class Backend<TDelegate> : Backend
...
         #endregion
 
         #region Methods
 
+        /// <summary>
+        /// Initializes the current intrinsic provider.
+        /// </summary>
+        protected void InitIntrinsicProvider()
+        {
+            IntrinsicProvider?.Dispose();
```
MoFtZ, line 768: QUESTION: `InitIntrinsicProvider` is only called from the constructor. I would expect it to always be `null` at this point - is it necessary to call `Dispose` here?

m4rs-mt: It is not necessary at the moment; however, I thought that this could avoid programming errors in the future. But I can simply remove it for now.

MoFtZ: No, that's fine. Happy with the line there.

PR #179, opened 2020-07-01
T4 Refactoring

m4rs-mt:
The current implementation of all T4 templates in terms of basic types and arithmetic has grown over time. It has become very challenging to adapt and add new arithmetic operations in a convenient way. This PR contains refactored T4 templates to simplify the overall modeling of all text templates in the ILGPU ecosystem.


Src/ILGPU/Static/UnaryMathOperations.xml
```diff
@@ -0,0 +1,153 @@
...
+    <Operation Name="TanF">
+        <Summary>Computes tan(x).</Summary>
+        <Flags>Floats</Flags>
+        <Call>IntrinsicMath.CPUOnly.Tan</Call>
+        <Implementation>{MathType}.Tan({Value0})</Implementation>
+    </Operation>
+    <Operation Name="TanHF">
+        <Summary>Computes tanh(x).</Summary>
+        <Flags>Floats</Flags>
+        <Call>IntrinsicMath.CPUOnly.TanH</Call>
```
MoFtZ, line 99: ISSUE: The unit test in ILGPU.Algorithms is failing due to changing the case from `Tanh` to `TanH`. I think it is related to remapping intrinsics. Not sure why `ACos` and `ASin` are unaffected.

Src/ILGPU.Tests/AtomicCASOperations.tt
```diff
@@ -7,7 +7,7 @@ using Xunit;
 using Xunit.Abstractions;
 
 <#
-var atomicTypes = DefaultIntTypes.Concat(FloatTypes.Select(t => (t.Item1, t.Item2))).ToArray();
+var atomicTypes = DefaultIntTypes.Concat(FloatTypes.Skip(1)).ToArray();
```
MoFtZ, line 10: ISSUE: Incorrectly using `FloatTypes.Skip(1)`.

This was intended to skip the FP16 float type, but it does not yet exist. Instead, it is incorrectly skipping FP32.

```suggestion
var atomicTypes = AtomicNumericTypes;
```

Src/ILGPU.Tests/AtomicOperations.tt
```diff
@@ -3,11 +3,12 @@
 <#@ assembly name="System.Core" #>
 <#@ import namespace="System.IO" #>
 <#@ output extension=".cs" #>
+using System.Diagnostics.CodeAnalysis;
 using Xunit;
 using Xunit.Abstractions;
 
 <#
-var atomicTypes = DefaultIntTypes.Concat(FloatTypes.Select(t => (t.Item1, t.Item2))).ToArray();
+var atomicTypes = DefaultIntTypes.Concat(FloatTypes.Skip(1)).ToArray();
```
MoFtZ, line 11: ISSUE: Again, incorrectly using `FloatTypes.Skip(1)`.

Src/ILGPU/Static/UnaryMathOperations.xml
```diff
@@ -0,0 +1,153 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+<Operations>
+    <Operation Name="Neg">
+        <Summary>The neg operation.</Summary>
+        <Flags>Ints</Flags>
```
MoFtZ, line 5: QUESTION: Can this be used on floats?

Src/ILGPU/Static/UnaryMathOperations.xml
m4rs-mt: I have tried to unify all names; however, it makes more sense to keep the old names, since the .Net `Math` class uses the same names.

PR #182, opened 2020-07-06
Adding links to project readme

mfagerlund:
Every time I wanted to visit one of the related pages I had to look around like a crazy person to find the links - I think this will make the process simpler in the future!


Mattias-NCVIB:
I was unaware that the discord link was temporary/ephemeral, in that case it stands to reason to remove it. I shall update my pull request promptly!

Awesome tool, btw! I'm looking at porting my NN framework to it.

Issue #122, opened 2020-05-22
Support for Span<T>/Memory<T>?


matthew-a-thomas:
Is it possible to convert an `ArrayView<byte>` into a `Span<byte>` or a `Memory<byte>`? (Without a naive block copy)

If not, have you considered adding this support? I suspect it might be possible with a custom [`MemoryManager`](https://docs.microsoft.com/en-us/dotnet/api/system.buffers.memorymanager-1?view=netcore-3.1).

The motivation is easier integration with framework code. Many things (string decoding, SHA hashing, etc) can use `Span` to reduce allocations / improve speed.

Thanks!

matthew-a-thomas:
Glancing at your source code, I think this would certainly be possible. You're already dealing with a raw pointer here:
https://github.com/m4rs-mt/ILGPU/blob/master/Src/ILGPU/ArrayView.cs#L194

All that remains is to package that pointer up (with the length) into a `Memory<byte>` through a custom `MemoryManager`. The [`MemoryMarshal`](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.memorymarshal?view=netcore-3.1) class can then take care of quickly casting from `Span<byte>` to `Span<T:unmanaged>`.

I have a little bit of experience with this kind of thing. One time I used a custom `MemoryManager` with [`MemoryMappedFile`](https://docs.microsoft.com/en-us/dotnet/api/system.io.memorymappedfiles.memorymappedfile?view=netcore-3.1) to provide byref access to bytes on disk (with minimal allocation; certainly without having to marshal bytes).

m4rs-mt:
@matthew-a-thomas Thank you for your feature request. The short answer is yes, it makes sense to add support for these types in the future. However, you should be very careful when converting an `ArrayView<T>` instance to a `Memory<T>` (or `Span<T>`) instance, because direct memory accesses only work in the case of the `CPU Accelerator`. In addition, passing `Memory<T>` or `Span<T>` instances to a kernel requires a kernel-launcher extension to convert these types to GPU and runtime compatible launcher code.

MoFtZ:
> However, you should be very careful when converting an `ArrayView<T>` instance to a `Memory<T>` (or `Span<T>`) instance, because direct memory accesses only work in the case of the `CPU Accelerator`. In addition, passing `Memory<T>` or `Span<T>` instances to a kernel requires a kernel-launcher extension to convert these types to GPU and runtime compatible launcher code.

[...]

PR #185, opened 2020-07-10
Added functionality to convert exchange buffers to spans and memory

Yey007:
Per [issue #122](https://github.com/m4rs-mt/ILGPU/issues/122), this adds support for converting an `ExchachangeBufferBase<T, TIndex>` to `Memory<T>` and `Span<T>`. However, I do have some doubts about what I have written so far. 

1. I believe `Memory<T>` ends up returning a copy in this case, as it uses the `GetAsArray()` method which creates a copy. Is this fine? I think the no-copy functionality can be left up to `Span<T>` in this case, and I haven't found a good way to do this without copying yet. 

2. I left them as explicit operators as I'm not sure if they should be implicit, explicit, or methods. Explicit makes sense to me in this case, but I don't know the convention

3. I have not written any tests for this. I didn't write tests for the other operators either, but those were more simple. Should these have tests?


MoFtZ:
> * I believe `Memory<T>` ends up returning a copy in this case, as it uses the `GetAsArray()` method which creates a copy. Is this fine? I think the no-copy functionality can be left up to `Span<T>` in this case, and I haven't found a good way to do this without copying yet.

The original issue was about accessing the buffer without a naive block copy, so if that's not possible, maybe drop support for `Memory<T>`? The caller can always wrap the return value of `GetAsArray()` themselves if they want `Memory<T>`.

> * I left them as explicit operators as I'm not sure if they should be implicit, explicit, or methods. Explicit makes sense to me in this case, but I don't know the convention

My personal preference would be for a method, as it reduces the need to 'know' the generic type.
```CSharp
// Explicit cast
var span = (Span<double>)exchangeBuffer;

// Method
var span = exchangeBuffer.AsSpan();
```

You should also consider the different scenarios for how this will be used. Currently, you have implemented a way for the caller to access the `CPUView` as `Span<T>`. This would help the caller to initialise the exchange buffer before copying it's contents to the accelerator.

It may also be nice to provide a convenience method for the reverse operation of reading results. Similar to `GetAsArray`, it would copy from the accelerator, synchronise the stream, then return a span. This would allow the caller to retrieve the contents of the exchange buffer without an additional copy operation for building the returned array, and without having to write the same 3 lines of code.


> * I have not written any tests for this. I didn't write tests for the other operators either, but those were more simple. Should these have tests?

Yes please, some tests would be good 👍

Yey007:
Alright cool, I'll implement these changes.

Src/ILGPU.Tests/ExchangeBufferOperations.cs
```diff
@@ -149,7 +150,7 @@ public void GetAsArray()
             var data = exchangeBuffer.GetAsArray();
             Accelerator.Synchronize();
 
-            Assert.Equal(expected.Length, data.Length);;
+            Assert.Equal(expected.Length, data.Length); ;
```
MoFtZ, line 153: MINOR: Remove extra `;`.

Src/ILGPU/Runtime/ExchangeBuffer.cs
```diff
@@ -13,6 +13,7 @@
 using ILGPU.Runtime.Cuda.API;
 using ILGPU.Util;
 using System;
+using System.Buffers;
```
MoFtZ, line 16: MINOR: Remove unused namespace.

Src/ILGPU/Runtime/ExchangeBuffer.cs
```diff
@@ -313,6 +314,42 @@ public void CopyFromAccelerator(AcceleratorStream stream)
             return buffer.Buffer;
         }
 
+        /// <summary>
```
MoFtZ, line 317: MINOR: Please move all the functionality from the `Operators` region to the `Methods` region.

m4rs-mt:
@MoFtZ I think having `AsSpan` and `GetAsSpan` seems confusing, as it is really hard to understand what the difference between these functions actually is. `AsSpanFromAccelerator` also seems like a nice idea to me. However, I would expect that an `AsXXX` method would only convert the underlying pointers and give me the result, while a `GetXXX` method (like `GetAsArray`) sounds like getting a copy of the underlying data.

Anyway, `AsSpanFromAccelerator` makes a copy, but does not allocate a new buffer to write to. Consequently, `GetAsSpanFromAccelerator` would be a bit misleading in my opinion.

Yey007:
Maybe we can do something like `ToSpan` and `GetAsSpan`? I feel `ToSpan` would imply it's taking what's already there and just giving it back as a different format.

MoFtZ:
@Yey007 `AsSpan` is probably more appropriate than `ToSpan`. I see `ToSpan` as a conversion step, e.g. `ToString`, or `ToList`, whereas `AsSpan` is about a different representation (see `MemoryExtensions.AsSpan` or `PipeReader.AsStream`).

Maybe what we need is `CPUViewAsSpan` and `GetAsSpan` ?

m4rs-mt:
@MoFtZ Yeah, I totally agree. `ToSpan` feels a bit like a conversion method that has to perform several internal transformations to convert the current object state to the returned object instance. I am also happy with `CPUViewAsSpan` and `GetAsSpan`. However, the first option makes it more explicit to the user.

Yey007:
I like what @MoFtZ mentioned. What was your suggestion again @m4rs-mt ?

MoFtZ:
One more thought, does anyone actually use `CPUView` as an ArrayView? Could we change the property to return Span?

m4rs-mt:
@Yey007 I am totally happy with @MoFtZ's proposal.
@MoFtZ It makes sense to return a `Span<T>` instance instead.

MoFtZ:
OK, so in summary:
- Change `CPUView` property to return `Span<T>`.
    - No longer need `CPUViewAsSpan()`, `AsSpan()` or `ToSpan()` methods.
- Add `GetAsSpan()` and `GetAsSpan(Stream)` methods.

Yey007:
Alright, doing it right now.

Yey007:
I do have a question though. Should `CPUView` keep it's name? After all, it's not really an `ArrayView` anymore. Maybe something like `Span` would be a better name?

m4rs-mt:
@Yey007 This makes sense to me 🤓

Yey007:
Great idea.

Src/ILGPU/Runtime/ExchangeBuffer.cs
```diff
@@ -277,15 +295,41 @@ public void CopyFromAccelerator(AcceleratorStream stream)
...
         /// </summary>
         /// <param name="extent"></param>
         /// <returns>The view.</returns>
         public ArrayView3D<T> As3DView(Index3 extent) =>
-            CPUView.BaseView.As3DView<T>(extent);
+            cpuView.BaseView.As3DView<T>(extent);
+
+        /// <summary>
+        /// Gets this exchnage buffer as a <see cref="Span{T}"/>, copying from the
+        /// accelerator in the process
```
MoFtZ, line 310: MINOR: Spelling and full stop.

```suggestion
        /// Gets this exchange buffer as a <see cref="Span{T}"/>, copying from the
        /// accelerator in the process.
```

Src/ILGPU/Runtime/ExchangeBuffer.cs
```diff
@@ -277,15 +295,41 @@ public void CopyFromAccelerator(AcceleratorStream stream)
...
+        /// </summary>
+        /// <returns>
+        /// The <see cref="Span{T}"/> which accesses the part of this buffer on the CPU.
+        /// Uses the default accelerator stream.
+        /// </returns>
+        public Span<T> GetAsSpan() =>
+            GetAsSpan(Accelerator.DefaultStream);
+
+        /// <summary>
+        /// Gets this exchnage buffer as a <see cref="Span{T}"/>, copying from the
```
MoFtZ, line 320: MINOR: Spelling.

```suggestion
        /// Gets this exchange buffer as a <see cref="Span{T}"/>, copying from the
```

Src/ILGPU/Runtime/ExchangeBuffer.cs
```diff
@@ -177,17 +179,33 @@ protected override void Dispose(bool disposing)
         /// </summary>
         public TIndex Extent { get; protected set; }
 
+        /// <summary>
+        /// Internal array view
+        /// </summary>
+        [SuppressMessage(
+            "Microsoft.Design",
+            "CA1051: Do not declare visible instance fields",
+            Target = "target")]
```
MoFtZ, line 188: ISSUE: Remove `Target = "target"`. This is used for methods that have multiple parameters, and you want to pick which parameter to suppress.

Src/ILGPU/Runtime/ExchangeBuffer.cs
```diff
@@ -177,17 +179,33 @@ protected override void Dispose(bool disposing)
...
+        protected ArrayView<T, TIndex> cpuView;
```
MoFtZ, line 189: MINOR: Move this from `Properties` region to `Instance` region.

Actually, maybe let's keep it here, and instead of ignoring the warning, let's change this field into a property.

```suggestion
        protected ArrayView<T, TIndex> CpuArrayView { get; set; }
```

Src/ILGPU/Runtime/ExchangeBuffer.cs
```diff
@@ -177,17 +179,33 @@ protected override void Dispose(bool disposing)
...
         /// <summary>
         /// Returns an array view to the CPU part of this buffer.
         /// </summary>
-        public ArrayView<T, TIndex> CPUView { get; protected set; }
+        [Obsolete("Replaced by Span property")]
+        [CLSCompliant(false)]
+        public ArrayView<T, TIndex> CPUView => cpuView;
+
+        /// <summary>
+        /// Returns a span to the part of this buffer in CPU memory
```
MoFtZ, line 199: MINOR: Missing full stop.

```suggestion
        /// Returns a span to the part of this buffer in CPU memory.
```

Src/ILGPU/Runtime/ExchangeBuffer.cs
Yey007: I misspell that word around 80 times on a daily basis for some reason

Src/ILGPU/Runtime/ExchangeBuffer.cs
Yey007: Yeah I don't know why I didn't do this to begin with. So much cleaner.

Yey007: However, I feel the name should remain CPUView to stay consistent with the View property and (I believe) what it was before. What are your thoughts?

MoFtZ: I was thinking to call the protected property `CPUArrayView`, and the public property `CPUView`. That way we can add the `[Obsolete]`attribute to the public `CPUView` property. And in future, if we remove `CPUView`, we can leave the protected property unchanged.

Yey007: Actually yeah, that sounds good. I didn't really think about breaking changes here, so that's probably something I should've considered.

Issue #79, opened 2020-04-11
Instance kernel functions


maaron:
It would be nice if the generic kernel load methods could work with type inference, so that you don't have to spell out the kernel argument types twice.  Unfortunately, it seems method groups aren't infer-able in this case, but I tried to do the following, which does compile but throws at runtime:

```
static class InferredKernel
{
    static Action<Index1, ArrayView<int>> Kernel =
        (index, output) =>
        {
            output[index] = index;
        };

    static void Test()
    {
        var kernel = accelerator.LoadAutoGroupedStreamKernel(Kernel);
    }
}
```

Is there any other way around this?

m4rs-mt:
@maaron Thank you for your feedback. Unfortunately, type inference is currently performed very naively by the C# compiler, which leads to these detailed type annotations. There is not much we could do at the moment to solve these problems.

The problem with lambda functions in C# is that the C# compiler creates an instance method that references the parent lambda closure (also in your case). ILGPU does not currently support instance methods. However, I have already thought about adding support for lambda functions and limited support for reference types. This would allow you to compile this kernel function.

However, you might be able to use the @MoFtZ approach found [here](https://github.com/m4rs-mt/ILGPU.Samples/issues/7) to work around this problem.

Kemsekov:
This is total madness but it actually works.

You can compile any predefined amount of functions only if they are stored as strings and reference them from `Execute` method which creates a loaded kernel.

So if you have a list of functions stored as string, you can just do some string concatenation and name referencing to generate c# source code for kernel creation that you need.

It is not a perfect solution, but hey, it works!

https://gist.github.com/Kemsekov/df25b58cd8b14b032685253cecacd641

Issue #136, opened 2020-06-03
[F# Support] : MoveToImmutable can only be performed when Count equals Capacity.


ykafia:
I've rewritten one sample in F#

```fsharp
open System
open ILGPU
open ILGPU.Runtime


let invoke (action : Action<'a,'b,'c>) (args : 'a*'b*'c) = 
    action.Invoke(args)

let MyKernel (index : Index1)(dataView : ArrayView<int>)(constant : int) =
    dataView.[index] <- int index + constant


let sample = 
    use mutable context = new Context()
    for accelID in Accelerator.Accelerators do
        use mutable accel = Accelerator.Create(context,accelID)
        printfn "Performing operations on %A" accel
        let mutable kernel = accel.LoadAutoGroupedStreamKernel<Index1, ArrayView<int>, int>(MyKernel)
        use mutable buffer = accel.Allocate<int>(1024)
        invoke kernel (Index1(buffer.Length),buffer.View,42)
        accel.Synchronize()
        buffer.GetAsArray()
            |> Array.iteri (fun i x -> if x <> 42 then printfn "Error at element location %d : %A found" i x)
        printf "Done"



[<EntryPoint>]
let main argv =
    sample
    0

```

After running it, it gives me this error 

```code
Performing operations on CPUAccelerator [WarpSize: 1, MaxNumThreadsPerGroup: 8, MemorySize: 9223372036854775807]
Unhandled exception. System.TypeInitializationException: The type initializer for '<StartupCode$fsgame>.$Program' threw an exception.
 ---> System.InvalidOperationException: MoveToImmutable can only be performed when Count equals Capacity.
   at System.Collections.Immutable.ImmutableArray`1.Builder.MoveToImmutable()
   at ILGPU.Backends.EntryPoints.EntryPointDescription..ctor(MethodInfo methodSource, ParameterInfo[] parameters, IndexType indexType)
   at ILGPU.Backends.EntryPoints.EntryPointDescription.FromImplicitlyGroupedKernel(MethodInfo methodSource)
   at ILGPU.Runtime.Accelerator.LoadAutoGroupedKernel[TDelegate](MethodInfo method)
   at ILGPU.Runtime.KernelLoaders.LoadAutoGroupedKernel[TIndex,T1,T2](Accelerator accelerator, Action`3 action)
   at ILGPU.Runtime.KernelLoaders.LoadAutoGroupedStreamKernel[TIndex,T1,T2](Accelerator accelerator, Action`3 action)
[...]

PR #186, opened 2020-07-11
Added support for non-capturing lambda kernels.

MoFtZ:
Fixes https://github.com/m4rs-mt/ILGPU/issues/79 and https://github.com/m4rs-mt/ILGPU/issues/136.

Added heuristic for detecting non-capturing lambdas. The code generator then shifts all method arguments one place to the left, dropping the unused 'this' argument.

Also added a way for code generation on the background worker threads to record an exception, and return it to the original thread. This was needed because during unit tests, exceptions thrown in the worker thread terminate the unit test but do not mark the test as failed.


Src/ILGPU/Frontend/ILFrontend.cs
```diff
@@ -160,20 +170,41 @@ private void DoWork()
...
+                try
+                {
+                    codeGenerationPhase.GenerateCodeInternal(
+                        current.Method,
+                        current.IsExternalRequest,
+                        detectedMethods,
+                        out Method method);
+                    current.SetResult(method);
+                }
+#pragma warning disable CA1031 // Do not catch general exception types
```
m4rs-mt, line 182: Maybe we should convert the pragma to a `[Suppress...]` attribute?

Src/ILGPU/Frontend/ILFrontend.cs
```diff
@@ -160,20 +170,41 @@ private void DoWork()
...
 
-                    foreach (var detectedMethod in detectedMethods)
-                        processing.Push(new ProcessingEntry(detectedMethod, null));
+                    try
+                    {
+                        foreach (var detectedMethod in detectedMethods)
+                            processing.Push(new ProcessingEntry(detectedMethod, null));
+
+                    }
+#pragma warning disable CA1031 // Do not catch general exception types
```
m4rs-mt, line 200: nit: same here

Src/ILGPU/Frontend/CodeGenerator/CodeGenerator.cs
```diff
@@ -82,17 +83,19 @@ private void SetupVariables()
...
                     case ILInstructionType.Ldarga:
                         variables.Add(new VariableRef(
-                            instruction.GetArgumentAs<int>(), VariableRefType.Argument));
+                            instruction.GetArgumentAs<int>() - LambdaArgumentOffset,
+                            VariableRefType.Argument));
                         break;
                     case ILInstructionType.Ldloca:
                         variables.Add(new VariableRef(
-                            instruction.GetArgumentAs<int>(), VariableRefType.Local));
+                            instruction.GetArgumentAs<int>() - LambdaArgumentOffset,
```
m4rs-mt, line 91: Are you sure about this piece of code? The 'this' argument shouldn't change all local variable indexes in the function, should it?

Src/ILGPU/Util/MethodExtensions.cs
```diff
@@ -27,6 +28,49 @@ static class MethodExtensions
...
+            // CompilerGenerated attribute because F# uses a different attribute
+            // i.e. [CompilationMapping(SourceConstructFlags.Closure)], which only exists
+            // in F#. As a side-effect, the detection therefore also allows instance
+            // methods on any class without instance fields or properties to be
+            // considered as a non-capturing lambda.
+            //
+            // IMPORTANT: It is possible for the lambda to capture static fields and
+            // properties, and still pass this detection because we do not inspect the
+            // IL instructions here. We are relying on the rest of the compilation to
+            // detect invalid cases.
```
m4rs-mt, line 58: Makes sense to me 👍

Src/ILGPU/Frontend/CodeGenerator/CodeGenerator.cs
MoFtZ: Ah, my mistake, I misread that opcode thinking it was a variation to `Ldarg`. Will fix.

Src/ILGPU/Frontend/ILFrontend.cs
MoFtZ: Done.

Src/ILGPU/Frontend/ILFrontend.cs
MoFtZ: Done.

Src/ILGPU/Frontend/CodeGenerator/CodeGenerator.cs
MoFtZ: Done.

PR #187, opened 2020-07-11
Updated PTX intrinsics

MoFtZ:
Changed the `MaxArchitecture` of `PTXIntrinsic` to be an exclusive range.

Added missing `Tanh` intrinsic for FastMath.

For architectures that do not support the new `Tanh` intrinsic, added a code generator to mimic the original behaviour by throwing `NonSupportedIntrinsicException`.

### DIFF
--- a/Src/ILGPU/Backends/PTX/PTXIntrinsics.cs
+++ b/Src/ILGPU/Backends/PTX/PTXIntrinsics.cs
@@ -11,6 +11,7 @@
 
 using ILGPU.AtomicOperations;
 using ILGPU.Frontend;
+using ILGPU.IR;
 using ILGPU.IR.Intrinsics;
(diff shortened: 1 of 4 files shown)


Src/ILGPU/Backends/PTX/PTXIntrinsics.cs
```diff
@@ -112,6 +113,16 @@ public static void Register(IntrinsicImplementationManager manager)
                 CreateIntrinsic(
                     nameof(AssertFailed),
                     IntrinsicImplementationMode.Redirect));
+
+            // Register math
+            manager.RegisterUnaryArithmetic(
```
m4rs-mt, line 118: I thought about the same fact to avoid invalid PTX code 🤓 🚀

PR #196, opened 2020-07-18
Added support for 64-bit allocations and address computations

m4rs-mt:
This PR adds support for 64-bit allocations and address calculations. I have tried a variety of different approaches to implement 64-bit allocation/address computations in a well integrated way. In addition, this PR adds the `LongIndexX` types to represent indexes working on 64-bit integer values. However, there are three major breaking changes to the current API:

- `ArrayViewX`/`MemoryBufferX`: `int Length` and `int LengthInBytes`
  Will be changed to `long Length` and `long LengthInBytes` to support 32-bit and 64-bit allocations using the same types. In contrast to the ".Net way" (to support `Length` and `LongLength`), this PR changes the return types of these properties since it feels more natural to consider all allocations to be based on 64 bits. This ensures native support for all upcoming GPU and HPC architectures.
- `IIndex`: `int Size`
    Will be changed to `long Size` by default since it is easily possible to overflow a 32-bit integer using an `Index2` or an `Index3` instance. However, the existing `IndexX` types still support the old `int Size` property, as they can be used to compute addresses for 32-bit array views only.

To ensure high-performance address computations and to safe registers in the presence of 32-bit buffers, all `ArrayView`s still support default indexers using `IndexX` arguments. I have added additional assertions to ensure that a 32-bit indexer access is only allowed in the case of a 32-bit addressable buffer. In other words: If the array view extends over a range that cannot be addressed via a 32-bit index, the ILGPU runtime triggers an assertion.

Please note that there are implicit conversions between the types `IndexX` and their long counterparts `LongIndexX` and vice versa. This ensures backwards compatibility in a variety of different places. There are runtime assertion checks to verify that a `LongIndexX` can be safely converted to an `IndexX` without an integer overflow to avoid programming errors. However, we might want to remove these implicit conversions in the future.

Note further that most programs should not be affected by these changes.


Src/ILGPU/Backends/EntryPoints/EntryPointDescription.cs
```diff
@@ -56,7 +56,7 @@ namespace ILGPU.Backends.EntryPoints
             // Try to get index type from first parameter
             var firstParamType = parameters[0].ParameterType;
             var indexType = firstParamType.GetIndexType();
-            if (indexType == IndexType.None)
+            if (indexType == IndexType.None || indexType > IndexType.Index3D)
```
MoFtZ, line 59: QUESTION: This excludes `LongIndexX` from being used as the first parameter of a kernel. Is this the intended behaviour?

```suggestion
            if (indexType < IndexType.Index1D || indexType > IndexType.LongIndex3D)
```

Src/ILGPU/VariableView.cs
```diff
@@ -40,7 +40,9 @@ namespace ILGPU
         /// <param name="baseView">The base view.</param>
         public VariableView(ArrayView<T> baseView)
         {
-            Trace.Assert(baseView.IsValid && baseView.Length == 1, "Invalid base view");
+            Trace.Assert(
+                baseView.IsValid && baseView.Length == 1,
```
MoFtZ, line 44: MINOR: `Length` is now of type `long`.

```suggestion
                baseView.IsValid && baseView.Length == 1L,
```

Src/ILGPU/Backends/PointerViews/LowerPointerViews.cs
```diff
@@ -107,10 +121,17 @@ public PointerViewLowering(Method.Builder builder)
                 pointer,
                 value.Offset);
 
+            var length = value.Length;
+            if (length.BasicValueType < BasicValueType.Int64)
```
MoFtZ, line 125: QUESTION: What is the purpose of comparing `length.BasicValueType < BasicValueType.Int64`?

Is this really just `length.BasicValueType != BasicValueType.Int64`? Or are `BasicValueType.FloatXX` allowed?

Src/ILGPU/Runtime/MemoryBuffer.cs
```diff
@@ -786,28 +790,7 @@ public T[] GetAsArray(AcceleratorStream stream, TIndex offset, TIndex extent)
...
-        /// Returns a sub view of the current view starting at the given offset.
-        /// </summary>
-        /// <param name="offset">The starting offset.</param>
-        /// <param name="subViewExtent">The extent of the new sub view.</param>
-        /// <returns>The new sub view.</returns>
-        [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        [Obsolete("Use View.GetSubView(TIndex, Index) instead")]
-        public ArrayView<T, TIndex> GetSubView(TIndex offset, TIndex subViewExtent) =>
-            View.GetSubView(offset, subViewExtent);
+            GetAsRawArray(stream, Index1.Zero, (int)LengthInBytes).Array;
```
MoFtZ, line 793: MINOR: This calls into `GetAsRawArray(AcceleratorStream stream, long byteOffset, long byteExtent)`.

No need to cast `LengthInBytes` to `int`.

```suggestion
            GetAsRawArray(stream, Index1.Zero, LengthInBytes).Array;
```

Src/ILGPU/Runtime/CPU/CPUMemoryBuffer.cs
```diff
@@ -38,34 +38,37 @@ public sealed class CPUMemoryBuffer<T, TIndex> : MemoryBuffer<T, TIndex>
...
             {
                 case AcceleratorType.CPU:
-                    Unsafe.CopyBlock(
-                        target.LoadEffectiveAddress(),
+                    Buffer.MemoryCopy(
                         sourceAddress,
-                        (uint)target.LengthInBytes);
+                        targetAddress,
+                        target.LengthInBytes,
+                        target.LengthInBytes);
```
MoFtZ, line 67: QUESTION: Is this copy going to overrun the source buffer?

Src/ILGPU.Tests/ArrayViews.cs
```diff
@@ -85,11 +85,39 @@ public void ArrayViewLeaIndex(int length)
             Verify(buffer, expected);
         }
 
+        internal static void ArrayViewLongLeaIndexKernel(
```
MoFtZ, line 88: QUESTION: Spelling?

```suggestion
        internal static void ArrayViewLongLengthIndexKernel(
```

Src/ILGPU/IIndex.cs
```diff
@@ -71,19 +173,100 @@ indexType switch
...
+        /// <summary>
+        /// Returns true if the given type is a 64-bit index type.
+        /// </summary>
+        /// <param name="type">The managed .Net index type.</param>
+        /// <returns>True, if the given index type is a 64-bit index type.</returns>
+        public static bool IsLongIndex(this Type type)
+        {
+            switch (type.GetIndexType())
+            {
+                case IndexType.Index1D:
```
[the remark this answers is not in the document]
m4rs-mt: I think we should not add it here to support custom user-defined index types. `GetIndexType(this Type indexType)` can return `IndexType.None` in the case of a non-annotated index type. What do you think?

Src/ILGPU/Backends/PointerViews/LowerPointerViews.cs
```diff
@@ -81,10 +86,19 @@ public PointerViewLowering(Method.Builder builder)
...
         {
-            var length = context.Builder.CreateGetField(
+            var builder = context.Builder;
+            var length = builder.CreateGetField(
                 value.Location,
                 value.View,
                 new FieldSpan(1));
+
+            // Convert to a 32bit length value
+            if (value.Is32Bit)
```
MoFtZ, line 96: ISSUE: Should this be the inverse?

```suggestion
            if (!value.Is32Bit)
```

m4rs-mt: I have to admit that this property name is slightly misleading. `Is32Bit` returns true if the type of the `GetViewLength` IR node is essentially a 32-bit value. In this case we need to convert the 64-bit length information to a 32-bit int value. However, we should find a better name 😄

Src/ILGPU/IIndex.cs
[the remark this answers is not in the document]
MoFtZ: Is that going to cause issues with the rest of ILGPU code that throws exceptions when they find `IndexType.None`?

Looking at the code, it looks like the only places that throw exceptions are in `EntryPointDescription` - the constructor, `FromImplicitlyGroupedKernel` and `Validate` methods.

OK, make sense 👍

Src/ILGPU/Backends/PointerViews/LowerPointerViews.cs
m4rs-mt: Floats are not allowed; the idea was to cover all cases where `length` is not a 64-bit value. However, I also prefer `!=` in this context.

Src/ILGPU/IndexTypes.tt
```diff
@@ -0,0 +1,716 @@
...
+        #region IGenericIndex
+
+        /// <summary>
+        /// Returns true if the current index is greater than or equal to 0 and
+        /// is less than the given dimension.
+        /// </summary>
+        /// <param name="dimension">The dimension bounds.</param>
+        /// <returns>True if the current index is inside the given bounds.</returns>
+        public readonly bool InBounds(<#= name #> dimension) =>
+            <#= def.Expression(" & ", p => $"{p} >= 0 & {p} < dimension.{p}") #>;
```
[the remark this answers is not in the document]
m4rs-mt: I chose `&/|| in favor of `&&/|| to avoid short-circuit evaluation on GPUs in favor of more efficient predicated control flow. These patterns are converted by the `IfConversion` transformation anyway; however, this only happens if the current optimization level is set to `OptimizationLevel.O2`.

Src/ILGPU/Backends/EntryPoints/EntryPointDescription.cs
m4rs-mt: Currently, this patch does not include support for implicitly grouped kernels that use the `LongIndexX` types. If we add this convenience feature, we can remove these checks and accept all compatible intrinsic index types.

Src/ILGPU.Tests/ArrayViews.cs
m4rs-mt: I guess this should be okay since this test essential checks the `LoadElementAddress` (LEA) instruction using 64-bit (long) indexers.

Src/ILGPU/Runtime/CPU/CPUMemoryBuffer.cs
m4rs-mt: I think this should not happen as `Buffer.MemoryCopy` seems to have the semantics of a safe `memcpy` function in C.

Src/ILGPU.Tests/ArrayViews.cs
MoFtZ: Ah, Lea = LoadElementAddress. Nevermind, name is fine :+1:

Src/ILGPU/IndexTypes.tt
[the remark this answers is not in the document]
MoFtZ: OK, makes sense :+1:

m4rs-mt:
Regarding OpenCL: I haven't been able to successfully allocate a "large" memory buffer on any OpenCL device yet. Although the API seems to support `IntPtr` instances (as you mentioned), it does not seem to work properly.

Src/ILGPU/Runtime/CPU/CPUMemoryBuffer.cs
MoFtZ: I mean that `sourceAddress = ComputeEffectiveAddress(sourceOffset);`. Then, `sourceAddress + target.LengthInBytes` may read beyond `sourceAddress + MemoryBuffer.LengthInBytes`.

Src/ILGPU/Backends/EntryPoints/EntryPointDescription.cs
MoFtZ: Not critical 👍

Src/ILGPU/Runtime/CPU/CPUMemoryBuffer.cs
m4rs-mt: I have analyzed this piece of code, and it should not happen that overruns the source buffer, since all externally visible methods perform boundary checks. The previous piece of code also performed a simple block copy into the target buffer.

PR #206, opened 2020-07-24
Added IsCLSCompliant property to type information

MoFtZ:
Consolidates the logic of what is `CLSCompliant` to make maintenance easier.

Also removed `DefaultIntTypes` as it appears unused.

### DIFF
--- a/Src/ILGPU/Static/TypeInformation.ttinclude
+++ b/Src/ILGPU/Static/TypeInformation.ttinclude
@@ -59,6 +59,8 @@ public class TypeInformation
 
     public bool IsFloat => Kind == TypeInformationKind.Float;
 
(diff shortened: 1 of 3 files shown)


Src/ILGPU/Static/TypeInformation.ttinclude
```diff
@@ -59,6 +59,8 @@ public class TypeInformation
 
     public bool IsFloat => Kind == TypeInformationKind.Float;
 
+    public bool IsCLSCompliant => !(IsUnsignedInt || Name == "Int8" || Name == "Half");
```
m4rs-mt, line 62: Should the `Half` type be included here?

MoFtZ: I assumed it was not compliant, but I guess it's a struct, not a primitive. Removed.

PR #207, opened 2020-07-24
Check mismatch between version number and tag during CI build

jgiannuzzi:
Also add a version suffix to CI build artifacts when building for a branch or a PR.

### DIFF
--- a/Src/ILGPU/ILGPU.csproj
+++ b/Src/ILGPU/ILGPU.csproj
@@ -9,7 +9,7 @@
 
     <PropertyGroup>
         <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
-        <Version>0.8.1.1</Version>
+        <VersionPrefix>0.8.1.1</VersionPrefix>
         <NeutralLanguage>en-US</NeutralLanguage>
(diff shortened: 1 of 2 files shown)


Src/ILGPU/ILGPU.csproj
```diff
@@ -9,7 +9,7 @@
 
     <PropertyGroup>
         <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
-        <Version>0.8.1.1</Version>
+        <VersionPrefix>0.8.1.1</VersionPrefix>
```
m4rs-mt, line 12: Does this break the existing Nuget package building functionality that uses the `$(Version)` macro?

jgiannuzzi: No, I checked it and when not supplying `--version-suffix`, then `VersionPrefix` is used as the `Version`.

See the NuGet package generated here for validation: https://github.com/jgiannuzzi/ILGPU/runs/906353556?check_suite_focus=true. It's a build for a tag, so there is no suffix added. The `publish-nuget` failed because I do not have a NuGet API key to publish ILGPU, obviously 😅

m4rs-mt: Awesome 🚀

PR #209, opened 2020-07-26
Fixed support for 64-bit allocations and Half type.

MoFtZ:


Src/ILGPU/Interop.cs
```diff
@@ -116,7 +117,13 @@ public static int SizeOf<T>(T structure)
             var firstSize = SizeOf<TFirst>();
             var secondSize = SizeOf<TSecond>();
 
-            return IntrinsicMath.DivRoundUp(secondSize * numSecondElements, firstSize);
+            var relativeSize = IntrinsicMath.DivRoundUp(
+                secondSize * numSecondElements,
+                firstSize);
+            Trace.Assert(
+                relativeSize >= int.MinValue & relativeSize <= int.MaxValue,
+                "32-bit index out of range");
```
m4rs-mt, line 125: Can we use `IndexTypeExtensions.AssertIntIndexRange` instead of this explicit `Trace.Assert`?

MoFtZ: Done.

PR #219, opened 2020-07-31
Ensure compiler exceptions are annotated with location information.

MoFtZ:
Attempts to trap all compiler exceptions, particularly those without location information, and generates a new `InternalCompilerException` with location information.

REQUEST FOR COMMENT: Naming of `ILGPU.InternalCompilerException`.

### DIFF
--- a/Src/ILGPU.Tests/KernelEntryPoints.cs
+++ b/Src/ILGPU.Tests/KernelEntryPoints.cs
@@ -491,8 +491,9 @@ namespace ILGPU.Tests
                 };
 
             using var buffer = Accelerator.Allocate<int>(length);
-            Assert.Throws<NotSupportedException>(() =>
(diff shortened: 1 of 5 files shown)


Src/ILGPU/InternelCompilerException.cs
```diff
@@ -0,0 +1,72 @@
+﻿// ---------------------------------------------------------------------------------------
+//                                        ILGPU
+//                        Copyright (c) 2016-2020 Marcel Koester
+//                                    www.ilgpu.net
+//
+// File: InternelCompilerException.cs
```
m4rs-mt, line 6: Question (Typo?): Internel -> Internal (affects all changes)

MoFtZ: Done 👼

PR #224, opened 2020-08-05
Blocking cuda streams

Yey007:
This was requested in issues so I thought it would be good for me to do this since it involves the backend a little bit more, which I want to understand more. I haven't really tested this because I'm not sure _how_ I would test it. If anyone can point me in the right direction in that regard, that would be great. Also for some reason git is showing that I added HalfConversion.cs which I didn't write, so I guess you can ignore that.


Src/ILGPU/Runtime/Cuda/CudaAccelerator.cs
```diff
@@ -449,6 +449,13 @@ public CudaCacheConfiguration CacheConfiguration
         protected override AcceleratorStream CreateStreamInternal() =>
             new CudaStream(this);
 
+        /// <summary>
+        /// Creates a CUDA stream with blocking calls
+        /// </summary>
+        /// <returns>The created CUDA stream</returns>
+        public CudaStream CreateBlockingStream() =>
```
MoFtZ, line 456: QUESTION: Should we expose a more generic API?
e.g. `public CudaStream CreateStream(StreamFlags flags)`

QUESTION: Should we expose `CreateStream(IntPtr)` to allow the caller to provide an externally created stream?

Src/ILGPU/Runtime/Cuda/CudaAccelerator.cs
```diff
@@ -449,6 +449,13 @@ public CudaCacheConfiguration CacheConfiguration
         protected override AcceleratorStream CreateStreamInternal() =>
             new CudaStream(this);
 
+        /// <summary>
+        /// Creates a CUDA stream with blocking calls
```
MoFtZ, line 453: MINOR: End sentences with full-stop.

Src/ILGPU/Runtime/Cuda/CudaAccelerator.cs
```diff
@@ -449,6 +449,13 @@ public CudaCacheConfiguration CacheConfiguration
...
+        /// </summary>
+        /// <returns>The created CUDA stream</returns>
```
MoFtZ, line 455: MINOR: End sentences with full-stop.

Src/ILGPU/Runtime/Cuda/CudaStream.cs
```diff
@@ -52,6 +53,21 @@ internal CudaStream(Accelerator accelerator)
...
         }
 
+        /// <summary>
+        /// Constructs a new cuda stream with given <see cref="StreamFlags"/>
+        /// </summary>
+        /// <param name="accelerator">The associated accelerator.</param>
+        /// <param name="flag">
+        /// Stream flag to use. Allows blocking and non-blocking streams.
+        /// </param>
+        internal CudaStream(Accelerator accelerator, StreamFlags flag)
```
MoFtZ, line 63: QUESTION: Should we add this method as-is? Or modify the existing `CreateStream(Accelerator)` method to accept `StreamFlags`?

m4rs-mt: That is a good question. Maybe we want to make the blocking/non-blocking functionality generally available to all accelerators. The different platform-dependent implementations can then decide for themselves how to realize the intended behavior.

Src/ILGPU/Runtime/Cuda/CudaAccelerator.cs
m4rs-mt: @MoFtZ I think we should also expose the `CreateStream(IntPtr)` constructor. This extends the possibilities to use low-level Cuda API interop.

Src/ILGPU/Runtime/Cuda/CudaStream.cs
Yey007: I'm sure we can make this work on the CPU, but I'm skeptical on how we would get this to work on OpenCL. Isn't the whole idea of queues based on enqueued operations happening at some point and not when you enqueue them?

samy2: I have no clear idea to the question, but to make things a bit clearer, when we create a stream without CU_STREAM_NON_BLOCKING, it does not create a "blocking" stream : the kernels executing in this stream are still asynchronous, BUT, from the CUDA documentation :
CU_STREAM_NON_BLOCKING: Specifies that work running in the created stream may run concurrently with work in stream 0 (the NULL stream), and that the created stream should perform no implicit synchronization with stream 0.
So if I understand correctly, the default behaviour of CUDA streams is a bit strange : they do implicit synchronization with stream 0 (the "default" stream), but not with other stream. I am not sure we want to emulate this strange behaviour to other accelerators. Maybe exposing public CudaStream CreateStream(StreamFlags flags) is enough.

Yey007: Looking at CL properties we can use, I don't see anything that would fit here to give the functionality as CUDA, maybe besides `CL_​QUEUE_​OUT_​OF_​ORDER_​EXEC_​MODE_​ENABLE`, but that's a stretch and would likely cause confusion.

MoFtZ: The `CudaStream` class currently has the constructor `internal CudaStream(Accelerator)`. My initial suggestion was, rather than adding a second constructor `internal CudaStream(Accelerator, StreamFlags)`, we just have a single constructor that accepts `StreamFlags`.

`StreamFlags` is currently only relevant to Cuda accelerators. Even the naming is Cuda-specific `CU_STREAM_DEFAULT` and `CU_STREAM_NON_BLOCKING`.

If we wanted to expose `StreamFlags`, I was thinking to add the method `public CudaStream CreateStream(StreamFlags)` to the `CudaAccelerator` class.

Src/ILGPU/Runtime/Accelerator.cs
```diff
@@ -407,12 +408,46 @@ public AcceleratorStream CreateStream()
...
         }
 
+        /// <summary>
+        /// Creates a new accelerator stream from an externally created one.
+        /// </summary>
+        /// <param name="ptr">The pointer to the externally created stream.</param>
+        /// <returns>The created accelerator stream.</returns>
+        public AcceleratorStream CreateStream(IntPtr ptr)
+        {
+            Bind(); return CreateStreamInternal();
```
MoFtZ, line 418: ISSUE: `ptr` is not passed to `CreateStreamInternal`.

Src/ILGPU/Runtime/Cuda/CudaNativeMethods.cs
```diff
@@ -12,11 +12,14 @@
...
 #pragma warning disable CS1591 // Missing XML comment for publicly visible type or member
 
+using System.Diagnostics.CodeAnalysis;
+
 namespace ILGPU.Runtime.Cuda
 {
     #region Enums
 
-    enum StreamFlags
+    [SuppressMessage("Naming", "CA1717")]
```
MoFtZ, line 21: ISSUE: Should use the full supression `[SuppressMessage("Naming", "CA1717:Only FlagsAttribute enums should have plural names")]` or maybe add the `[Flags]` attribute instead?

Src/ILGPU/Runtime/OpenCL/CLAccelerator.cs
```diff
@@ -526,10 +527,19 @@ protected override CLKernel CreateKernel(CLCompiledKernel compiledKernel)
             MethodInfo launcher) =>
             new CLKernel(this, compiledKernel, launcher);
 
-        /// <summary cref="Accelerator.CreateStream"/>
+        /// <summary cref="Accelerator.CreateStream()"/>
         protected override AcceleratorStream CreateStreamInternal() =>
             new CLStream(this);
 
+        /// <summary cref="Accelerator.CreateStream(StreamFlags)"/>
+        protected override AcceleratorStream CreateStreamInternal(StreamFlags flag) =>
```
MoFtZ, line 535: ISSUE: Yeah, the more I think about it, let's not add support for `StreamFlags` on all accelerators - just CudaAccelerator. Then we dont have to import the `ILGPU.Runtime.Cuda` namespace.

MoFtZ:
OK, so my thoughts are that we should only add `CreateStream(StreamFlags)` to the `CudaAccelerator` rather than all accelerators. And we should add `CreateStream(IntPtr)` to the `CudaAccelerator` and `CLAccelerator` individually rather than to the base class `Accelerator`.

My reasoning is that these are advanced methods that lock you in to a specific technology. So, if the caller needs to supply an externally created stream, they should be explicitly casting to `CudaAccelerator` or `CLAccelerator` first. It also avoids the problem with the `CPUAccelerator` throwing exceptions when `CreateStream(StreamFlags)` or `CreateStream(IntPtr)` are called - the base `Accelerator` class should provide an API that is compatible for all accelerators.

What are your thoughts on this?

samy2:
@MoFtZ , yes, it makes sense.

Src/ILGPU/Runtime/Cuda/CudaNativeMethods.cs
Yey007: Didn't know that was an attribute. I'll get right to it.

Src/ILGPU/Runtime/Cuda/CudaStream.cs
Yey007: Yup, that makes sense. Since the constructor is internal, this also wouldn't be a breaking API change.

Src/ILGPU/Runtime/Cuda/CudaNativeMethods.cs
Yey007: I can't seem to find the flags attribute. Was it recently added? This branch is behind quite a few commits as I forgot to pull before starting...

m4rs-mt:
@MoFtZ I am completely satisfied with this design decision. As you said, these functions are closely related to the underlying technology and should not be generalized to all accelerators by default.

MoFtZ:
@Yey007 @m4rs-mt 
As we discussed on the Discord channel, the `CreateStream(IntPtr)` should have an extra boolean parameter to indicate whether ILGPU is responsible for disposing the handle.

Src/ILGPU/Runtime/Cuda/CudaAccelerator.cs
```diff
@@ -260,7 +260,7 @@ private void SetupAccelerator()
             CudaException.ThrowIfFailed(
                 CurrentAPI.GetDeviceName(out string name, DeviceId));
             Name = name;
-            DefaultStream = new CudaStream(this, IntPtr.Zero);
+            DefaultStream = new CudaStream(this, StreamFlags.CU_STREAM_NON_BLOCKING);
```
MoFtZ, line 263: ISSUE: The default Cuda stream is supposed to use Stream #0.
```suggestion
            DefaultStream = new CudaStream(this, IntPtr.Zero, responsible: false);
```

Src/ILGPU/Runtime/Cuda/CudaStream.cs
```diff
@@ -83,13 +91,16 @@ public override void Synchronize()
...
-                CudaException.ThrowIfFailed(
-                    CudaAPI.Current.DestroyStream(streamPtr));
-                streamPtr = IntPtr.Zero;
+                if (streamPtr != IntPtr.Zero)
+                {
+                    CudaException.ThrowIfFailed(
+                        CudaAPI.Current.DestroyStream(streamPtr));
+                    streamPtr = IntPtr.Zero;
+                }
+                base.Dispose(disposing);
```
MoFtZ, line 102: ISSUE: We should always call `base.Dispose(disposing);`, even when we are not responsible for disposing the stream.

This is because the base class still needs to cleanup after itself.

Src/ILGPU/Runtime/OpenCL/CLAccelerator.cs
```diff
@@ -12,6 +12,7 @@
 using ILGPU.Backends.IL;
 using ILGPU.Backends.OpenCL;
 using ILGPU.Resources;
+using ILGPU.Runtime.Cuda;
```
MoFtZ, line 15: ISSUE: Should not include `using ILGPU.Runtime.Cuda` in OpenCL code.

Src/ILGPU/Runtime/OpenCL/CLStream.cs
```diff
@@ -28,6 +28,12 @@ public sealed class CLStream : AcceleratorStream
 
         private IntPtr queuePtr;
 
+        internal CLStream(Accelerator accelerator, IntPtr ptr)
```
MoFtZ, line 31: ISSUE: `CLStream` also needs the `responsible` boolean parameter, like `CudaStream`.

Yey007: Oops, forgot that.

Src/ILGPU/Runtime/Cuda/CudaAccelerator.cs
Yey007: shouldn't it be responsible of disposing itself?

MoFtZ: ILGPU does not actually create a default stream in Cuda. It is passing 0 as the stream, and Cuda interprets this as the default stream. Also, the existing dispose code checks for `streamPtr != IntPtr.Zero`, so it doesn't really mater what value we set for `responsible`.

Src/ILGPU/Runtime/OpenCL/CLStream.cs
```diff
@@ -27,6 +27,14 @@ public sealed class CLStream : AcceleratorStream
...
+        private readonly bool responsibleForHandle;
+
+        internal CLStream(Accelerator accelerator, IntPtr ptr, bool responsible)
+            : base(accelerator)
+        {
+            queuePtr = ptr;
+            responsibleForHandle = responsible;
+        }
 
         internal CLStream(CLAccelerator accelerator)
```
MoFtZ, line 39: ISSUE: This constructor should set `responsibleForHandle` to true.

Src/ILGPU/Runtime/Cuda/CudaAccelerator.cs
Yey007: I see. Makes sense.

Src/ILGPU/Runtime/Cuda/CudaStream.cs
```diff
@@ -83,11 +91,14 @@ public override void Synchronize()
         /// <summary cref="DisposeBase.Dispose(bool)"/>
         protected override void Dispose(bool disposing)
         {
-            if (streamPtr != IntPtr.Zero)
+            if (responsibleForHandle)
```
m4rs-mt, line 94: nit: can we fold this into an and-also conditional (`&&`)?

Src/ILGPU/Runtime/Cuda/CudaNativeMethods.cs
```diff
@@ -12,11 +12,14 @@
...
 
+using System;
+
 namespace ILGPU.Runtime.Cuda
 {
     #region Enums
 
-    enum StreamFlags
+    [Flags]
+    public enum StreamFlags
```
m4rs-mt, line 22: nit: Maybe we should add some comments to this type.

Src/ILGPU/Runtime/OpenCL/CLStream.cs
```diff
@@ -63,11 +72,14 @@ internal CLStream(CLAccelerator accelerator)
         /// <summary cref="DisposeBase.Dispose(bool)"/>
         protected override void Dispose(bool disposing)
         {
-            if (queuePtr != IntPtr.Zero)
+            if (responsibleForHandle)
```
m4rs-mt, line 75: nit: same here

Src/ILGPU/Runtime/Cuda/CudaNativeMethods.cs
```diff
@@ -12,11 +12,22 @@
...
+    /// <para><see cref="CU_STREAM_DEFAULT"/> specifies a blocking stream</para>
+    /// <para>
+    /// <see cref="CU_STREAM_NON_BLOCKING"/> specifies a non-blocking stream.
+    /// It is used by default
+    /// </para>
+    /// </summary>
+    [Flags]
+    public enum StreamFlags
     {
         CU_STREAM_DEFAULT = 0,
```
m4rs-mt, line 32: We should add some comments to the enumeration members as they have become publicly visible.
Maybe we can just convert the comments inside the `<para>...` regions and apply them to the individual enumeration members.

Src/ILGPU/Runtime/Cuda/CudaStream.cs
```diff
@@ -27,29 +27,37 @@ public sealed class CudaStream : AcceleratorStream
...
+        internal CudaStream(Accelerator accelerator, IntPtr ptr, bool responsible)
             : base(accelerator)
         {
             streamPtr = ptr;
+            responsibleForHandle = responsible;
         }
 
         /// <summary>
-        /// Constructs a new cuda stream.
+        /// Constructs a new cuda stream with given <see cref="StreamFlags"/>
```
m4rs-mt, line 48: nit: missing `.`

Src/ILGPU/Runtime/Cuda/CudaStream.cs
```diff
@@ -27,29 +27,37 @@ public sealed class CudaStream : AcceleratorStream
...
+        /// Stream flag to use. Allows blocking and non-blocking streams.
+        /// </param>
+        internal CudaStream(Accelerator accelerator, StreamFlags flag)
             : base(accelerator)
         {
             CudaException.ThrowIfFailed(
                 CudaAPI.Current.CreateStream(
-                    out streamPtr,
-                    StreamFlags.CU_STREAM_NON_BLOCKING));
+                    out streamPtr, flag));
```
m4rs-mt, line 59: nit: we should break the line after the `,` in this case.

Src/ILGPU/Runtime/OpenCL/CLAccelerator.cs
```diff
@@ -526,10 +526,22 @@ protected override CLKernel CreateKernel(CLCompiledKernel compiledKernel)
...
             new CLKernel(this, compiledKernel, launcher);
 
-        /// <summary cref="Accelerator.CreateStream"/>
+        /// <summary cref="Accelerator.CreateStream()"/>
         protected override AcceleratorStream CreateStreamInternal() =>
             new CLStream(this);
 
+        /// <summary>
+        /// Creates a <see cref="CLStream"/> object from an externally
+        /// created stream/queue using its pointer
```
m4rs-mt, line 535: nit: missing `.`

Src/ILGPU/Runtime/OpenCL/CLAccelerator.cs
```diff
@@ -526,10 +526,22 @@ protected override CLKernel CreateKernel(CLCompiledKernel compiledKernel)
...
+        /// </summary>
+        /// <param name="ptr">The pointer to use while creating the new stream</param>
```
m4rs-mt, line 537: nit: missing `.`

Src/ILGPU/Runtime/OpenCL/CLAccelerator.cs
```diff
@@ -526,10 +526,22 @@ protected override CLKernel CreateKernel(CLCompiledKernel compiledKernel)
...
+        /// <param name="responsible">
+        /// Whether ILGPU is responsible of disposing this stream.
+        /// </param>
+        /// <returns>The created stream</returns>
```
m4rs-mt, line 541: nit: missing `.`

m4rs-mt (changes requested):
Great work 🚀 After addressing all the nits this should be good to go 🤓

m4rs-mt (approved):

PR #227, opened 2020-08-06
Fixed ability to override registration of Tanh PTX intrinsic

MoFtZ:
In https://github.com/m4rs-mt/ILGPU/pull/187, we implemented support for the new Tanh PTX intrinsic. To prevent older Cuda architectures from incorrectly using this intrinsic, we also registered a custom intrinsic which throws `NotSupportedException`.

When we integrate with ILGPU.Algorithms, calling `EnableAlgorithms` will also register a custom intrinsic for Tanh. However, there is no specific rule for choosing which intrinsic to use.

This PR defines that the last registered intrinsic implementation is the preferred implementation.

A PR will also need to be raised in ILGPU.Algorithms to avoid registering the Tanh intrinsic if not required.


m4rs-mt:
I found the same problem yesterday and was about to implement it in the same way as you 😄

PR #228, opened 2020-08-06
Fixed CL grid dimensions returning wrong values

### LINKED ISSUE #223: Invalid Grid.DimXYZ dimensions in OpenCL
Querying the `Grid` dimensions in OpenCL kernels yields the total number of threads in the scheduled thread grid. According to the behavior of the `Grid.DimXYZ` property implementations on all other platforms, these properties should return the number of scheduled grid blocks instead of the total number of dispatched threads.

Yey007:
This closes #223 . At the moment, it's sort of janky and I would like everyone's advice on how this can be better accomplished. Essentially, the issue was that OpenCL's `get_global_size` returns the number of total threads launched across all blocks, not the number of blocks. In order to fix this, I just divided by the number of threads per block. Because of this, the method that generates `Grid` now looks kind of messy, as I couldn't find any existing functionality to do what is necessary. I'm not sure exactly how to refactor this so it's cleaner, so your input would be appreciated. I have tested it and it does work for single- and multi-dimensional data. I also want to thank @MoFtZ for helping me tremendously with understanding the backend and helping solve this issue.

### DIFF
--- a/Src/ILGPU.Tests/GridOperations.cs
+++ b/Src/ILGPU.Tests/GridOperations.cs
@@ -1,6 +1,4 @@
-﻿using ILGPU.IR.Values;
-using ILGPU.Runtime;
-using ILGPU.Runtime.OpenCL;
+﻿using ILGPU.Runtime;
 using System;
 using System.Diagnostics;
 using Xunit;
--- a/Src/ILGPU.Tests/GridOperations.cs
+++ b/Src/ILGPU.Tests/GridOperations.cs
@@ -58,27 +58,25 @@ namespace ILGPU.Tests
         internal static void GridLaunchDimensionKernel(ArrayView<int> data)
         {
             data[0] = Grid.DimX;
-            data[1] = Grid.DimY;
-            data[2] = Grid.DimZ;
         }
 
+        // This test is one-dimensional and uses small sizes for the sake of passing
+        // tests on the CI machine, but on a machine with more threads it works
+        // for higher dimensions and higher sizes.
         [Fact]
         public void GridLaunchDimension()
         {
-            using var buffer = Accelerator.Allocate<int>(3);
+            using var buffer = Accelerator.Allocate<int>(1);
             var kernel = Accelerator.LoadStreamKernel<ArrayView<int>>
                 (GridLaunchDimensionKernel);
 
-            kernel((new Index3(1, 2, 3), new Index3(4, 5, 6)), buffer.View);
+            kernel((1, 2), buffer.View); 
             Accelerator.Synchronize();
 
             var data = buffer.GetAsArray();
-            int[] expected = { 1, 2, 3 };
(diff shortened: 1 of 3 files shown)


Src/ILGPU/Backends/OpenCL/CLCodeGenerator.Values.cs
```diff
@@ -510,10 +510,7 @@ public void GenerateCode(SetField value)
 
         /// <summary cref="IBackendCodeGenerator.GenerateCode(GridDimensionValue)"/>
         public void GenerateCode(GridDimensionValue value) =>
-            MakeIntrinsicValue(
```
MoFtZ, line 513: MINOR: In the end, no change was necessary here, so revert the changes.

Src/ILGPU.Tests/GridOperations.cs
```diff
@@ -51,5 +52,29 @@ public void GridDimension(int xMask, int yMask, int zMask)
...
+        // tests on the CI machine, but on a machine with more threads it works
+        // for higher dimensions and higher sizes.
+        [Fact]
+        public void GridLaunchDimension()
+        {
+            using var buffer = Accelerator.Allocate<int>(1);
+            var kernel = Accelerator.LoadStreamKernel<ArrayView<int>>
+                (GridLaunchDimensionKernel);
+
+            kernel((1, 2), buffer.View);
```
MoFtZ, line 71: ISSUE: Actually, I know I said it was fine, but on second thought, this is going to pass on OpenCL even with the old/broken code.

The original issue was that launching (10, 256) would cause `Grid.DimX` to return 2560 instead of 256.
In this case, launching (1, 2) would incorrectly return 2, regardless of whether the code was fixed or not.

Yey007: Why is this the case? Am I missing something here?

Yey007: Isn't the first number the Grid dimension? When we discussed, I believe we said the problem was it was returning 2560 instead of 10.

MoFtZ: Oops, you're right. The old code would return 2. The fixed code will return 1.
Don't mind me...

Src/ILGPU/Backends/OpenCL/CLCodeGenerator.Values.cs
```diff
@@ -510,10 +510,7 @@ public void GenerateCode(SetField value)
...
-                value,
-                CLInstructions.GetGridSize,
-                value.Dimension);
+            MakeIntrinsicValue(value, CLInstructions.GetGridSize, value.Dimension);
```
m4rs-mt, line 513: nit: can we revert this piece of code since there are no real changes?

Issue #231, opened 2020-08-07
a problem about the SimpleStructures samples


huiyiyuxin:
Thank you for the powerful library. 
I Meet with a problem.
 when I run the "SimpleStructures" samples, I get such errors. 
How could I solve it.
Best Wishes.
![image](https://user-images.githubusercontent.com/28803563/89603503-de571180-d89b-11ea-9c8b-cf6fc78d0a54.png)

MoFtZ:
hi @huiyiyuxin. The sample code will try to run on each Accelerator on your computer. Do you know which accelerator is failing? Also, are all the accelerators failing or just some of them?

There is a CPU accelerator, which uses your CPU to emulate the GPU.
The Cuda accelerator is used by Nvidia graphics cards.
The OpenCL accelerator is used by AMD graphics cards, and some CPUs also provide support for OpenCL.

It might be a good idea to ensure you have the latest drivers.
From my experience, the OpenCL accelerator for my CPU had a number of issues before updating to the latest driver. Also, it only provides limited support for all the features of ILGPU.

m4rs-mt:
hi @huiyiyuxin. This error message can be related to the fact that the ILGPU runtime library cannot access the `CustomDataType` in the project. We have added the assembly attribute `[InternalsVisibleTo(...)]` to the sample projects in order to make all internal types visible to the runtime library. Is it possible that in your case the attribute was not compiled correctly/not added to the project?

huiyiyuxin:
Dear m4rs-mt and MoFtZ， Thank you  for your help, I added the assembly attribute to the projects, and it can work at CPU. Best Wishes

PR #235, opened 2020-08-11
Ensure TypeLoadExceptions inform the caller of the visibility issue.

MoFtZ:
Provide assistance to resolve common issue with visibility of types (https://github.com/m4rs-mt/ILGPU/issues/231).

### DIFF
--- a/Src/ILGPU/Backends/Backend.cs
+++ b/Src/ILGPU/Backends/Backend.cs
@@ -650,24 +650,48 @@ namespace ILGPU.Backends
             TBackendHook backendHook)
             where TBackendHook : IBackendHook
         {
-            using var kernelContext = new IRContext(Context);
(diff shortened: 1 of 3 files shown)


Src/ILGPU/Backends/Backend.cs
```diff
@@ -650,24 +650,48 @@ public static void EnsureRunningOnPlatform(TargetPlatform platform)
...
+                    tle);
+            }
+            catch (InternalCompilerException)
+            {
+                // If we already have an internal compiler exception, re-throw it.
+                throw;
+            }
+            catch (Exception e)
+            {
+                // Wrap generic exceptions with location information.
```
MoFtZ, line 690:
```suggestion
                // Wrap generic exceptions.
```

PR #245, opened 2020-08-18
Fixed CLAccelerator.Synchronize() only synchronizing the default stream

### LINKED ISSUE #240: Invalid Accelerator synchronization in OpenCL
The current implementation of `CLAccelerator.Synchronize` synchronizes the default accelerator stream only. However, the intended behavior would be to synchronize all accelerator streams (like in the case of the `CPUAccelerator` and the `CudaAccelerator`).

Steps to reproduce:
```c#
var stream = Accelerator.CreateStream();

// ...

kernel(stream);

// ...

Accelerator.Synchronize(); // Will not be synchronized here
```

Yey007:
Fixes #240


Yey007:
Ok, so unfortunately I messed the pr up so I'm going to close this and open a new one.

PR #246, opened 2020-08-18
Fixed CLAccelerator.Synchronize() not synchronizing all streams

### LINKED ISSUE #240: Invalid Accelerator synchronization in OpenCL

Yey007:
Fixes #240


Yey007:
I don't know how to fix this at this point, could anyone point me in the right direction? I just don't understand git sometimes...

Src/ILGPU/Runtime/OpenCL/CLAccelerator.cs
```diff
@@ -527,12 +527,25 @@ protected override CLKernel CreateKernel(CLCompiledKernel compiledKernel)
             new CLKernel(this, compiledKernel, launcher);
 
         /// <summary cref="Accelerator.CreateStream"/>
-        protected override AcceleratorStream CreateStreamInternal() =>
-            new CLStream(this);
+        protected override AcceleratorStream CreateStreamInternal()
+        {
+            var stream = new CLStream(this);
+            ChildObjects.Add(new WeakReference<AcceleratorObject>(stream));
```
m4rs-mt, line 533: ISSUE: You should not add this reference to the child list since it will be added automatically to the list in its own constructor.

Src/ILGPU/Runtime/OpenCL/CLAccelerator.cs
```diff
@@ -527,12 +527,25 @@ protected override CLKernel CreateKernel(CLCompiledKernel compiledKernel)
...
+            return stream;
+        }
 
         /// <summary cref="Accelerator.Synchronize"/>
-        protected override void SynchronizeInternal() =>
-            DefaultStream.Synchronize();
+        protected override void SynchronizeInternal()
+        {
+            for(int i = 0; i < ChildObjects.Count; i++)
```
m4rs-mt, line 540: ISSUE: Concurrency; this will not work as it is not using the internal synchronization lock.

Src/ILGPU/Runtime/AcceleratorObject.cs
```diff
@@ -105,8 +105,8 @@ partial class Accelerator
         /// The list of linked child objects.
         /// </summary>
         [DebuggerBrowsable(DebuggerBrowsableState.Never)]
-        private List<WeakReference<AcceleratorObject>> childObjects =
-            new List<WeakReference<AcceleratorObject>>();
+        protected List<WeakReference<AcceleratorObject>> ChildObjects
```
m4rs-mt, line 108: We should not expose this private property to anyone I guess; maybe we should define a thread-safe function to iterate over all instances of a specific type and invoke a user-defined callback; something like this:
````c#
protected void ForEachChildObject<T>(Action<T> callback) where T : AcceleratorObject
{
    lock (syncRoot)
    {
         ...
    }
}
```

m4rs-mt (changes requested):

m4rs-mt:
> I don't know how to fix this at this point; I'm trying to remove GridOperations.cs which I didn't modify. Could anyone point me in the right direction? I just don't understand git sometimes...

It looks like you have accidentally modified the `GridOperations.cs` file by changing whitespaces.

Src/ILGPU/Runtime/OpenCL/CLAccelerator.cs
Yey007: When I created a stream, it didn't add itself to the list. I compared pointers, and the only stream in the list was the default stream. Is this potentially another bug?

Src/ILGPU/Runtime/OpenCL/CLAccelerator.cs
Yey007: I see. I didn't understand that the GC was asynchronous, although I don't see why it would be otherwise anyways.

Src/ILGPU/Runtime/OpenCL/CLAccelerator.cs
Yey007: Well, whatever I did, I must have done it wrong. They're in the list now...

Src/ILGPU/Runtime/AcceleratorObject.cs
```diff
@@ -165,6 +165,27 @@ internal void RegisterChildObject<T>(T child)
...
+        /// <typeparam name="T">The type of child object</typeparam>
+        /// <param name="callback">The action to perform on the object</param>
+        protected void ForEachChildObject<T>(Action<T> callback)
+            where T : AcceleratorObject
+        {
+            lock (syncRoot)
+            {
+                foreach(var child in childObjects)
+                {
+                    bool success = child.TryGetTarget(out var acceleratorObject);
```
m4rs-mt, line 180: nit: let's merge the `child.TryGetTarget` expression into the `if` statement.

Src/ILGPU.Tests/GridOperations.cs
```diff
@@ -12,7 +12,6 @@ public abstract class GridOperations : TestBase
         protected GridOperations(ITestOutputHelper output, TestContext testContext)
             : base(output, testContext)
         { }
-
```
m4rs-mt, line 15: nit: accidentally removed a line?

Yey007: In the last PR, I pulled this file from the master branch but it still refuses to believe I didn't touch it...

Yey007: In order to fix this, what should I do? I can close the PR and push the branch again if necessary.

Yey007:
Ah, it finally says I didn't modify GridOperations.cs. I must have messed up somewhere. Also don't know why that added three commits but whatever.

m4rs-mt (approved):

MoFtZ:
DISCUSSION: I recently found these two OpenCL functions: [clEnqueueMarkerWithWaitList](https://www.khronos.org/registry/OpenCL/sdk/2.0/docs/man/xhtml/clEnqueueMarkerWithWaitList.html) and [clEnqueueBarrierWithWaitList](https://www.khronos.org/registry/OpenCL/sdk/2.0/docs/man/xhtml/clEnqueueBarrierWithWaitList.html).

Instead of calling `CLStream.Synchronize`, maybe we can add a marker to each stream, and then use [clWaitForEvents](https://www.khronos.org/registry/OpenCL/sdk/2.0/docs/man/xhtml/clWaitForEvents.html) to wait for all the streams at the same time.

Using these markers avoids a race condition with calling `CLStream.Synchronize`. For example, if we had two queues - while synchronizing on the first queue, it is possible for more commands to be added to the second queue. Then, when we synchronize on the next second queue, we would be waiting on these extra commands to finish.

I'm thinking `clEnqueueBarrierWithWaitList` more closely matches the CUDA behaviour.

Thoughts?

Yey007:
Adds another to the list of useful CL functions you've found 😃. Yeah, this sounds like a good idea, but maybe I am misunderstanding you or the OpenCL documentation: how does this solve the race condition? Isn't it still possible for a command to be added to a queue while we're synchronizing?

MoFtZ:
> how does this solve the race condition? Isn't it still possible for a command to be added to a queue while we're synchronizing?

If we just call `CLStream.Synchronize`, we have to wait for the first queue to finish before we even synchronize the second queue, If the first queue takes a long time to finish, then there is a long time for extra commands to be added to the second queue.

Instead, when we call `Accelerator.Synchronize`, we can iterate over each CLStream add queue a barrier marker. Then, as a final step, wait on all the marker events. This helps to solve the race condition because any newly queued commands will be blocked by the barrier, and there is a shorter time to wait between adding a marker to the first and second queues.

NB: There is still a slight chance that a command will be queued on the second queue before we can add the barrier marker.

Yey007:
Alright makes sense. I think that chance is a chance we're gonna have to take. We can probably add a comment saying something along the lines of "never try to add another operation (from another thread) while accelerator synchronization is running".

I'll get started on this, unless someone else wanted to do it.

Yey007:
One question...
When you referred to adding a "marker", you meant the event returned from `clEnqueueBarrierWithWaitList`, right? I just want to make sure we're not thinking of different implementations.

Yey007:
I also don't quite understand the difference between `clEnqueueBarrierWithWaitList` and  `clWaitForEvents`. They sound a lot like the same thing to me, don't they both just wait for some events to complete?

Yey007:
I don't know if this is what you're thinking, but we may be able to utilize a feature of `clEnqueueBarrierWithWaitList` to our advantage:

> If event_wait_list is NULL, then this particular command waits until all previous enqueued commands to command_queue have completed.

MoFtZ:
Think of it this way - we have command queues, and we can add commands like 'run kernel' and 'copy memory' to that queue. Each command that we add returns an 'event' object, that can be used to track when the command has finished running. Also, when adding a command, we can optionally specify that the command must wait for some events to complete before it can run.

You can think of a marker as similar to a command - it gets added to the command queue and returns an event. These markers just happen to have additional functionality.

If you use `clEnqueueMarkerWithWaitList` and supply some events for it to wait, it will act as a no-op command. But if you supply NULL, it will wait for all previously queued commands - without you having to explicitly track it.

`clEnqueueBarrierWithWaitList` differs to `clEnqueueMarkerWithWaitList` by blocking other commands - "This command blocks command execution, that is, any following commands enqueued after it do not execute until it completes."

MoFtZ:
ah sorry, didn't see your reply 😄 

yes, that was what I was thinking - using the NULL feature so that we do not have to track previous events, and also using the returned event from each stream so that we can call clWaitForEvents on a single list.

> 
> 
> I don't know if this is what you're thinking, but we may be able to utilize a feature of `clEnqueueBarrierWithWaitList` to our advantage:
> 
> > If event_wait_list is NULL, then this particular command waits until all previous enqueued commands to command_queue have completed.

MoFtZ:
@Yey007 I have updated #261 with some methods that might help this PR.

Yey007:
Sorry, I didn't see your comments before pushing. We partially wrote the same thing, although you definitely wrote it better😃 . How should we go about this? Should I just wait for your PR to be accepted?

Yey007:
Whoa, yeah. There are a _a lot_ of safety checks missing in my implementation. I don't even know how it worked.

MoFtZ:
> How should we go about this? Should I just wait for your PR to be accepted?

If you copy/paste my changes in `CLAPI.cs` and `NativeMethods.cs`, then there should be no merge conflicts, and either of our PRs can be merged.

Src/ILGPU/Runtime/OpenCL/CLAccelerator.cs
```diff
@@ -543,8 +546,31 @@ protected override CLKernel CreateKernel(CLCompiledKernel compiledKernel)
...
+            ForEachChildObject((CLStream obj) =>
+            {
+                // Setting the wait list count to 0 and the wait list to null
+                // will wait for all operations enqueued prior to this one
+                CLAPI.BarrierWithWaitList(obj.CommandQueue, 0, null, arr[index]);
+                index++;
+            });
+            // We use all the events we obtained from BarrierWithWaitList to wait for all
+            // queued operations to finish
+            CLAPI.WaitForEvents(arr.Length, arr[0]);
```
MoFtZ, line 572: SUGGESTION: You can do this in one pass and maintain a list outside the `ForEachChildObject` loop.

Regarding the IntPtr array, you can use `stackalloc` to allocate a low-cost buffer.

```suggestion
            var streamEvents = new List<IntPtr>();
            try
            {
                ForEachChildObject<CLStream>(stream =>
                {
                    IntPtr* resultEvent = stackalloc IntPtr[1];
                    CLException.ThrowIfFailed(
                        CLAPI.EnqueueBarrierWithWaitList(
                            stream.CommandQueue,
                            Array.Empty<IntPtr>(),
                            resultEvent));
                    streamEvents.Add(*resultEvent);
                });
                
                if (streamEvents.Count > 0)
                    CLException.ThrowIfFailed(CLAPI.WaitForEvents(streamEvents.ToArray()));
            }
            finally
            {
                foreach (var streamEvent in streamEvents)
                    CLAPI.ReleaseEvent(streamEvent);
            }
```

Yey007: Good idea. What I was struggling with originally was maintaining a list of `IntPtr*`, which C# wouldn't let me do. I don't know why I didn't think of just dereferencing the pointers.

Yey007: One question though. Is there a difference in passing `Array.Empty` or null to the API? The documentation said it should be null, that's why I'm curious.

Src/ILGPU/Runtime/OpenCL/API/CLAPI.cs
```diff
@@ -759,8 +759,10 @@ public static CLError CreateContext(IntPtr device, out IntPtr context)
             int workDimensions,
             IntPtr* workOffsets,
             IntPtr* globalWorkSizes,
-            IntPtr* localWorkSizes) =>
```
MoFtZ, line 762: ISSUE: Sorry, wasn't clear - you only needed to copy/paste the changes for adding the new APIs, not the changes for EnqueueBarrier, which is to fix a different bug.

MoFtZ (changes requested):

Src/ILGPU/Runtime/OpenCL/CLAccelerator.cs
MoFtZ: Not much difference. When you use the `fixed` statement on `Array.Empty`, it returns null - so that satisfies the OpenCL API.

Yey007: I see. `Array.Empty` seems good then.

Src/ILGPU/Runtime/OpenCL/API/CLAPI.cs
Yey007: I'll fix that right away.

Src/ILGPU/Runtime/OpenCL/CLAccelerator.cs
```diff
@@ -543,8 +544,41 @@ protected override CLKernel CreateKernel(CLCompiledKernel compiledKernel)
...
+                    // Dereference pointer so we can store it.
+                    streamEvents.Add(*resultEvent);
+                });
+
+                // Wait for all the events to fire, which would mean
+                // all operations queued on an accelerator prior to
+                // synchronization have finished.
+                if (streamEvents.Count > 0)
+                    CLException.ThrowIfFailed(CLAPI.WaitForEvents(
+                        streamEvents.ToArray()));
```
MoFtZ, line 573: MINOR: Our EditorConfig is saying we should add braces here.
```suggestion
                if (streamEvents.Count > 0)
                {
                    CLException.ThrowIfFailed(CLAPI.WaitForEvents(
                        streamEvents.ToArray()));
                }
```

Src/ILGPU/Runtime/OpenCL/CLAccelerator.cs
```diff
@@ -543,8 +544,41 @@ protected override CLKernel CreateKernel(CLCompiledKernel compiledKernel)
...
+            }
+            finally
+            {
+                // Clean up the events we made.
+                foreach (var streamEvent in streamEvents)
+                    CLAPI.ReleaseEvent(streamEvent);
```
MoFtZ, line 579: ISSUE: Should probably call `CLException.ThrowIfFailed` here.

MoFtZ (changes requested):

m4rs-mt:
ISSUE: the current state of this PR seems to fail the `CLExchangeBufferOperations` tests on AMD and Intel platforms.

m4rs-mt (changes requested):
I am not sure why this implementation fails to pass the `CLExchangeBufferOperation` tests with the error message `CL_INVALID_COMMAND_QUEUE`.

MoFtZ:
@m4rs-mt @Yey007 
In the unit test, when we run the kernel, we call `TestBase:Execute`, and the implementation of this method creates a [new stream](https://github.com/m4rs-mt/ILGPU/blob/9e6fbff188c6127a5b906aecb354a2dc1d7c6f55/Src/ILGPU.Tests/Generic/TestBase.cs#L107).

The unit test is failing because at the end of the `Execute` method, the stream is being disposed. This sets the command queue pointer to IntPtr.Zero, but the CLStream instance is still registered with the Accelerator. Then, when we call `Accelerator.Synchronize`, it attempts to enqueue a barrier command onto an invalid command queue pointer (`IntPtr.Zero`).

m4rs-mt:
@MoFtZ I guess an easy `IntPtr.Zero` check should solve this issue 🤓

MoFtZ:
> 
> 
> @MoFtZ I guess an easy `IntPtr.Zero` check should solve this issue 🤓

@m4rs-mt That may not be enough. The CLStream was disposed because we hold a weak reference to it.

When `Accelerator.Synchronize` is called, and it is looping through the streams, we might need to keep a list of strong references to CLStream instances. Otherwise, while waiting for `clWaitForEvents` to finish, a CLStream instance might be disposed prematurely.

m4rs-mt:
@MoFtZ Yeah, you are right. We have to maintain another list of strong references to avoid automatic disposal of stream instances while we are trying to wait for the events to finish.

m4rs-mt:
I polished and manually merged this PR.

PR #249, opened 2020-08-22
New Native DLL-Interop API

m4rs-mt:
This PR adds a new DLL-interop API that allows us to create platform and DLL-specific implementation wrappers. This simplifies native library loading and makes the code easier to understand.


MoFtZ (approved):
QUESTION: On the original Cuda Windows API, we implemented `CreateContextD3D11` by calling `cuD3D11CtxCreate_v2`. And on Unix, this was implemented by throwing an exception.

How do we deal with that with this new code? I'm assuming the Unix version of the DLL will not have a binding for `cuD3D11CtxCreate_v2`.

m4rs-mt:
@MoFtZ Yeah, that's right. Maybe we could extend the DLL loader so that we can mark certain functions so that they are only available on certain platforms?

MoFtZ:
I'm wondering whether for something like this, we should wait for C# Source Generators, or even just use T4 templates.

The benefit I see of generating a dynamic type is for use by CuBlas, where the DLL name changes on each new version, and it would be nice for the user to be able to specify their own DLL name at runtime (e.g. a new version of CuBlas is released, but we have not updated ILGPU.Algorithms to support that version).

PR #259, opened 2020-09-09
Added new Loop Unrolling capabilities

### LINKED ISSUE #251: Loop Unrolling
ILGPU currently does not support loop unrolling in any form. We should add a new loop-unrolling transformation to aggressively unroll certain loops to improve performance. On the one hand, this will increase the overall size of the code, but on the other hand it will also significantly increase the potential for optimization (similar to inlining).

m4rs-mt:
This PR adds all required analysis and transformations to perform full and partial loop unrolling (closes #251).

### DIFF
--- a/Src/ILGPU/IR/Analyses/LoopInfo.cs
+++ b/Src/ILGPU/IR/Analyses/LoopInfo.cs
@@ -13,6 +13,7 @@ using ILGPU.IR.Analyses.ControlFlowDirection;
 using ILGPU.IR.Analyses.TraversalOrders;
 using ILGPU.IR.Values;
 using ILGPU.Util;
+using System;
 using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Diagnostics;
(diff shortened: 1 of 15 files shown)


Src/ILGPU/IR/Values/Terminators.cs
```diff
@@ -362,6 +387,21 @@ protected ConditionalBranch(in ValueInitializer initializer)
...
+        /// </summary>
+        /// <param name="block">The branch target to exclude.</param>
+        /// <param name="otherBlock">The other branch target (if any).</param>
+        public bool TryGetOtherBranchTarget(BasicBlock block, out BasicBlock otherBlock)
+        {
+            otherBlock = null;
+            if (NumTargets > 2)
+                return false;
+            otherBlock = Targets[0] == block ? Targets[1] : Targets[0];
+            return true;
```
MoFtZ, line 402: QUESTION: Is this correct? The description says get the other branch if `block` is one of the targets.
```suggestion
            if (Targets[0] == block)
            {
                otherBlock = Targets[1];
                return true;
            }
            else if (Targets[1] == block)
            {
                otherBlock = Targets[0];
                return true;
            }
            return false;
```

Src/ILGPU/IR/Analyses/LoopInfo.cs
```diff
@@ -344,6 +493,22 @@ internal InductionVariableOperation(int index, Value value, T kind)
...
+
+        /// <summary>
+        /// Tries to map a loop variable to an integer constant.
+        /// </summary>
+        /// <param name="value">The value to map to an integer bound.</param>
+        /// <returns>The mapped integer (if any).</returns>
+        private static int? TryGetIntegerBound(Value value) =>
+            value is PrimitiveValue primitive && value.BasicValueType.IsInt()
+            ? primitive.Int32Value
+            : (int?)null;
```
MoFtZ, line 506: SUGGESTION: `default` is the equivalent of `(int?)null` in this instance.
```suggestion
            : default;
```

Src/ILGPU/IR/Transformations/ConstantPropagation.cs
```diff
@@ -0,0 +1,54 @@
...
+using ILGPU.IR.Construction;
+using ILGPU.IR.Values;
+using ILGPU.Util;
+
+namespace ILGPU.IR.Transformations
+{
+    /// <summary>
+    /// Applies aggressive constant propagation to all methods.
+    /// </summary>
+    public sealed class ConstantPropagation : UnorderedTransformation
```
MoFtZ, line 21: QUESTION: Does not appear to be used. For a different patch?

MoFtZ (commented):
ISSUE: All `CudaFixedBuffers_02` unit tests are failing.
Many `CudaArrays_02.ArraysSimple` unit tests are failing.

m4rs-mt:
> ISSUE: All CudaFixedBuffers_02 unit tests are failing.

@MoFtZ That's one of the reasons why this is a draft 😄

Src/ILGPU/IR/Analyses/LoopInfo.cs
m4rs-mt: Hmm I tried this myself and `default` seems to be converted to an instance of `int?` with value `0` instead of having no value at all.

MoFtZ: Might be because the first part of the conditional is returning `int`. Try `default(int?)`.

PR #261, opened 2020-09-10
Fixed sequential completion of OpenCL commands.

### LINKED ISSUE #258: Synchronization issues with OpenCL
When a memory copy is performed in Cuda, followed by a kernel launch, the two operations are performed sequentially.

In OpenCL, even though we are using an in-order queue, the kernel launch sometimes occurs before the memory copy has finished.

I noticed this behaviour in the past, and [this article](https://community.intel.com/t5/OpenCL/clEnqueueWriteBuffer-does-not-finish-before-Kernel/td-p/1077032) seems to confirm it:

> Intel Employee:
> You can get an event from each of the non-blocking clEnqueueWriteBuffer call and then wait for those events to complete prior to launching the kernel.
> Otherwise, you are asking for trouble, since non-blocking calls are not guaranteed to finish when you launch your kernel: you just lucked out with NVidia runtime.
> 

We should probably modify `CLStream` to track the last event, and use this when queuing subsequent commands.

MoFtZ:
Fixes #258.

### DIFF
--- a/Src/ILGPU/Runtime/OpenCL/CLAPI.cs
+++ b/Src/ILGPU/Runtime/OpenCL/CLAPI.cs
@@ -764,8 +764,10 @@ namespace ILGPU.Runtime.OpenCL
             int workDimensions,
             IntPtr* workOffsets,
             IntPtr* globalWorkSizes,
-            IntPtr* localWorkSizes) =>
-            clEnqueueNDRangeKernel(
+            IntPtr* localWorkSizes)
+        {
+            CLException.ThrowIfFailed(EnqueueBarrier(queue));
+            return clEnqueueNDRangeKernel(
                 queue,
                 kernel,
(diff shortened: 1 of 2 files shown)


MoFtZ:
OK, I have made the newly added `CLAPI` functions `internal` rather than `public`. Unlikely that external users will need to call these methods at this point in time - we do not expose the OpenCL concept of events on any of the `CLAPI` functions.

And rather than an `out` parameter, I have used `IntPtr*` for the returned event in `CLAPI.EnqueueBarrierWithWaitList`. This places the responsibility for allocation on the caller - previously, it could have been easy to discard the `out` parameter without properly releasing the event.

MoFtZ:
Rebased onto new DLL-interop generator.

m4rs-mt (approved):

Issue #256, opened 2020-09-04
Are structs with fixed size arrays supported?


binaryalgorithm:
unsafe struct FSMUnit
        {
            public int tx;
            public int ty;
            public fixed int test[256];
            public fixed int test2[256];
        }
        unsafe static void MathKernel(Index2 index, ArrayView2D<FSMUnit> grid)
        {
            int tx = index.X;
            int ty = index.Y;

            grid[tx, ty].tx = tx;            
            grid[tx, ty].ty = ty;

            for (int n = 0; n < 256; n++)
            {
                grid[tx, ty].test[n] = n;
                grid[tx, ty].test2[n] = (256 - n);
            }
        }

(I am using 9.0 latest stable)

This runs except that any field after test[256] won't map correctly (even if it's just another int and not an array). Every struct fills in the tx, ty, and all 256 test values, but test2 doesn't fill in this case, except element 0 in the array gets the value "1", all other values remain at "0" (the initial value).

I have tried with annotation [StructLayout(LayoutKind.Sequential, Pack = 1)] , but this didn't change the result. It may be possible that a SubView with the correct offset can be used, but is there a reason why the struct offset mapping doesn't work after the first fixed sized array?

Also tried with:
           for (int n = 0; n < 256; n++)
            {
                grid[tx, ty].test[n] = tx;
                grid[tx, ty].test2[n] = ty;
            }

The result for say unit 64, 65 is that the test array contains all "64" values as expected, and the first element of test2 array contains "65", while the others are not set, as if it can only access the test2 array base address, but not an array index location.

It is dependent on the order of items in the struct, for example:
        unsafe struct FSMUnit
        {
            public fixed int test[256];
            public fixed int test2[256];
            public int tx;
            public int ty;
        }

^ this has the same behavior on the arrays, but tx / ty fields no longer get set at all.

binaryalgorithm:
I setup the kernel compiler and exported the source (OpenCL 2.0) :
https://pastebin.com/LYte700x

binaryalgorithm:
I see that it isn't using fixed array within the generated struct, and it doesn't match the expected memory layout; instead it tries to make a field for each byte generally. This might be an OpenCL specific problem with how it compiles it?

Yey007:
I tested this first on Cuda. I can confirm that it throws a NotSupportedException when running [this code](https://gist.github.com/Yey007/b3c26fc755ad0d610c72119f3add9cb9). The exception reads:

[...]

PR #262, opened 2020-09-13
Fixed incorrect field access on structures with multiple fixed buffers.

MoFtZ:
Fixes https://github.com/m4rs-mt/ILGPU/issues/256.

When a structure contains a fixed buffer, ILGPU will add padding bytes to correctly align the data structure. However, when there are multiple fields in the structure (such as another fixed buffer),  these padding fields alter the field index of the MSIL code.

This PR introduces a new node type (`PaddingType`) to identify which fields of a structure are for padding. A new type was used because it was not always possible to pass information about which fields were padding e.g. when merging structures.


m4rs-mt (approved):
Nice patch 🚀

Issue #213, opened 2020-07-27
Crash in kernel compilation when using multiple fixed buffers in a struct (0.9-beta1)


dje-dev:
Thank you for the new fixed feature, this will be very useful. I believe the code below demonstrates a problem in code generation (CUDA only) with this new feature. 

Perhaps related, if only the first fixed array is defined but made larger (2048 elements) a different error is seen: "device kernel image image is invalid."
 ```c#
public unsafe struct TestStruct
{
  public fixed int Bytes1[2];
  public fixed int Bytes2[1];   // Compilation is successful if this line is commented out
}
  
  static unsafe void MyKernel(Index1 index, TestStruct ss)
  {
    // PROBLEM: yields IndexOutOfRangeException in PTXCodeGenerator.GenerateCode
    if (LocalStruct.Bytes1[0] == 7) return; 
    
    // PROBLEM: yields "CUDA_ERROR_INVALID_PTX"
    if (ss.Bytes1[0] == 7) return; 
}


  static readonly TestStruct LocalStruct;

  static void Main()
  {
    using (var context = new Context())
      foreach (var acceleratorId in Accelerator.Accelerators)
        using (var accelerator = Accelerator.Create(context, acceleratorId))
          accelerator.LoadAutoGroupedStreamKernel<Index1, TestStruct>(MyKernel);          
  }
```

m4rs-mt:
@dje-dev Thank you for submitting this issue.

> Perhaps related, if only the first fixed array is defined but made larger (2048 elements) a different error is seen: "device kernel image image is invalid."

This error message indicates that a particular structure is too to fit into constant memory of your GPU and is issued by the GPU driver. The upcoming capability API (#103) will provide high-level error messages in the future.

Regarding the compilation errors: I can confirm that loading an unsafe structure from a static field does not seem to work properly. This piece of code throws an internal compiler error, indicating that some functions to handle these cases are currently missing. However, I cannot reproduce the problems you mentioned regarding "default" buffer accesses. Both cases you posted seem to work like a charm.

PR #263, opened 2020-09-14
Fixed loading unsafe structure from a static field.

MoFtZ:
Fixes https://github.com/m4rs-mt/ILGPU/issues/213.

When loading an unsafe structure from a static field, and the structure contains a fixed buffer, we need to load all the elements of the fixed buffer. This is accomplished by pinning the first element of the fixed buffer, then using unsafe code to copy the remaining elements one byte at a time.

Also had to alter equality comparison of `TypeNode` so that `PrimitiveType` and `PaddingType` are considered equal. Otherwise, the additional padding bytes would fail to consider a `PrimitiveValue` to be compatible with `PaddingType`.

### DIFF
--- a/Src/ILGPU/IR/Construction/Structures.cs
+++ b/Src/ILGPU/IR/Construction/Structures.cs
@@ -14,7 +14,9 @@ using ILGPU.IR.Values;
 using ILGPU.Resources;
 using ILGPU.Util;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
+using System.Runtime.InteropServices;
 using ValueList = ILGPU.Util.InlineList<ILGPU.IR.Values.ValueReference>;
 
 namespace ILGPU.IR.Construction
@@ -61,7 +63,8 @@ namespace ILGPU.IR.Construction
             var instanceBuilder = CreateStructure(location, structureType);
             for (int i = 0, e = typeInfo.NumFields; i < e; ++i)
             {
-                var rawFieldValue = typeInfo.Fields[i].GetValue(instance);
+                var field = typeInfo.Fields[i];
+                var rawFieldValue = field.GetValue(instance);
                 Value fieldValue = CreateObjectValue(
                     location,
                     rawFieldValue);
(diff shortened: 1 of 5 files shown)


Src/ILGPU/IR/Construction/Structures.cs
```diff
@@ -80,6 +84,34 @@ partial class IRBuilder
                 else
                 {
                     instanceBuilder.Add(fieldValue);
+
+                    // If the field value we just added was the start of a fixed buffer,
+                    // copy the remaining elements, byte by byte.
+                    if (field.GetCustomAttribute<FixedBufferAttribute>() != null)
```
MoFtZ, line 90: QUESTION: Is this the best way to detected a fixed buffer?

Src/ILGPU/IR/Construction/Structures.cs
```diff
@@ -80,6 +84,34 @@ partial class IRBuilder
...
+                        var initialBytes = fieldValue.Type.Size;
+                        var numBytes = fieldTypeInfo.Size;
+                        var handle = GCHandle.Alloc(rawFieldValue, GCHandleType.Pinned);
+                        try
+                        {
+                            unsafe
+                            {
+                                byte* ptr = (byte*)handle.AddrOfPinnedObject();
+                                for (int j = initialBytes; j < numBytes; j++)
+                                {
```
MoFtZ, line 102: QUESTION: Do I need to worry about endian-ness?

Src/ILGPU/IR/Construction/Structures.cs
m4rs-mt: In general, there is no guarantee that all .Net compatible compilers attach this attribute. What we could to is to check whether the native `size` of a type enforces a custom "padding".

Src/ILGPU/IR/Construction/Structures.cs
m4rs-mt: I guess this should work 🤓

Src/ILGPU/IR/Types/TypeNode.cs
```diff
@@ -301,5 +301,27 @@ public T As<T>(ILocation location)
...
+
+        #region Operators
+
+        /// <summary>
+        /// Returns true if the left and right type nodes are the same.
+        /// </summary>
+        /// <param name="left">The left type node.</param>
+        /// <param name="right">The right type node.</param>
+        /// <returns>True, if the left and right type nodes are the same.</returns>
+        public static bool operator ==(TypeNode left, TypeNode right) =>
```
m4rs-mt, line 313: I am not sure whether we should really implement these operators as the compiler is based on the invariant that if the types are the "same" (using `==`) they actually represent the same type. Is this really required?

Src/ILGPU/IR/Construction/Structures.cs
MoFtZ: Changed to compare field size.

Src/ILGPU/IR/Types/TypeNode.cs
MoFtZ: I've changed the PR to handle this differently. It only affect the one method in `StructureValues.cs`, so I've just changed the assertion.

MoFtZ:
Replaced the check for `FixedBufferAttribute` with a check on field size, for padding.

Reverted changes to `TypeNode` equality and replaced with non-intrusive change to assertion.

Added new assertion for `PaddingType.BasicValueType == BasicValueType.Int8`.

m4rs-mt (approved):

PR #268, opened 2020-09-21
Added new arithmetic simplifications to fold common logical patterns.

m4rs-mt:
The current constant-folding functionality does not support logical `and` and `or` operations. This PR adds support for applying De Morgan's rules to binary `and` and `or` operations in the presence of unary `neg` values.


MoFtZ (changes requested):
MINOR: Spelling mistake in commit message 'logial' vs 'logical'.

MoFtZ (approved):

PR #275, opened 2020-09-21
Fixed invalid alignment of local and shared-memory allocations in PTXBackend

m4rs-mt:
We have updated the global alignment rules in #260. This PR fixes invalidly aligned local and shared-memory allocations that are not compatible with the updated alignment information.

### DIFF
--- a/Src/ILGPU/Backends/PTX/PTXCodeGenerator.cs
+++ b/Src/ILGPU/Backends/PTX/PTXCodeGenerator.cs
@@ -585,7 +585,8 @@ namespace ILGPU.Backends.PTX
                 Builder.Append(addressSpacePrefix);
 
                 Builder.Append(".align ");
-                Builder.Append(allocaInfo.ElementAlignment);
+                Builder.Append(
+                    PointerAlignments.GetInitialAlignment(allocaInfo.Alloca));
                 Builder.Append(" .b8 ");
 
                 var name = namePrefix + offset++;


Src/ILGPU/Backends/PTX/PTXCodeGenerator.cs
```diff
@@ -585,7 +585,8 @@ protected void GenerateCodeInternal(int registerOffset)
                 Builder.Append(addressSpacePrefix);
 
                 Builder.Append(".align ");
-                Builder.Append(allocaInfo.ElementAlignment);
+                Builder.Append(
```
MoFtZ, line 588: QUESTION: Isn't this just the same thing?

`PointerAlignments.GetInitialAlignment` will be supplied an `Alloca`, which then calls `Alloca.AllocaType.Alignment`. This is currently identical to `ElementAlignment`.

m4rs-mt: `PointerAlignments.GetInitialAlignment` checks if the current type size can be used as alignment information. This ensures that types like `Float2` can be automatically aligned with 8 bytes instead of the basic alignment information of 4 bytes.

MoFtZ: Ah, I see. I was looking at old code before c31fcb5f8f69b81890324a079eb349c8a5f5a387.

MoFtZ (approved):

PR #279, opened 2020-09-26
Implement Capability API.

### LINKED ISSUE #103: New Capability API
Different kernels use different hardware, vendor, and platform-specific features. Depending on the features actually available, we have to perform different operations. For example, an atomic operation might require a specific software implementation to "emulate" the required functionality. Or it may not be possible to implement a specific kernel function on a target platform configuration. In this case ILGPU should be able to provide detailed high-level error messages, why a specific kernel is not compatible and which function/source line caused this problem.

MoFtZ:
Fixes #103.

Adds a new `CapabilityContext` base class for tracking available capabilities. Currently, there are no common features between Cuda and OpenCL. If there were, it would be added to this base class.

There are also accelerator-specific `CudaCapabilityContext` and `CLCapabilityContext` classes. These are passed to the respective backends to identity supported features.

There is also a new `CapabilityNotSupportedException`. We may be able to remove this and just use the standard `NotSupportedException`.


MoFtZ:
Added some capabilities for Float16 on Cuda - Min/Max/TanH are not available on SM_60.

MoFtZ:
DISCUSSION: Also noticed a potential issue - I was code-generating functions to create `Exceptions`, but this means that we cannot share the error strings.

Maybe this should be done manually. What do you think?

MoFtZ:
ISSUE: How will this prevent ILGPU.Algorithms from overriding the Cuda Float32 TanH intrinsic?

Src/ILGPU/Context.cs
```diff
@@ -170,9 +171,10 @@ public Context(ContextFlags flags, OptimizationLevel optimizationLevel)
                 : new ILFrontend(frontendDebugInformationManager, 1);
 
             // Create default IL backend
+            var cpuCapabilities = new CPUCapabilityContext();
```
m4rs-mt, line 174: Question: Do we really need to pass a `CPUCapabilityContext` instance to each `ILBackend` instance?

Src/ILGPU/IR/Analyses/Loops.cs
```diff
@@ -740,7 +740,7 @@ private Loops(CFG<TOrder, TDirection> cfg)
             {
                 foreach (var predecessor in member.GetPredecessors<TDirection>())
                 {
-                    if (!nodeMapping[predecessor].IsInSCC && 
+                    if (!nodeMapping[predecessor].IsInSCC &&
```
m4rs-mt, line 743: Suggestion: Maybe we can skip this change for now since it might make more sense to include this tiny update in the loop unrolling PR. What do you think?

Src/ILGPU/Backends/OpenCL/CLCodeGenerator.Values.cs
```diff
@@ -256,6 +257,14 @@ public void GenerateCode(IfPredicate predicate)
         /// <summary cref="IBackendCodeGenerator.GenerateCode(GenericAtomic)"/>
         public void GenerateCode(GenericAtomic atomic)
         {
+            if ((atomic.ArithmeticBasicValueType == ArithmeticBasicValueType.Int64 ||
```
m4rs-mt, line 260: Suggestion: Maybe we can try to refactor this a bit 🤓

Src/ILGPU/Backends/OpenCL/CLTypeGenerator.cs
```diff
@@ -117,24 +118,30 @@ public sealed class CLTypeGenerator : DisposeBase, ICLTypeGenerator
         /// </summary>
         /// <param name="basicValueType">The basic-value type to resolve.</param>
         /// <returns>The resolved OpenCL type name.</returns>
-        public static string GetBasicValueType(BasicValueType basicValueType) =>
-            BasicTypeMapping[(int)basicValueType];
+        public string GetBasicValueType(BasicValueType basicValueType) =>
+            basicValueType == BasicValueType.Float16 && !Capabilities.Float16
```
m4rs-mt, line 122: Suggestion: Maybe we can try to refactor these pieces as well 🤓

m4rs-mt (approved):

Src/ILGPU/Context.cs
MoFtZ: Changed to have the `CPUCapabilityContext` constructed internally in the `ILBackend`.

Src/ILGPU/IR/Analyses/Loops.cs
MoFtZ: Rebased successfully now that the loop unrolling PR in merged.

Src/ILGPU/Backends/OpenCL/CLCodeGenerator.Values.cs
MoFtZ: Refactored into common method.

Src/ILGPU/Backends/OpenCL/CLTypeGenerator.cs
MoFtZ: No change. Unfortunately, this is using `BasicValueType`, while the function below is `ArithmeticBasicValueType`.

MoFtZ:
> 
> DISCUSSION: Also noticed a potential issue - I was code-generating functions to create `Exceptions`, but this means that we cannot share the error strings.
> 
> Maybe this should be done manually. What do you think?

Refactored to use generic message with the capability defining a `FeatureName`.

> 
> 
> ISSUE: How will this prevent ILGPU.Algorithms from overriding the Cuda Float32 TanH intrinsic?

It does not help. ILGPU.Algorithms will have to known which intrinsic to register.

m4rs-mt (approved):

PR #280, opened 2020-09-27
Added support for System.ValueTuple.

### LINKED ISSUE #266: Does kernel function support tuple type？

MoFtZ:
Fixes #266.

According to [this post](https://stackoverflow.com/a/54014473) from a developer on the C# compiler, `System.ValueTuple` can be used with the `unmanaged` constraint, even though it also uses `LayoutKind.Auto`.

However, when running the unit tests, there appears to be an issue with some tests.
It looks like copying the ValueTuples to the input buffer is correct, so I'm assuming it is related to the kernel.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -255,6 +255,7 @@ Src/ILGPU/HalfConversion.cs
 Src/ILGPU/IR/Construction/ArithmeticOperations.cs
 Src/ILGPU/IR/Construction/CompareOperations.cs
 Src/ILGPU/IR/Intrinsics/IntrinsicMatchers.cs
+Src/ILGPU/IR/Types/ValueTuples.cs
 Src/ILGPU/IndexTypes.cs
 Src/ILGPU/IntrinsicMath.CPUOnly.cs
 Src/ILGPU/ReductionOperations.cs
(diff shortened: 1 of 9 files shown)


MoFtZ:
`System.ValueTuple` uses `LayoutKind.Auto`, which can rearrange the fields.
For example, `System.ValueTuple<short, long>` is actually stored with the `long` first, then the `short`.
To fix this, we dynamically calculate the offsets of the fields, and reorder them according to the offset.

There could also be trailing padding bytes.
For example, `System.ValueTuple<short, short, short>` is sized at 8 bytes.
To fix this, we dynamically calculate the size of the value tuple.

MoFtZ:
Simplified calculating the size of `System.ValueTuple` by addnig a new overload to the existing `Interop.SizeOf` method. The new method `Interop.SizeOf(Type)` still will enforce the `unmanaged` constraint.

m4rs-mt:
The actual ordering of `Tuple` fields are unfortunately determined at runtime 🙄 

> There could also be trailing padding bytes.
For example, System.ValueTuple<short, short, short> is sized at 8 bytes.

Oh, that's really interesting; I was not aware of this fact. However, I guess that your patch provides a reasonable way to deal with these ugly implementation details of .Net value tuples.

MoFtZ:
Minor update to fix incomplete comment in `TypeInformationManager.cs`.

m4rs-mt (approved):

PR #285, opened 2020-09-28
Fixed misaligned field access for empty structures.

MoFtZ:
Empty structures in .NET occupy a hidden field.

```CSharp
struct EmptyStruct { }  // Size = 1 byte

struct ExampleStruct    // Size = 6 bytes
{
    public EmptyStruct Field1;   // Offset 0
    public ushort Field2;        // Offset 2
    public byte Field3;          // Offset 4
}
```

However, when building up the internal type information ILGPU would incorrectly treat `ExampleStruct` as having only 2 fields. Therefore, the access to `Field2` and `Field3` would use offset 0 and offset 2, rather than offset 2 and offset 4.


m4rs-mt (approved):

m4rs-mt:
Awesome patch 🚀. I guess we should merge this before releasing v0.9.1.

PR #291, opened 2020-10-05
Added LoopInvariantCodeMotion transformation

m4rs-mt:
Added new LoopInvariantCodeMotion (LICM) transformation to move loop-invariant values out of loops. This transformation reduces the computational effort within loops by trying to reuse values, which can be precalculated, instead of (potentially) expensive recalculations in every iteration.


m4rs-mt:
Put this PR into `Draft` mode due to #292.

m4rs-mt:
Depends on #293 and must be further investigated to ensure that all compatible loops can still be fully unrolled after applying the LICM transformation.

MoFtZ (approved):

PR #295, opened 2020-10-09
Fixed test verification of expected length.

MoFtZ:
When verifying the length of the data buffer, it is possible to supply a custom data length. However, we were incorrectly comparing the original length of the data buffer with the expected length.


MoFtZ:
Changed back into draft - appears to break some existing unit tests in ILGPU.Algorithms.

m4rs-mt (approved):

PR #298, opened 2020-10-21
Fixed arithmetic simplifications.

### LINKED ISSUE #296: Regression in arithmetic simplications affecting ILGPU v0.9.2-beta1
The changes in https://github.com/m4rs-mt/ILGPU/pull/283 has caused a regression in ILGPU.Algorithms.

For example, ILGPU.Algorithms.Tests.Cuda.CudaGroupExtensionTests_Debug.ExclusiveScanAdd fails with "an illegal memory access was encountered".

Looking at the PTX output, there's a one line difference.

Requires further investigation.

m4rs-mt:
This PR fixes #296.

### DIFF
--- a/Src/ILGPU.Tests/BinaryIntOperations.tt
+++ b/Src/ILGPU.Tests/BinaryIntOperations.tt
@@ -3,6 +3,7 @@
 <#@ assembly name="System.Core" #>
 <#@ import namespace="System.IO" #>
 <#@ output extension=".cs" #>
+using ILGPU.Runtime;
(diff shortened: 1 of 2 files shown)


MoFtZ (approved):
This fixes the current bug, but does not deal with folding related operations.
e.g. addition/subtraction, multiplication/division, shift left/right

m4rs-mt:
@MoFtZ Yes, you are right. I guess this should be fine for the moment. We can think about more advanced simplifications in the future 🤓

PR #299, opened 2020-10-21
Relaxed data-length assertion in TestBase.

m4rs-mt:
#295 introduced an assertion issue in the `Test.Cuda` project affecting all `WarpOperation` tests. This PR relaxes the assertion added in #295 to circumvent this issue.


MoFtZ (commented):
QUESTION: Should the Warp unit tests be updated to use a smaller expected buffer instead?

m4rs-mt:
@MoFtZ Yes, I guess you are right. This might be the better alternative.

PR #302, opened 2020-10-30
Added support for RTX 30xx cards.

MoFtZ:
Added support for RTX 30xx cards, as spotted in https://github.com/m4rs-mt/ILGPU/pull/25#issuecomment-719117666.

### DIFF
--- a/Src/ILGPU/Backends/PTX/PTXArchitecture.cs
+++ b/Src/ILGPU/Backends/PTX/PTXArchitecture.cs
@@ -94,6 +94,11 @@ namespace ILGPU.Backends
         /// </summary>
         public static readonly PTXArchitecture SM_80 = new PTXArchitecture(8, 0);
 
+        /// <summary>
+        /// The 8.6 architecture.
+        /// </summary>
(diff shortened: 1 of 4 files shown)


m4rs-mt (approved):
Awesome patch 🚀

MoFtZ:
I'm thinking I might raise another PR, that would add a ContextFlag to ignore the driver check. That way, it could be used on newer cards until ILGPU is updated.

What do you think?

m4rs-mt:
@MoFtZ Makes sense to me. In these cases you could try ILGPU on newer hardware without waiting for us to update the code and do further compatibility checks.

MoFtZ:
Actually, made a different/easier change.

If the graphics card is on an PTX architecture that ILGPU does not know about, the user must be using a newer graphics cards. This newer card will require new drivers, so it should work just fine.

The only issue would be if the newer drivers dropped support for the PTX instruction sets that we support, but the risk is very small.

m4rs-mt:
@MoFtZ I totally agree. The risk should be very small. Usually, an instruction can become obsolete when upgrading the PTX version, which should not happen when upgrading the drivers.

PR #304, opened 2020-10-30
Enhanced allocation alignment

m4rs-mt:
The current allocation alignment rules (for shared and local memory) are based on the actual type used to perform the allocation:
```c#
var sharedMemory = SharedMemory.Allocate<int>(42);
```

Since the available memory size (e.g. in the case of shared memory) is usually limited, many programmers often tend to leverage reinterpret casts to use the available space in combination with different types:

```c#
var sharedMemory = SharedMemory.Allocate<int>(42);
...

var float2Shared = sharedMemory.Cast<Float2>();
```

Unfortunately, ILGPU is currently not able to use efficiently vectorized-IO operations in these cases. This is caused by the alignment information that has been determined using the initially allocated type (`int` in this case). Many programmers currently seem to work around this limitation by adapting their programs in a way that they use the "largest" type when allocating a chunk of memory and then use additional reinterpret casts, as shown in the following example:
```c#
var allocatedMemory = SharedMemory.Allocate<Float2>(42);
...
var sharedMemory = allocatedMemory.Cast<Float2>();
...

var float2Shared = allocatedMemory.Cast<Float2>();
```

This PR automatically determines the "best" alignment (given by the maximal alignment of all potentially involved types) for each allocation. Hence, it is no longer required to work around this limitation.

Note that this feature in enabled in the scope of the `O2` pipeline.

### DIFF
--- a/Src/ILGPU/Backends/PTX/PTXCodeGenerator.cs
+++ b/Src/ILGPU/Backends/PTX/PTXCodeGenerator.cs
@@ -451,6 +451,15 @@ namespace ILGPU.Backends.PTX
         #region General Code Generation
 
         /// <summary>
+        /// Returns the alignment in bytes for the given alloca.
+        /// </summary>
+        /// <param name="alloca">The alloca to get the alignment information for.</param>
+        /// <returns>The determined and used alignment in bytes.</returns>
+        protected int GetAllocaAlignment(Alloca alloca) =>
+            PointerAlignments?.GetAllocaAlignment(alloca) ??
+            AllocaAlignments.GetInitialAlignment(alloca);
+
+        /// <summary>
         /// Declares a new label.
         /// </summary>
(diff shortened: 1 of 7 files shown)


Src/ILGPU/Backends/PTX/PTXCodeGenerator.cs
```diff
@@ -396,6 +396,11 @@ internal PTXCodeGenerator(in GeneratorArgs args, Method method, Allocas allocas)
         /// </summary>
         public bool FastMath { get; }
 
+        /// <summary>
+        /// Returns true if the current backend uses enhanced features.
+        /// </summary>
+        public bool EnhancedFeatures { get; }
+
```
MoFtZ, line 403: MINOR: Unused variable?

MoFtZ (approved):

PR #306, opened 2020-10-31
Fixed regression in testing size of expected buffer.

MoFtZ:
Alternate fix for #299.


m4rs-mt:
@MoFtZ Maybe we should change the `[Theory]` semantics to become a `[Fact]` instead. This avoids using the `warpMultiplier`, which might lead to `ArrayOutOfBounds` exceptions using this fix otherwise.

m4rs-mt (approved):

Issue #310, opened 2020-11-03
[0.9.2-beta1] OpenCL: Float16 (Half) type is not supported on this device


deng0:
When I try to build a Kernel for OpenCL with 0.9.2-beta1 I get the following exception:

```
[ILGPU.CapabilityNotSupportedException: Float16 (Half) type is not supported on this device.]
   bei ILGPU.Backends.OpenCL.CLTypeGenerator.GetBasicValueType(BasicValueType basicValueType)
   bei ILGPU.Backends.OpenCL.CLTypeGenerator..ctor(IRTypeContext typeContext, CLCapabilityContext capabilities)
   bei ILGPU.Backends.OpenCL.CLBackend.CreateKernelBuilder(EntryPoint entryPoint, BackendContext& backendContext, KernelSpecialization& specialization, GeneratorArgs& data)
   bei ILGPU.Backends.CodeGeneratorBackend`4.Compile(EntryPoint entryPoint, BackendContext& backendContext, KernelSpecialization& specialization)
   bei ILGPU.Backends.Backend.Compile[TBackendHook](Method kernelMethod, EntryPointDescription& entry, KernelSpecialization& specialization, TBackendHook backendHook)
```

The same worked with ILGPU 0.9.1.
(I'm not sure if that matters, but as far as i can tell, my source does not even use Float16.)

MoFtZ:
It looks like your device does not support Float16.

During the constructor of CLTypeGenerator, it is trying to initialize the basic types. Will need to add a check to conditionally register Float16.

m4rs-mt:
@deng0 Thank you for your issue report. As @MoFtZ said, we have to conditionally register the `fp16` types to circumvent this exception from being thrown, @MoFtZ right?

m4rs-mt:
@MoFtZ I wonder why this problem did not show up on my old AMD card that I used to test the master branch before publishing the beta, which does not support `fp16` types.

MoFtZ:
@m4rs-mt Not sure why your AMD card did not trigger this issue. Could it be that it a recent driver update added support for Float16?

m4rs-mt:
@MoFtZ It turns out that the AMD driver claims to have FP16 support in the beginning. However, executing a kernel that uses FP16 operations fails 😢

PR #312, opened 2020-11-04
Skip registration of Float16 on unsupported devices.

MoFtZ:
Fixes https://github.com/m4rs-mt/ILGPU/issues/310.


MoFtZ:
Added additional check for Float16 to throw capability exception rather than generic exception.

m4rs-mt (approved):
🚀

PR #313, opened 2020-11-04
Directly throw capability exceptions to the caller.

MoFtZ:


MoFtZ:
Throwing directly does not contain the line/location information. Abandoning PR - requires more thought.

Issue #301, opened 2020-10-30
Act of a cpu and a accelerator differs (may be bug)


sohaspaces:
I think this library really great, but this code using the library prints result which doesn't make sense.
```c#
using System;
using System.Linq;
using System.Runtime.InteropServices;
using ILGPU;
using ILGPU.Runtime;

namespace ILGPUTest
{
    class Program
    {
        static void Main(string[] args)
        {
            var accelerator = Accelerator.Create(new Context(), Accelerator.Accelerators.FirstOrDefault((x) => x.AcceleratorType == AcceleratorType.Cuda));
            
            var gpu = accelerator.LoadAutoGroupedStreamKernel<Index1, ArrayView<float>, MyStruct>(Gpu);
            var view = accelerator.Allocate<float>(1);

            var v = new MyStruct() { x = 255 };


            gpu(new Index1(1), view, v);
            view.CopyTo(out var gpuResult, new LongIndex1(0));
            Console.WriteLine(gpuResult);

            Cpu(out var cpuResult, v);
            Console.WriteLine(cpuResult);
        }
        static void Cpu(out float output, MyStruct input)
        {
            output = input.x;
        }
        static void Gpu(Index1 index, ArrayView<float> output, MyStruct input)
        {
            output[0] = input.x;
        }

    }

    public struct MyStruct
    {
        public byte x;
    }
}
```

Obviously, the method gpu (Compiled for CUDA by using Acclerator.LoadAutoGroupedStreamKernel) and Cpu is supposed to act the same as each other except for memory to write the result on.
But the result of the first one (=-1) and the second one(=255) differs.

The value of the field x of the input is 255, so the result must be 255 though it was cast as float (System.Single).

But it can be -1 if the byte value is cast as sbyte (because of 0xFF is 255 in byte, and -1 in sbyte).
And i guess it is clue to solve this bug

sohaspaces:
I found the fact that the problem doesn't occur when the type of accelerator is AcceleratorType.CPU.
I didn't test the case of AcceleratorType.OpenCL because My graphic card is NVIDIA that doesn't support OpenCL.
But I think this can help to fix the problem by reducing the range in which the problem occurs.

m4rs-mt:
@sohaspaces Thank you for submitting this issue and welcome to the ILGPU community. A small hint: Instead of writing
```c#
[...]

PR #314, opened 2020-11-05
Fixed loading unsigned fields from structures. 

MoFtZ:
Fixes https://github.com/m4rs-mt/ILGPU/issues/301.

<del>When calling `ToTargetUnsignedFlags()`, we use the original managed type of the variable. When we ask the original managed type if it is unsigned, structures return false (as expected). However, the ILGPU IR simplifies single element structures into a primitive type. In this case, if the primitive type is unsigned, `ToTargetUnsignedFlags()` incorrectly returns false (should be true).</del>

<del>The fix is to ask the managed type of the returned IR Type Node if it is unsigned.</del>

### DIFF
--- a/Src/ILGPU.Tests/StructureValues.cs
+++ b/Src/ILGPU.Tests/StructureValues.cs
@@ -323,6 +323,44 @@ namespace ILGPU.Tests
                 expected);
             Verify(buffer, new[] { expected });
         }
+
+        internal struct UnsignedFieldStruct
+        {
+            public byte x;
+            public ushort y;
+            public uint z;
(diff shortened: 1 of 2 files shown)


sohaspaces:
Thank you for solving my issue!

MoFtZ:
@sohaspaces I would hold off on using this PR - it broke other functionality, which is why this PR is in draft.

MoFtZ:
When loading a single field of structure, apply the .NET rules for loading onto evaluation stack.

m4rs-mt (approved):

PR #316, opened 2020-11-09
Enhanced support for local memory allocations.

### LINKED ISSUE #287: Allocated local memory on the GPU is uninitialized
Within a kernel, if we allocate memory e.g. `var temp = new int[9];`, the contents of that array will be uninitialized. However, if we do that allocation on the CPU accelerator, it will be initialized to 0, as per the conventions of .NET.

This can make it difficult to debug issues, since it behaves differently on the GPU. Should we initialized the contents of the array?

m4rs-mt:
This PR adds a new class `LocalMemory` that gives programmers the ability to explicitly allocate (and/or initialize) buffers in the local memory address space. At the same time, it removes all "explicit" array values from the IR and maps allocations to `Alloca` IR values (like in the case of shared memory allocations). This simplifies reasoning about allocations in the scope of ILGPU programs and makes the `LowerArrays` pass obsolete.

Furthermore, it maps "managed" (C#-like) array allocations via `var x = new int[..];` to the `LocalMemory.AllocateZero` method that initializes all arrays with their default values (fixes #287).


Src/ILGPU/Frontend/Intrinsic/LocalMemoryIntrinsic.cs
```diff
@@ -0,0 +1,74 @@
...
+                ? builder.CreateNewView(
+                    location,
+                    builder.CreateAlloca(
+                        location,
+                        allocationType,
+                        MemoryAddressSpace.Local,
+                        context[0]),
+                    context[0])
+                : throw context.Location.GetNotSupportedException(
+                    ErrorMessages.NotSupportedSharedMemoryIntrinsic,
```
MoFtZ, line 70: ISSUE: The error message would be `The shared memory intrinsic 'Allocate' is not supported`.

MoFtZ (changes requested):
QUESTION: There are several references to "shared memory" in the code comments. Should these be changed to "local memory"?

m4rs-mt:
@MoFtZ I have replaced all occurrences of "shared memory" with "local memory". Furthermore, I have added several 2D and 3D convenience methods to the `LocalMemory` class.

Question: Should we add additional 2D and 3D overloads for `AllocateZero`?

MoFtZ (approved):

MoFtZ:
@m4rs-mt I would say it's a nice to have, but not critical. 

We added AllocateZero to `LocalMemory` to provide the same functionality as allocating a new array in .NET e.g. `new int[]`. Also note that we do not offer AllocateZero on the `SharedMemory` class. Maybe we should remove `AllocateZero` from LocalMemory? And only use it internally when creating a `new` array.

m4rs-mt:
@MoFtZ That makes sense to me. I fully agree that we should not reveal the `AllocateZero` method to users.

Costigan:
I don't understand the final decision wrt AllocateZero() in both SharedMemory and LocalMemory.  If my kernel creates some local memory and then immediately initializes it to some non-zero values of my choosing, will the kernel have taken the time to initialize the memory to zero first, then to the values I want?  Can I choose to not have that behaviour?

MoFtZ:
@Costigan The final decision was to only provide an `Allocate` function, which will leave the memory uninitialized.

If the kernel uses `new` to allocate memory, e.g. `var buffer = new int[4];`, ILGPU will internally use `AllocateZero` to initialize the memory with zeros. The reason for this is compatibility with the .NET framework, which will zero the allocated memory.

PR #320, opened 2020-11-16
Added support for .NET 5.0.

MoFtZ:
ILGPU already supports .NET 5.0 through .NET Standard 2.1. This PR adds explicit support for .NET 5.0, and fixes a few minor issues found with the new compiler.

QUESTION: Should the unit tests be updated to .NET 5.0? or multi-targeted with .NET 5.0?

### DIFF
--- a/Src/ILGPU/ILGPU.csproj
+++ b/Src/ILGPU/ILGPU.csproj
@@ -7,6 +7,10 @@
         <Configurations>Debug;Release</Configurations>
     </PropertyGroup>
 
+    <PropertyGroup Condition="'$(Channel)' == 'experimental'">
+        <TargetFrameworks>$(TargetFrameworks);net5.0</TargetFrameworks>
+    </PropertyGroup>
+
     <PropertyGroup>
         <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
         <VersionPrefix>0.10.0-beta1</VersionPrefix>
--- a/Src/ILGPU.Tests.CPU/ILGPU.Tests.CPU.csproj
+++ b/Src/ILGPU.Tests.CPU/ILGPU.Tests.CPU.csproj
@@ -7,6 +7,10 @@
     <IsPackable>false</IsPackable>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(Channel)' == 'experimental'">
+    <TargetFrameworks>$(TargetFrameworks);net5.0</TargetFrameworks>
+  </PropertyGroup>
+
   <PropertyGroup>
     <RunSettingsFilePath>$(MSBuildProjectDirectory)\..\ILGPU.Tests\.test.runsettings</RunSettingsFilePath>
   </PropertyGroup>
--- a/Src/ILGPU.Tests.Cuda/ILGPU.Tests.Cuda.csproj
+++ b/Src/ILGPU.Tests.Cuda/ILGPU.Tests.Cuda.csproj
@@ -7,6 +7,10 @@
     <IsPackable>false</IsPackable>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(Channel)' == 'experimental'">
+    <TargetFrameworks>$(TargetFrameworks);net5.0</TargetFrameworks>
+  </PropertyGroup>
+
   <PropertyGroup>
     <RunSettingsFilePath>$(MSBuildProjectDirectory)\..\ILGPU.Tests\.test.runsettings</RunSettingsFilePath>
   </PropertyGroup>
--- a/Src/ILGPU.Tests.OpenCL/ILGPU.Tests.OpenCL.csproj
+++ b/Src/ILGPU.Tests.OpenCL/ILGPU.Tests.OpenCL.csproj
@@ -7,6 +7,10 @@
     <IsPackable>false</IsPackable>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(Channel)' == 'experimental'">
+    <TargetFrameworks>$(TargetFrameworks);net5.0</TargetFrameworks>
+  </PropertyGroup>
+
   <PropertyGroup>
     <RunSettingsFilePath>$(MSBuildProjectDirectory)\..\ILGPU.Tests\.test.runsettings</RunSettingsFilePath>
   </PropertyGroup>
--- a/Src/ILGPU.Tests/ExchangeBufferOperations.cs
+++ b/Src/ILGPU.Tests/ExchangeBufferOperations.cs
@@ -134,10 +134,6 @@ namespace ILGPU.Tests
         // No need for kernel, assuming copy tests pass.
         // Just going to confirm integrity in this test.
         [Fact]
-        [SuppressMessage(
-            "Microsoft.Performance",
-            "CA1814: PreferJaggedArraysOverMultidimensional",
-            Target = "target")]
         public void GetAsArray()
         {
             using var exchangeBuffer = Accelerator.AllocateExchangeBuffer<long>(
--- a/Src/ILGPU.Tests/ILGPU.Tests.csproj
+++ b/Src/ILGPU.Tests/ILGPU.Tests.csproj
@@ -7,6 +7,10 @@
     <StartupObject />
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(Channel)' == 'experimental'">
+    <TargetFrameworks>$(TargetFrameworks);net5.0</TargetFrameworks>
+  </PropertyGroup>
+
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|AnyCPU'">
     <LangVersion>latest</LangVersion>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
--- a/Src/ILGPU/ArrayView.cs
+++ b/Src/ILGPU/ArrayView.cs
@@ -93,10 +93,6 @@ namespace ILGPU
     /// Represents an abstract array view.
     /// </summary>
     /// <typeparam name="T">The element type.</typeparam>
-    [SuppressMessage(
-        "Microsoft.Design",
-        "CA1040: AvoidEmptyInterfaces",
-        Justification = "Can be used in generic constraints")]
     public interface IArrayView<T> : IArrayView<T, Index1, LongIndex1>
         where T : unmanaged
     { }
--- a/Src/ILGPU/Backends/EntryPoints/ParameterCollection.cs
+++ b/Src/ILGPU/Backends/EntryPoints/ParameterCollection.cs
@@ -15,6 +15,7 @@ using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.Collections.Immutable;
+using System.Runtime.CompilerServices;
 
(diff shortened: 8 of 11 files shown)


sohaspaces:
I think that this should be included in 9.2.0 features if it's possible!

MoFtZ:
@sohaspaces is there a particular urgency for adding this to v0.9.2? ILGPU can already be used with .NET 5.0 because it supports .NET Standard 2.1.

sohaspaces:
Umm, I guess that the phrase 'if it's possible' misleaded you.
There's not urgency to for it.
I didn't have intention to urge this PR as soon as possible. 

I meant that this should be included if the cost is almost zero, because if there's not loss but advantage.

m4rs-mt:
@MoFtZ I think we should continue testing on .NET Core 3.1 - just to be sure 🤓

m4rs-mt:
@MoFtZ Not directly related: I think we should discontinue support for .NET Core 2.0 in v0.9.3/v0.10 as it is considered obsolete anyway.

m4rs-mt (approved):
Awesome work 🚀

MoFtZ:
@m4rs-mt .NET Core 2.1 is an LTS release and is supported until [August 21, 2021](https://dotnet.microsoft.com/platform/support/policy/dotnet-core). Shall we update our .NET Core 2.0 support to 2.1?

m4rs-mt:
@MoFtZ I guess we can either upgrade to .NET Core 2.1 or even to .NET Core 3.1 🤓

MoFtZ:
Updating to .NET Core 3.1 makes sense, since .NET Core 3.0 is no longer supported.

I would be reluctant to drop .NET Core 2.x until we can verify that no current users are affected, particularly since .NET Core 2.1 is still supported.

The other one to consider is .NET Framework v4.7. Do we upgrade to v4.8? Again, how will this impact existing users of ILGPU.

Also, should I make the changes in this PR?

jgiannuzzi:
First of all, thanks a lot for your PR @MoFtZ!

Here are my 2 cents with regards to support of the various .NET runtimes:

### Library targets

- I think we should drop support for .NET Core 2.x entirely, as it requires an extra build target (`netcoreapp2.0`, as it does not support .NET Standard 2.1) and I feel that the intersection between people _not_ using the latest .NET Core and people using the latest ILGPU is probably zero.
- On the .NET framework front, any version we choose requires an extra build target (as it does not support .NET Standard 2.1), so upgrade to v4.8 would only mean cutting off people using v4.7 with no advantages in terms of maintenance. I thus think we should keep targeting `net47`.
- .NET 5.0 apps can use .NET Standard 2.1 libraries. As we are not using any new feature of .NET 5.0 (yet), I think we should not target `net5.0`, and only rely on `netstandard2.1`

In summary, I think we should target `net47` and `netstandard2.1` on Windows, and `netstandard2.1` on Unix. This also means that the resulting NuGet package will be about one third smaller — nice!

### Test targets

Currently we are only targeting .NET Core 3.1. I think we should start running our tests on .NET Framework 4.7 and .NET 5.0 too, so we have more guarantees that ILGPU does work on these runtimes.

### CI pipeline

I suggest to change our CI pipeline to build and test on the 3 SDKs mentioned above, so that the goal of this PR is met (explicit support for building and running ILGPU with .NET 5.0), and we would finally have coverage for .NET Framework 4.7.

I _think_ that this could be achieved by using a matrix and running the tests specifically on the corresponding runtime with `-f`. This means that we would only setup one SDK in each job (instead of 2 in @MoFtZ's current proposal). All of those jobs would run in parallel, so there should not be any slowdown in terms of CI feedback.

Additionally, I think that the NuGet package that gets uploaded as an artifact should be the one built with .NET Core 3.1 (and .NET Framework), and _not_ the one built with .NET 5.0, as it's not an LTS release and we wouldn't want to introduce any bugs that might stem from the compiler.
When an LTS release of .NET 5 appears (.NET 5.1?), we could start publishing that build instead, still targeting `netstandard2.1` for compatibility with .NET Core 3.x.

I'm willing to contribute the changes to our CI pipeline as a PR to @MoFtZ's branch, so that it could get picked up in this PR.

This ended up being a few more cents that I originally intended 😅

@MoFtZ @m4rs-mt what are your thoughts?

m4rs-mt:
@jgiannuzzi Thank you very much for your awesome and helpful "cents" 🤓🚀. I totally agree that we should remove support for `netcoreapp2.0` and should keep the `net47` target since upgrading to `net48` will not yield any benefits.

Regarding the CI pipeline: I also agree that we should ensure building the ILGPU package with the help of the `netcoreap3.1` pipeline instead of the current (potentially "unreliable") `net5.0` target SDK.

MoFtZ:
@jgiannuzzi Thanks for the great feedback. The only addition I would make would be to add the `net5.0` framework to the ILGPU project. When compiling against `netstandard2.1`, it does not appear to pick up the `net5.0`-specific compiler issues.

Maybe the CI pipeline could produce a Nuget package with only net47 and netstandard2.1?

m4rs-mt:
> it does not appear to pick up the net5.0-specific compiler issues.

@MoFtZ Do you refer to an invalid "linking" of libraries and packages or to (potential) issues in the C# compiler?

MoFtZ:
@m4rs-mt For example, commit f4663856758adb5dfc06a2b4a2783e4a09aaab50 generates an error during compilation when using `net5.0`. This compiler error is not generated when compiling the other frameworks.

jgiannuzzi:
@MoFtZ good point — I agree that it could be beneficial to have a `net5.0` target so we already ensure build compatibility. The only downside I see is that someone building on .NET Core 3.x would now get build errors.

It sounds doable in terms of CI, including packaging only `net47` and `netstandard2.1` — if @m4rs-mt agrees with adding the `net5.0` target, I'll try to get that working in the coming days.

MoFtZ:
@jgiannuzzi Maybe it would be easier to use a condition in the csproj to include net5.0 - similar to the one we already have for Linux, but a custom condition like `experimental`. This solves the issue of building locally, and allows the CI to turn on the flag. Additionally, this also simplifies the nuget packaging.

m4rs-mt:
@MoFtZ Yes, you are right; the program analysis extensions and their application to the programs to be compiled may affect compilation compatibility 👍
@jgiannuzzi I also think it might be easier to add an additional condition to include `net5.0` as described by @MoFtZ.

MoFtZ:
@jgiannuzzi @m4rs-mt I've made the change to compile .NET 5.0 based on an `experimental` property.

This also allowed me to remove my changes to the CI pipeline that used both .NET Core 3.1 and .NET 5.0.

Src/ILGPU/ILGPU.csproj
```diff
@@ -7,6 +7,11 @@
         <Configurations>Debug;Release</Configurations>
     </PropertyGroup>
 
+    <PropertyGroup Condition="'$(Experimental)' == 'true'">
+        <TargetFrameworks Condition="'$(OS)' == 'Windows_NT'">net47;netcoreapp2.0;netstandard2.1;net5.0</TargetFrameworks>
```
m4rs-mt, line 11: Question: `netcoreapp2.0`?

m4rs-mt:
@MoFtZ Currently, the CI seems to fail because on an invalid version string. Any ideas why?

Src/ILGPU/ILGPU.csproj
MoFtZ: Rather than trying to remove it using this PR, I've left `netcoreapp2.0` alone.
Can raise a separate PR to that it's a bit more obvious in our commit history.

MoFtZ:
> 
> 
> @MoFtZ Currently, the CI seems to fail because on an invalid version string. Any ideas why?

It looks like the CI pipeline is building commit `03130715cc46b0fa6419e6a59ca8b14d642244d1`.
This commit id has the short form of `031307`, which we use as a version suffix.
The full version then becomes `0.9.2-0313071`.

It just so happens that the short commit id only contains numbers.
This appears to be causing the MSBuild target to think it is an invalid version string.
`error : '0.9.2-0313071' is not a valid version string.`

A workaround could be adding a letter to the suffix. e.g. G for Git. `0.9.2-G0313071`

m4rs-mt:
@MoFtZ Sounds good to me. Let's go with the `G` for git.

Src/ILGPU/ILGPU.csproj
m4rs-mt: Makes sense 🤓

MoFtZ:
Rebased and removed `netcoreapp2.0`.

m4rs-mt (approved):

jgiannuzzi:
Little update: I am working on changing the CI to support building the library in parallel with the various target frameworks we want to support (`net47`, `netstandard2.1`, and _experimental_ `net5.0`). After the `net47` and `netstandard2.1` jobs are successful, a NuGet packaging job would run, followed by tests running against it in parallel for our various test targets (`net47`, `netcoreapp3.1` and `net5.0`).

With this setup, we would ensure that we support building with .NET 5.0, but also and more importantly, that an app using .NET 5.0 would work with the package we build with .NET Core 3.1.

It would also mean that a someone can build the library from source under any of the supported SDKs, and can run the tests on at least one target. If we could find a way to have a conditional that tests which .NET runtime is available, we could even make sure that someone running the tests wouldn't get any errors about missing runtimes (no idea whether this is possible though!).

MoFtZ:
Rebased to fix merge conflicts.

@jgiannuzzi FYI, I've got the CI "working", but it runs the all the tests on an OS in sequence.

jgiannuzzi:
@MoFtZ I have seen that, great stuff! The only issue is that we end up not testing .NET 5.0 against the NuGet package we would publish. I think it's quite important to have our CI support that use case.

m4rs-mt:
@MoFtZ Regarding the "linting" issues in the experimental build: I think it makes sense that I check them all myself, since they involve a lot of highly sensitive parts of the compiler internals. What do you think?

jgiannuzzi:
Hi @MoFtZ, after discussing @m4rs-mt offline, we think it would be best to merge your PR without the CI changes, so I can focus on them in a separate PR. Could you please revert the changes on `.github/workflows/ci.yml`?

To be clear, there is nothing wrong with your changes, and I will use them as a jumping point. It's just for clarity purposes that we would like to split the CI changes into a separate PR.

MoFtZ:
> 
> 
> @MoFtZ Regarding the "linting" issues in the experimental build: I think it makes sense that I check them all myself, since they involve a lot of highly sensitive parts of the compiler internals. What do you think?

@m4rs-mt sounds good to me. We might want to turn down the linting level too - seems to only affect the Release build.





> 
> 
> Hi @MoFtZ, after discussing @m4rs-mt offline, we think it would be best to merge your PR without the CI changes, so I can focus on them in a separate PR. Could you please revert the changes on `.github/workflows/ci.yml`?
> 
@jgiannuzzi Done.

m4rs-mt:
@MoFtZ I think so; this should help us to reduce the number of changes for now. However, we can safely merge this PR which allows us to build the project locally using the `experimental` config to "solve" the linting issue.

m4rs-mt (approved):

PR #323, opened 2020-11-18
New convenient alignment methods for views

m4rs-mt:
This PR adds support for `view.AlignTo<T>(int alignmentInBytes)` methods. These methods can be used to conveniently align the current array view to an explicitly defined `alignmentInBytes`. It also includes two new IR values, namely `PointerAsIntCast` and `AlignViewTo`. Both values have been added to all required static program analyses and program transformations. In addition, all backends have been adjusted to support the new operations.

There are also several utility methods that can be useful in most cases (see `ArrayViews.tt` for more information):
```c#
view.AlignTo128();
view.AlignTo256();
...
```

Sample usage:
```c#
var (prefix, main) = view.AlignTo128();
// prefix is the sub view, which includes all unaligned items at the beginning
// main is the sub view, which includes all remaining aligned items in the source view
```

Note that this PR also depends on #322, since the new `AlignViewTo` IR node is essentially implemented via a structure that can cause issues during lowering.

### DIFF
--- a/Src/ILGPU.Tests/ArrayViews.cs
+++ b/Src/ILGPU.Tests/ArrayViews.cs
@@ -1,4 +1,6 @@
-﻿using System.Linq;
+﻿using System;
+using System.Diagnostics;
+using System.Linq;
 using Xunit;
 using Xunit.Abstractions;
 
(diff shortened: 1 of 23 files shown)


m4rs-mt:
Note that this PR changes the currently implemented "default" structure padding semantics in 02d7d2a4a9fc3f446835365ebc037b0d033cf979. Structures that contain `View` elements cannot be padded properly until all view-depended elements have been lowered. However, this behavior is non-trivial to test separately because these cases usually appear during lowering and IR construction phases. Therefore, this PR contains this fix, as well.

MoFtZ (approved):

PR #324, opened 2020-11-18
Support for PopCount, CLZ and CTZ intrinsics

m4rs-mt:
This PR provides support for the intrinsic functions `PopCount`, `CountLeadingZeros` and `CountTrailingZeros`, which can be efficiently implemented on GPUs. It also contains several software implementations in case the intrinsic functions are not supported.

This PR essentially ports the related [PR](https://github.com/m4rs-mt/ILGPU.Algorithms/pull/63) from the Algorithms library to the main ILGPU library. However, I used a different set of software implementations that use less conditional branches.


MoFtZ (approved):

MoFtZ:
SUGGESTION: Use the function names `LeadingZeroCount` and `TrailingZeroCount`, as these align with the [BitOperations](https://docs.microsoft.com/en-us/dotnet/api/system.numerics.bitoperations?view=net-5.0) class.

PR #328, opened 2020-11-22
Added new IfConditionalConversion to fold nested and-also and or-else block chains

m4rs-mt:
Currently, ILGPU cannot fold nested `&&` and `||` CFG-based block jumps into logical `&` and `|` operations. This PR adds a new transformation called `IfConditionalConversion`, which converts such block jumps into efficient logical operations without branches.
Consider the sample below:
```c#
if (a > 0 && b < c || d > 4) {
   ...
}
```
which can be compiled using `v0.9.2` + `O2` into the following sequence of PTX instructions:
```asm
L_9710:
setp.le.s32	%p1, %r1, 0;
@%p1 bra	L_9712;
bra	L_9711;

L_9711:
setp.lt.s32	%p2, %r2, %r3;
@%p2 bra	L_9713;
bra	L_9712;

L_9712:
setp.le.s32	%p3, %r4, 4;
@%p3 bra	L_9714;
bra	L_9713;

L_9713:
mov.b32	%r5, %ctaid.x;
mov.b32	%r6, %tid.x;
mov.b32	%r7, %ntid.x;
mul.lo.s32	%r8, %r5, %r7;
add.s32	%r9, %r6, %r8;
mul.wide.u32	%rd4, %r9, 4;
add.u64	%rd3, %rd1, %rd4;
st.b32	[%rd3], 42;
bra	L_9714;
```
With the help of this PR this piece of code is then converted into (since all blocks are considered "small enough" and have no side effects):
```c#
if (a > 0 & b < c | d > 4) {
   ...
}
```
which results in the following sequence of PTX instructions:
```asm
setp.gt.s32	%p1, %r1, 0;
setp.lt.s32	%p2, %r2, %r3;
and.pred	%p3, %p1, %p2;
setp.gt.s32	%p4, %r4, 4;
or.pred	%p5, %p3, %p4;
@%p5 bra	L_11021;

L_11020:
...

L_11021:
mov.b32	%r5, %ctaid.x;
mov.b32	%r6, %tid.x;
mov.b32	%r7, %ntid.x;
mul.lo.s32	%r8, %r5, %r7;
add.s32	%r9, %r6, %r8;
mul.wide.u32	%rd4, %r9, 4;
add.u64	%rd3, %rd1, %rd4;
st.global.b32	[%rd3], 42;
bra	L_11020;
```

### DIFF
--- a/Src/ILGPU/IR/BasicBlockMapping.cs
+++ b/Src/ILGPU/IR/BasicBlockMapping.cs
@@ -243,6 +243,26 @@ namespace ILGPU.IR
             return removed;
         }
 
+        /// <summary>
+        /// Asserts that the set has been cleared.
+        /// </summary>
+        [SuppressMessage(
+            "Performance",
+            "CA1822:Mark members as static",
+            Justification = "For debugging purposes only")]
+        [Conditional("DEBUG")]
+        readonly partial void AssertCleared();
+
+        /// <summary>
+        /// Removes all elements from this set.
+        /// </summary>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public void Clear()
+        {
+            Array.Clear(visited, 0, visited.Length);
+            AssertCleared();
+        }
+
         #endregion
     }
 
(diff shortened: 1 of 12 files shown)


Src/ILGPU/IR/BasicBlockMapping.cs
```diff
@@ -243,6 +243,26 @@ public bool Remove(BasicBlock block)
...
+        [Conditional("DEBUG")]
+        readonly partial void AssertCleared();
+
+        /// <summary>
+        /// Removes all elements from this set.
+        /// </summary>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public void Clear()
+        {
+            Array.Clear(visited, 0, visited.Length);
```
MoFtZ, line 262: QUESTION: Should the `visited` array be reset?

MoFtZ (approved):

Src/ILGPU/IR/BasicBlockMapping.cs
MoFtZ: Actually, my mistake. Looks like I mistook the code as clearing `values` rather than `visisted`. All good.

PR #342, opened 2020-12-04
Added support for printf-like output in Kernels

m4rs-mt:
Cuda and OpenCL provide the ability to print/output basic values into the console output stream for debugging. This is especially useful for device-specific concurrency problems that need to be analyzed. This PR provides platform independent support for `ILGPU.Interop.Write` and `ILGPU.Interop.WriteLine` that accept (very) basic .Net-like format specifiers of the form `Test output {0} and test output {1}` (can be formatted to "`Test output 1.0 and test output -45`" using `Interop.Write("Test output {0} and test output {1}", 1.0, -45)`).

Note that more advanced formatting options are not currently available.
Also note that the argument types are automatically derived by the compiler.
Note that only elementary value types are supported.


samy2:
I just tested that in v0.10.0-beta1, and that works great! This feature will greatly facilitate my Cuda debugging sessions! Thank you so much!

Issue #356, opened 2020-12-11
Cuda vs CPU different results


CsabaStupak:
I'm playing with the library, trying to write my own reducer. My input array is `[4,1,0]` and it should return `sum = 5`. It works fine for _CPU_ but works incorrectly for _Cuda_ - it returns `sum = 4`. I know that the algorithm library has such _sum_ reducer already. However I need to implement more complicated reducer algorithms so I started with such simple.

The code is:

```
var kernelSum = accelerator.LoadStreamKernel<
     ArrayView<float>, ArrayView<float>, float>(
     KernelSum);

var memConfig = SharedMemoryConfig.RequestDynamic<float>(2);

kernelSum((1, 2, memConfig), src, trg, 1f);

private static void KernelSum(
	ArrayView<float> src, ArrayView<float> trg, float fScale)
{
	// 1st reduction of the large input array
	int nPos = Group.IdxX;
	int nStride = Group.DimX;

	float fReduce = 0;
	for (var ndx = nPos; ndx < src.Length; ndx += nStride)
		fReduce += src[ndx];

	var dynMem = SharedMemory.GetDynamic<float>();
	dynMem[nPos] = fReduce;

	// The result dynamic memory can be large
	// Each thread will reduce two items from the dynamic memory
	int nDynLng = (int)dynMem.Length;
	for (nStride = nDynLng; nStride > 1; )
	{
		Group.Barrier();

		nDynLng = nStride;
		nStride = (nStride + 1) / 2;

		if (nPos + nStride < nDynLng)
			dynMem[nPos] += dynMem[nPos + nStride];
	}

	Group.Barrier();

	// Reduction finished, let the 1st thread write the result
	if (Group.IsFirstThread)
		trg[0] = fScale * dynMem[0];
}
```

Not sure if there is a bug in my code or something is wrong in the IGL library Cuda part. I debugged the code using _CPU_ accelerator and it looked fine. I did not try with _Cuda_ however.

Maybe this issue is related to algorithm issue I reported:
#https://github.com/m4rs-mt/ILGPU.Algorithms/issues/68

Thanks for any advice :-)

MoFtZ:
@CsabaStupak I have not run the code, just a visual inspection, but the potential issue that stands out to me is the following code:

```CSharp
	// The result dynamic memory can be large
	// Each thread will reduce two items from the dynamic memory
	int nDynLng = (int)dynMem.Length;
	for (nStride = nDynLng; nStride > 1; )
[...]

PR #357, opened 2020-12-12
Fixed returning correct length of dynamic shared memory buffer.

MoFtZ:
Fixes https://github.com/m4rs-mt/ILGPU/issues/356.

It appears that the `ArrayView` returned by `SharedMemory.GetDynamic<T>` currently returns -1 for its length on Cuda and OpenCL accelerators - the CPU accelerator is not affected.

On the Cuda accelerator, there is a [special register](https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#special-registers-dynamic-smem-size) that contains the size of the dynamic memory (in bytes). This needs to be divided by the size of the data element type.

@m4rs-mt I'm not familiar with the OpenCL implementation of SharedMemory. How should this be handled?

### DIFF
--- a/Src/ILGPU/Backends/OpenCL/CLCodeGenerator.Values.cs
+++ b/Src/ILGPU/Backends/OpenCL/CLCodeGenerator.Values.cs
@@ -569,6 +569,52 @@ namespace ILGPU.Backends.OpenCL
                 value,
                 CLInstructions.GetLaneIndexOperation);
 
+        /// <summary cref="IBackendCodeGenerator.GenerateCode(DynamicMemoryLengthValue)"/>
+        public void GenerateCode(DynamicMemoryLengthValue value)
+        {
+            if (value.AddressSpace != MemoryAddressSpace.Shared)
+                throw new InvalidCodeGenerationException();
+
+            // Resolve the name of the global variable containing the length of the
+            // shared dynamic memory buffer.
+            var dynamicView = value.GetFirstUseNode().ResolveAs<Alloca>();
+            var lengthVariableName = GetSharedMemoryAllocationLengthName(dynamicView);
+
+            // Load the dynamic memory size (in bytes) from the dynamic length variable
+            // and divide by the size in bytes of the array element.
+            var target = Allocate(value);
+            using var statement = BeginStatement(target);
+            statement.AppendCast(value.Type);
+            var operation = CLInstructions.GetArithmeticOperation(
+                BinaryArithmeticKind.Div,
+                value.BasicValueType.IsFloat(),
+                out bool isFunction);
+            if (isFunction)
+            {
+                statement.AppendCommand(operation);
+                statement.BeginArguments();
+            }
+            else
+            {
+                statement.OpenParen();
+            }
+
+            statement.AppendCast(dynamicView.ArrayLength.Type);
+            statement.AppendCommand(lengthVariableName);
+
+            if (!isFunction)
+                statement.AppendCommand(operation);
+
+            statement.AppendArgument();
+            statement.AppendCast(value.ElementType.BasicValueType);
+            statement.AppendConstant(value.ElementType.Size);
+
+            if (isFunction)
+                statement.EndArguments();
+            else
+                statement.CloseParen();
+        }
+
         /// <summary cref="IBackendCodeGenerator.GenerateCode(PredicateBarrier)"/>
         public void GenerateCode(PredicateBarrier barrier)
         {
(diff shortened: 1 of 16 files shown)


m4rs-mt:
@MoFtZ Awesome work 🚀 As far as I understood the issue was that we are not aware of the actual length of a dynamic shared-memory allocation inside the compiler and your proposed IR value `DynamicSharedMemoryLengthValue` represents the length of a dynamic allocation. Should we change the value slightly to represent arbitrary dynamic-length arrays instead? I am asking because the function `CreateDynamicAllocaArray` receives an explicit address space and the `DynamicSharedMemoryLengthValue` seems to work for shared-memory allocations only. This can easily lead to programming issue in the future, I think 😟 

Regarding OpenCL: In OpenCL the amount of shared memory is specified via an explicit buffer input definition that receives the size of the buffer in bytes. I am currently not aware of any intrinsic functionality that should allow us to query this information. In the worst case, we have to pass an additional parameter to the kernel and store it in a "global" variable accessible by all functions.

MoFtZ:
Renamed `DynamicSharedMemoryLengthValue` to `DynamicMemoryLengthValue` as per review comments.

Added changes for OpenCL, passing the buffer length as an input parameter to the CLKernel, and then assigning it to a global variable.

Src/ILGPU/Backends/OpenCL/CLInstructions.Data.cs
```diff
@@ -157,6 +157,11 @@ partial class CLInstructions
         /// </summary>
         public const string GetLaneIndexOperation = "get_sub_group_local_id()";
 
+        /// <summary>
+        /// Resolves the current dynamic shared memory size.
+        /// </summary>
+        public const string GetDynamicSharedMemoryLengthOperation = "-1";
+
```
MoFtZ, line 164: Obsolete - should be removed.

Src/ILGPU/Backends/OpenCL/CLCodeGenerator.Values.cs
```diff
@@ -569,6 +569,52 @@ public void GenerateCode(SetField value)
...
 
+        /// <summary cref="IBackendCodeGenerator.GenerateCode(DynamicMemoryLengthValue)"/>
+        public void GenerateCode(DynamicMemoryLengthValue value)
+        {
+            if (value.AddressSpace != MemoryAddressSpace.Shared)
+                throw new InvalidCodeGenerationException();
+
+            // Resolve the name of the global variable containing the length of the
+            // shared dynamic memory buffer.
+            var dynamicView = value.GetFirstUseNode().ResolveAs<Alloca>();
```
m4rs-mt, line 580: Question: Currently, the invariant seems to be the case that the only occurrence of the `DynamicMemoryLengthValue` is the construction of an allocation. Is this guaranteed in all cases? Is there the possibility that we will add another pass in the future that could "rewire" the uses of this node to some other operation?

m4rs-mt:
@MoFtZ Awesome patch 🚀  I added another comment regarding potential issues (I may be wrong here) that could arise in the future 🤓

Src/ILGPU/Backends/OpenCL/CLCodeGenerator.Values.cs
MoFtZ: For now, it's guaranteed. If we use `DynamicMemoryLengthValue` somewhere else in the IR generation, then that could cause issues.

The problem is that `DynamicMemoryLengthValue` is used to construct the `Alloca`, so we cannot pass the ID of the IR node. If we could, then we could decouple the `CLCodeGenerator` from relying on `Alloca`.

Src/ILGPU/IR/Construction/IRBuilder.cs
```diff
@@ -106,6 +106,22 @@ protected IRBuilder(BasicBlock basicBlock)
...
             Append(new LaneIdxValue(GetInitializer(location)));
 
+        /// <summary>
+        /// Creates a node that represents the length of dynamic memory.
+        /// </summary>
+        /// <param name="location">The current location.</param>
+        /// <param name="elementType">The element type.</param>
+        /// <param name="addressSpace">The target address space.</param>
+        /// <returns>A reference to the requested value.</returns>
+        public ValueReference CreateDynamicMemoryLengthValue(
```
m4rs-mt, line 116: Proposal: Maybe we should make this function `private` to the `IRBuilder` to ensure that it will be used in the scope of a dynamic allocation only 🤓 What do you think?

m4rs-mt (approved):

Src/ILGPU/IR/Construction/IRBuilder.cs
MoFtZ: Fixed.

Issue #346, opened 2020-12-06
Row-/Column-major data layouts and the .Net array interop


m4rs-mt:
ILGPU provides built-in 2D and 3D memory buffers (`MemoryBuffer2D` and `MemoryBuffer3D`) and the associated array views (`ArrayView2D` and `ArrayView3D`). They "use" a row-major data layout that perfectly matches the fastest varying indices (FVI) in the scope of GPU kernels (based on `Index3` and and the "default" way of computing global thread indices in Cuda, OpenCL, OpenGL Compute and DirectX Compute). Consider the following piece of code demonstrating the "standard" way of computing a global linear index in memory using a 1D grid of 1D blocks and a 3D grid of 1D blocks:
```c++
__device__ int getGlobalIdx_1D_1D(){
    return blockIdx.x *blockDim.x + threadIdx.x;
}

__device__int getGlobalIdx_3D_1D(){
    int blockId = blockIdx.x + blockIdx.y * gridDim.x
        + gridDim.x * gridDim.y * blockIdx.z;
    int threadId = blockId * blockDim.x + threadIdx.x;
    return threadId;
}
```
Refer to the `simpleIndexing.cu` sample in the Cuda samples directory for more information. The ILGPU index types are structured in the same way, so that Graphics and GPGPU developers can easily benefit from the same "multidimensional index" to "linear index" mapping. The analogy is always: we start processing from the top-left corner of a 2D image and move downwards line by line (or row by row).

However, there are several `CopyFrom` and `GetAsXDArray` functions that consume or return multidimensional arrays of the form `T[,,]`. They currently transpose the input/output arrays. This is due to the reason that `cpuData[x, y]` is intended to be accessible via `gpuData[x, y]`:
```c#
// Setup .Net array buffer
var buff = new int[1024, 4];
buff[17, 2] = 42;

// Initialize GPU buffer
using var t = accl.Allocate(buff);
// Width = 1024
// Height = 4

// Would also yield 42 on a GPU
// int data = t.View[17, 2];

// Yields also 42
var buff2 = t.GetAs2DArray();
int val = buff[17, 2];
```

[...]

PR #370, opened 2020-12-31
Remove ArrayView's size limit of 2^31 - 1.

### LINKED ISSUE #366: Allocate<float>(new LongIndex2(x, y)) fails if x*y does not fit into int
long x = 69;
                    long y = 31527848;
                    Console.WriteLine("Trying to allocate " + (x * y).ToString("0,000") + " floats on GPU");
                    Console.WriteLine("2^31: " + ((long)1 << 31).ToString("0,000"));

                    //var result = gpu.Allocate<float>(new LongIndex2(x, y));
                    // the above throws an exception:
                    //Trying to allocate 2,175,421,512 floats on GPU
                    //2^31: 2,147,483,648
                    //Process terminated.Assertion Failed
                    //32-bit index out of range
                    //at ILGPU.ArrayView2D`1..ctor(ArrayView`2 baseView)
                    //at ILGPU.ArrayView2D`1.op_Implicit(ArrayView`2 baseView)
                    //at ILGPU.Runtime.MemoryBuffer2D`1..ctor(MemoryBuffer`2 buffer)
                    //at ILGPU.Runtime.Accelerator.Allocate[T](LongIndex2 extent)

                    var result = gpu.Allocate<float>(x * y); // this works just fine

Ruberik:
Fixes #366, at least allowing such arrays to be created.

### DIFF
--- a/Src/ILGPU.Tests/ArrayViews.cs
+++ b/Src/ILGPU.Tests/ArrayViews.cs
@@ -583,37 +583,45 @@ namespace ILGPU.Tests
             ArrayView<byte, LongIndex1> data,
             ArrayView<byte, LongIndex3> source)
         {
-            var reconstructedIndex = source.Extent.ReconstructIndex(index.Size * (source.Extent.Size / data.Extent.Size));
+            var reconstructedIndex = source.Extent.ReconstructIndex(
+                index.Size * (source.Extent.Size / data.Extent.Size));
             data[index] = source[reconstructedIndex];
         }
 
         [Theory]
         [InlineData(127)]
-        [InlineData(1300)]  // Makes a 2.04GB MemoryBuffer on the GPU. Prefer not to go larger so this test will pass on more machines.
+        [InlineData(1300)]  // Makes a 2.04GB MemoryBuffer on the GPU.
+        // Prefer not to go larger so this test will pass on more machines.
         [KernelMethod(nameof(ArrayViewSparseMultidimensionalAccessKernel))]
         public void ArrayViewSparseMultidimensionalAccess(long length)
         {
-            // This test allocates over 2GB of memory (with length == 1300), then fills it with 0s and 255s.
-            // Rather than copying the whole array and checking the result, which would take a long time to
-            // check, we'll copy every outputReductionRatio'th element and check that result.
+            // This test allocates over 2GB of memory (with length == 1300), then fills
+            // it with 0s and 255s.
+            // Rather than copying the whole array and checking the result, which would
+            // take a long time to check, we'll copy every outputReductionRatio'th
+            // element and check that result.
             var extent = new LongIndex3(length);
             long outputReductionRatio = 103;
-            using var buffer = Accelerator.Allocate<byte, LongIndex1>(extent.Size / outputReductionRatio);
+            using var buffer = Accelerator.Allocate<byte, LongIndex1>(
+                extent.Size / outputReductionRatio);
             var expectedData = new byte[extent.Size / outputReductionRatio];
             for (int i = 0; i < expectedData.Length; i++)
             {
-                // Find what row of the source element i comes from. Odd rows contain 255, and even rows contain 0.
+                // Find what row of the source element i comes from.
+                // Odd rows contain 255, and even rows contain 0.
                 long sourceRow = (i * outputReductionRatio) / (length * length);
-                expectedData[i] = sourceRow % 2 == 0 ? 0 : byte.MaxValue;
+                expectedData[i] = sourceRow % 2 == 0 ? byte.MinValue : byte.MaxValue;
             }
             using (var source = Accelerator.Allocate<byte, LongIndex3>(extent))
             {
                 // Fill alternating rows with 0 and 255.
                 for (int i = 0; i < length; i++)
                 {
-                    source.MemSet(i % 2 == 0 ? 0 : byte.MaxValue, i * (length * length), length * length);
+                    source.MemSet(i % 2 == 0 ? byte.MinValue : byte.MaxValue,
+                        i * (length * length), length * length);
                 }
-                Execute((int)(extent.Size / outputReductionRatio), buffer.View, source.View);
+                Execute((int)(extent.Size / outputReductionRatio),
+                    buffer.View, source.View);
             }
             Verify(buffer, expectedData);
         }
--- a/Src/ILGPU.Tests/ArrayViews.cs
+++ b/Src/ILGPU.Tests/ArrayViews.cs
@@ -588,44 +588,6 @@ namespace ILGPU.Tests
             data[index] = source[reconstructedIndex];
         }
 
-        [Theory]
-        [InlineData(127)]
-        [InlineData(1300)]  // Makes a 2.04GB MemoryBuffer on the GPU.
-        // Prefer not to go larger so this test will pass on more machines.
-        [KernelMethod(nameof(ArrayViewSparseMultidimensionalAccessKernel))]
-        public void ArrayViewSparseMultidimensionalAccess(long length)
-        {
-            // This test allocates over 2GB of memory (with length == 1300), then fills
-            // it with 0s and 255s.
-            // Rather than copying the whole array and checking the result, which would
-            // take a long time to check, we'll copy every outputReductionRatio'th
-            // element and check that result.
-            var extent = new LongIndex3(length);
-            long outputReductionRatio = 103;
-            using var buffer = Accelerator.Allocate<byte, LongIndex1>(
-                extent.Size / outputReductionRatio);
-            var expectedData = new byte[extent.Size / outputReductionRatio];
-            for (int i = 0; i < expectedData.Length; i++)
-            {
-                // Find what row of the source element i comes from.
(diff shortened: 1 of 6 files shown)


Ruberik:
Sorry for the force-pushes. I tried github's editor for the simple change, but it added the newline, and then I had to flail around with git to fix that and squash it into one commit. It should be good now, assuming the checks pass.

m4rs-mt:
@Ruberik awesome work 🚀 and thank you very much for your PR. This should just work. Should we include additional test cases that test a huge variety of different scenarios in which we "cast" from 32-bit views to 64-bit views and vice versa? 

@MoFtZ what do you think?

Ruberik:
I'm happy to add some test cases. Obviously there's already a test that I'm failing -- maybe a "don't let people make big ArrayViews" test -- but I'll add a couple more no sooner than Monday.

m4rs-mt:
@Ruberik awesome 👍

MoFtZ:
> @Ruberik awesome work 🚀 and thank you very much for your PR. This should just work. Should we include additional test cases that test a huge variety of different scenarios in which we "cast" from 32-bit views to 64-bit views and vice versa?
> 
> @MoFtZ what do you think?

The difficulty I see comes from the limitations of .NET itself - arrays are limited to 32-bit lengths, even with support for `LongLength`. This means that the usual `MemoryBuffer GetAsArray` call will fail on these large arrays.

We recently switched `MemoryBuffer. CopyTo` to use `Span<T>`, but even that appears limited to 32-bit lengths.

Ruberik:
@m4rs-mt I added a test as a starting point for conversation. I realize it fails the presubmit checks, which I can fix, but I'll wait for your feedback so I don't have to interact more with Git than I need to.

Can I ask what you mean about "casting" between 32-bit and 64-bit views?

Ruberik:
@m4rs-mt I realize this isn't passing presubmit checks, but before fixing them I'd like to figure out what would be helpful to see in tests. I'd also like to know whether you think my testing approach is reasonable at all -- making sure the output is small, so it's fast to verify.

Ruberik:
I ended up fixing the presubmit, but I can't reproduce the problem locally. If anyone has any guidance, I'd be happy to take it, but I don't think it's a good use of my time for me to keep hacking at this without support. I'm happy to leave this for somebody else to fix at their convenience.

m4rs-mt:
@Ruberik Sorry for getting so late to you regarding your PR. I took a closer look and the test is really nice 👍🚀.

> I ended up fixing the presubmit problems that existed before, but now there's just a failure in the presubmit that says it crashed with no further information. All the tests pass for me locally. 

Awesome 🚀 Failing Cuda tests on the remote machine can have different issues. I am pretty sure that your change will not cause any problems. I will rerun the tests. The problem should disappear once #376 is merged 🤓 

> Can I ask what you mean about "casting" between 32-bit and 64-bit views?

With casts I referred to the conversion from and to "linearized" versions of the different indices. I have spotted a few locations which are also affected by the same issue (like [this](https://github.com/m4rs-mt/ILGPU/blob/master/Src/ILGPU/IndexTypes.tt#L278) line). This will also affect reconstructing 2D array views from 1D ones and so on. However, I think that we need another solution for the array views regarding their underlying memory layout. This will require a refinement process anyway (I will respond to #346).

Src/ILGPU.Tests/ArrayViews.cs
```diff
@@ -576,6 +576,54 @@ public void ArrayViewMultidimensionalAccess(long length)
...
+            ArrayView<byte, LongIndex3> source)
+        {
+            var reconstructedIndex = source.Extent.ReconstructIndex(
+                index.Size * (source.Extent.Size / data.Extent.Size));
+            data[index] = source[reconstructedIndex];
+        }
+
+        [Theory]
+        [InlineData(127)]
+        [InlineData(1300)]  // Makes a 2.04GB MemoryBuffer on the GPU.
```
m4rs-mt, line 591: Although I like the idea of allocating a larger buffer on the GPU, I am afraid that this can easily lead to serious issues in the CI and on several developer machines. Since multiple test instances will run in parallel, this can easily cause OutOfMemory exceptions and unexpected test crashes. Maybe we should consider doing "dry runs" instead that just simulate a long array view that will be converted into a multidimensional one? (e.g. using the `ArrayView(ArrayViewObject, long index, long extent)` consructor)

m4rs-mt: Sorry again for not getting back to you sooner on this PR 😞.

Ruberik: That's OK! I'm glad we're on it together now.

I share your concern about the test as written. I think I understand what you're suggesting, though as I understand it (please correct me if I'm wrong), the test you're proposing would mostly just test whether it's possible to _construct_ a large ArrayView2D, and not whether it's possible to use it. It's not clear to me that such a test adds much value, since the only thing it's going to check for is that nobody put in a new spurious size check in code called during a constructor. I think it's more interesting to check that nothing is broken end-to-end (which is now done), and that nothing _will become_ broken end-to-end, which I worry isn't possible if we can't actually work with over 2GB of memory. I don't have a good solution.

Src/ILGPU/IndexTypes.tt
```diff
@@ -633,6 +622,22 @@ namespace ILGPU
...
+<# if (def.IsIntIndex) { #>
+        /// <summary>
+        /// Reconstructs a 1D index from a linear index.
+        /// </summary>
+        /// <param name="linearIndex">The linear index.</param>
+        /// <param name="dimension">The 1D dimension for reconstruction.</param>
+        /// <returns>The reconstructed 1D index.</returns>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static <#= name #> ReconstructIndex(
+            long linearIndex,
```
Ruberik, line 634: Some tests would be appropriate, I think, for all of these guys. Should I put them in a new file, Indices.cs, or in ArrayViews.cs?

Src/ILGPU.Tests/ArrayViews.cs
m4rs-mt: Yes, I totally agree that such a test may be of no use at all 🤔 Maybe we can work around this problem by allocating a much smaller buffer, adjusting the base pointer to point to `basePointer - 2GB` in GPU memory, and try to create an `ArrayView` instance (bypassing the existing assertions) to simulate a much larger view in order to index elements above the `2GB` boundary. However, I am not sure how to implement such a test without too much effort....

Src/ILGPU/IndexTypes.tt
m4rs-mt: This might also be worth investigating. Related to the question of whether we should write additional tests for each of them, is the discussion in #346, I will comment on in the next few days. I had an offline discussion with several people from the web and the community about possible ways to implement a significantly better API that could affect the current implementation of all `ArrayViewXD`, `IndexXD` and `LongIndexXX` types. This also raises the question on how much effort we should put into the tests discussed above.

Src/ILGPU.Tests/ArrayViews.cs
Ruberik: Yeah, I guess testing construction and indexing is better than nothing, but I share your concerns about effort... and I worry that it might spuriously fail in other situations when very reasonable assertions are added.

Src/ILGPU/IndexTypes.tt
Ruberik: I added a few simple tests, just for IndexXD, that I think are better than nothing.

Ruberik:
I added a few tests for reconstructing indices. It isn't clear to me how best to test large ArrayViews if we can't actually make a big ArrayView. My thought is that it requires as way of:
- Creating tests that aren't run in parallel with the others
- Creating tests that are only run on machines with lots of memory, or in the presubmit

That combination would let us write a test like the one I originally put in, or something more comprehensive. Without actually being able to make a large MemoryBuffer to test it, I don't think there's a lot we can do without uncomfortable and difficult hacks.

m4rs-mt (approved):

m4rs-mt:
@Ruberik I fully agree. Let's see how the discussion about different memory setups continues. We can always come back to this point and try to find additional ways to test this.

Nit: requires minor polishing by removing the conflicts. After that this should be good to do 🚀

Ruberik:
Great, thanks for the approval!

> Nit: requires minor polishing by removing the conflicts. After that this should be good to do 🚀

I don't see the conflicts, either in the thread or in the "Files changed" list. Where do you see them?

m4rs-mt:
>  I don't see the conflicts, either in the thread or in the "Files changed" list. Where do you see them?

Hm... the GitHub web interface just tells me that there are conflicts which makes it impossible to rebase this PR on top of the master branch. I can do this manually. However, you might want to rebase it yourself and force push your changes 😃

Ruberik:
That's just bizarre. I can see that I'm over a hundred commits behind, but the web interface isn't offering to merge or let me settle conflicts. I'll dig into it tomorrow.

Ruberik:
I *think* it's up-to-date now. If I never interact with Git again, it will be too soon.

PR #373, opened 2021-01-05
Fixed edge cases in LoopUnrolling

m4rs-mt:
The current `LoopUnrolling` implementation has a critical bug related to nested conditionals with side effects. Without this PR, ILGPU can crash using `O1` and `O2`.

### DIFF
--- a/Src/ILGPU/IR/Analyses/LoopInfo.cs
+++ b/Src/ILGPU/IR/Analyses/LoopInfo.cs
@@ -212,7 +212,7 @@ namespace ILGPU.IR.Analyses
             out InlineList<(PhiValue, Value)> phiValues)
         {
             var phis = loop.ComputePhis();
-            var phiInductionsVariables = new HashSet<PhiValue>();
+            var visitedPhis = new HashSet<PhiValue>();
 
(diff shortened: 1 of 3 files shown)


Src/ILGPU/IR/Analyses/LoopInfo.cs
```diff
@@ -234,21 +234,36 @@ out InlineList<(PhiValue, Value)> phiValues)
...
                     branch));
-                phiInductionsVariables.Add(phiValue);
+                visitedPhis.Add(phiValue);
             }
 
-            // Check all
+            // Check all phi values
             foreach (var phi in phis)
             {
                 // Check whether this phi is an induction variable
```
MoFtZ, line 243: QUESTION: Is this comment about `Check whether this phi is an induction variable` still valid?

MoFtZ (approved):

Src/ILGPU/IR/Analyses/LoopInfo.cs
m4rs-mt: Nice one 👍 Fixed invalid comment 🤓

PR #380, opened 2021-01-11
Fixed support for SizeOf instruction.

MoFtZ:
As reported in https://github.com/m4rs-mt/ILGPU.Algorithms/issues/73, ILGPU is able to load the [Sizeof](https://docs.microsoft.com/en-us/dotnet/api/system.reflection.emit.opcodes.sizeof?view=net-5.0) opcode, but does not translate the instruction.

This PR attempts to get the `Matrix4x4.Decompose` function to compile in an ILGPU kernel.

### DIFF
--- a/Src/ILGPU/Frontend/CodeGenerator/Objects.cs
+++ b/Src/ILGPU/Frontend/CodeGenerator/Objects.cs
@@ -9,6 +9,7 @@
 // Source License. See LICENSE.txt for details
 // ---------------------------------------------------------------------------------------
 
+using ILGPU.Backends;
 using ILGPU.IR;
 using ILGPU.IR.Values;
 using ILGPU.Resources;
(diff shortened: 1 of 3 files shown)


Src/ILGPU/Frontend/CodeGenerator/Objects.cs
```diff
@@ -125,5 +125,30 @@ private void MakeStoreObject(Type type)
...
+            // parameter can be used only in the body of the type or method that defines
+            // it. When that type or method is instantiated, the generic type parameter
+            // is replaced by a value type or reference type.
+            if (type.IsValueType)
+            {
+                Load(Interop.SizeOf(type));
+            }
+            else
+            {
+                Load(IntPtr.Size);
```
m4rs-mt, line 150: I really like the simplicity of this PR to implement the `sizeof` MSIL instruction. However, I am wondering whether this is really correct. In my opinion, loading the size of a pointer should depend on the current `TargetPlatform` (since it is possible to construct an `X86` context on an `X64` machine).

MoFtZ: > (since it is possible to construct an `X86` context on an `X64` machine).

Ah, I see. I did not know we could do this. I had assumed that with the call to `EnsureRunningOnNativePlatform`, it always had to match. Looking at the code again, looks like it only applies to auto grouped kernels.

m4rs-mt: Good point! I guess we should remove this functionality in the future to simplify the compiler internals in general 🤓

m4rs-mt (approved):

m4rs-mt:
Nit: needs a minor adjustment since I have merged #393 first 😢

m4rs-mt (approved):
Awesome work 🚀

PR #381, opened 2021-01-11
Changed to official T4 text transformation tool.

MoFtZ:
The [official T4 text transformation tool](https://github.com/mono/t4) used to have [an issue compiling our T4 templates](https://github.com/m4rs-mt/ILGPU/pull/117#discussion_r428334222). They recently released a new version which appears to work as expected.


MoFtZ:
Damn, it doesn't work - looks like the T4 tool requires the .NET 5.0 SDK to be installed.

PR #405, opened 2021-02-02
Reverses a backwards assertion in ArrayView.cs

Ruberik:
The extent of the new ArrayView should surely be smaller than the extent of the one underlying it, as opposed to larger.


m4rs-mt (approved):
Excellent 👍

PR #413, opened 2021-02-07
SPIR-V Backend

Yey007:
I've started work on a SPIR-V backend. This is probably way more than I can handle alone, but I'm hoping I can get everyone's advice or we can work together.


Yey007:
Oh, looks like I accidentally changed some PTX files. I'll revert that later.

Yey007:
I should probably also clarify that I have no idea whether any of this works yet. It's sort of hard to test without building out almost the entire backend and the runtime.

m4rs-mt:
@Yey007 Looks pretty good as a first starting point 🚀. However, I guess we need to issue assembled SPIR-V instructions, right? My understanding is that most drivers only accept assembled binary shaders.

Yey007:
Hmm, that's a valid point. My plan was to just use the official SPIR-V assembler from [SPIR-V Tools](https://github.com/KhronosGroup/SPIRV-Tools) after bytecode is emitted, but given we want to keep dependencies low, that may not be an option. I'll have to look into how we can assemble these ourselves. I doubt any drivers can assemble SPIR-V on their own.

Yey007:
Perhaps it's best to have two implementations. We can start by leveraging SPIR-V tools, then look into assembling the instructions later (but before merging to master).

m4rs-mt:
I guess it should not be too hard to use the SPIR-V byte code grammar from the [official GitHub repo](https://www.khronos.org/registry/spir-v/#spec) and generate an assembler ourselves. We can then use the SPIR-V disasm tools to first verify our implementation and second inspect the generated assembler code. However, we could also automatically generate two backends, one emitting strings and the other one emitting byte code using generics or an abstract code generator.

Yey007:
Sorry, I think I may have some terminology mixed up. So is there supposed to be another layer between what we're emitting now (the string instructions) and the binary which the drivers take? Are you saying we should just emit the binary instead of the string instructions?

m4rs-mt:
Sorry for the confusion. I was thinking that we could build another thin abstraction layer between the actual backend (including its code generator) and the actual instructions generated. This would allow us to switch between the string and binary representation without touching the backend (or its associated code generator).

Yey007:
I see. We could perhaps put all the instructions into a list of `SPIRVHeaders.Op`s and then compile that accordingly. That does raise the question of how to handle operations that take arguments, but I'm sure there is a way we can solve that. One idea that comes to mind is to have an `SPIRVOp` class which contains the OpCode for that instruction and an array/list of arguments (references to other instances of the `SPIRVOp` class)

m4rs-mt:
Alternatively, we can try to automatically generate a `SPIRVBuilder` class that has a generation method for each `Op`. Each method then takes the appropriate arguments as parameters and emits them to the underlying output stream.

Yey007:
That's probably the better idea of the two. The JSON grammar file looks comprehensive and should do the job. This also looks like the perfect task for the fancy new source generators! Does ILGPU use C# 9 yet?

Yey007:
I also think maybe we should generate an `ISPIRVBuilder` in addition to implementations for both string and assembly output. What do you think?

Yey007:
I've generated an ISPIRVBuilder interface that we can implement in different ways (bytecode, string representation, etc.) from the grammar file. I haven't included the file in the commit, but can do so if necessary. I'm also not sure what's up with the two blank lines at the start of the templates. I'll need to look into that.

Yey007:
Looks like it's also failing to build the templates, which is interesting, because it seemed to work fine on my end.

Yey007:
Hey @m4rs-mt I just wanted to give you an update on what's going on. In short, I cannot get the text template to work with the CI/`dotnet msbuild`/`dotnet t4`. It works with Visual Studio and Rider, but not with the tool for some odd reason. @MoFtZ also gave it a try, and reported the same behavior. The tool reports that it cannot find some of the necessary classes (in `System.Runtime.Serialization.Json`) even though it has no problem finding the assembly and importing the namespace. I've done a lot of research but found nothing and am unsure on how to move forward with the PR if this doesn't work. Could you perhaps take a look as well? Try transforming `ISPIRVBuilder.tt` with `dotnet t4` or `dotnet msbuild`.

Yey007:
One option is to go back to the old approach, just swapping out the string generation with actual bytecode generation. It still seems a little less convenient, but doable.

Yey007:
Closing in favor of multiple, smaller PRs as suggested by @m4rs-mt.

PR #421, opened 2021-02-17
NextGen Memory API

m4rs-mt:
This (draft) PR presents a preview of "the" new memory API, based on the discussion in #346 and mainly on @Joey9801's proposal. It refines the existing `Index`, `LongIndex`, `ArrayView`, `MemoryBuffer`, `MemoryBufferCache` and `ExchangeBuffer` types. The new API distinguishes between a coherent, strongly typed `ArrayView<T>` structure and its n-D versions `ArrayViewXD<T, TStride>`, which carry dimension-dependent stride information (The actual logic for computing element addresses is moved from the `IndexXD` types to the newly added `StrideXD` types). This allows developers to explicitly specify a particular stride of a view, `reinterpret` the data layout itself (by changing the stride), and perform compile-time optimizations based on explicitly typed stride information. Consequently, ILGPU's optimization pipeline is able to remove the overhead of these abstractions in most cases (except in rare use cases where strange-looking strides are used). It also makes all memory transfer-related operations explicit in terms of what memory layout the underlying data will have after an operation is performed.

In addition, this PR moves all `copy` related methods to the `ArrayView` instances instead of exposing them on the memory buffers. This realizes a "separation of concerns": One the one hand, a `MemoryBuffer` holds a reference to the native memory area and controls its lifetime. On the other hand, `ArrayView` structures manage the contents of these buffers and make them available to the actual GPU kernels.

Example:
```c#
// Simple 1D allocation of 1024 longs with TStride = Stride1D.Dense (all elements are accessed contiguously in memory)
var t = accl.Allocate1D<long>(1024);

// Advanced 1D allocation of 1024 longs with TStride = Stride1D.General(2) (each memory access will skip 2 elements)
// -> allocates 1024 * 2 longs to be able to access all of them
var t = accl.Allocate1D<long, Stride1D.General>(1024, new Stride1D.General(2));

// Simple 1D allocation of 1024 longs using the array provided
var data1 = new long[1024];
var t2 = accl.Allocate1D(data1);

// Simple 2D allocation of 1024 * 1024 longs using the array provided with TStride = Stride2D.DenseX
// (all elements in X dimension are accessed contiguously in memory)
// -> this will *not* transpose the input buffer as the memory layout will be identical on CPU and GPU
var data2 = new long[1024, 1024];
var t3 = accl.Allocate2DDenseX(data2);

// Simple 2D allocation of 1024 * 1024 longs using the array provided, with TStride = Stride2D.DenseY
// (all elements in Y dimension are accessed contiguously in memory)
// -> this *will* transpose the input buffer to match the desired data layout
var data3 = new long[1024, 1024];
var t4 = accl.Allocate2DDenseY(data3);

```

Please note that the current changes do NOT pass integration testing and are NOT intended to compile and run kernel programs.
The changes are only an API preview to investigate the "look and feel" of the (potentially) next generation memory API.

Please also note that this is a draft PR and should not be considered complete in terms of API interfaces.

I will add more information to this PR about any changes and provide more examples 🤓


Src/ILGPU/Runtime/ArrayViewExtensions.cs
```diff
@@ -0,0 +1,1072 @@
...
+        public static void CopyToCPUSync<T, TView>(
+            this TView source,
+            AcceleratorStream stream,
+            out T cpuData,
+            long length)
+            where TView : IContiguousArrayView<T>
+            where T : unmanaged
+        {
+            source.CopyToCPUSync(stream, out cpuData, length);
+            stream?.Synchronize();
```
MoFtZ, line 509: QUESTION: Can `stream` ever be `null`? i.e. when would a caller pass `null`?

MoFtZ (commented):
@m4rs-mt
Overall, this patch is looking great 🚀

Personally, I'm finding the new style of declaring ArrayViews with Stride a bit too verbose. However, I'm wondering if this is more about familiarity more than anything else. I definitely see the flexibility it provides, and it addresses the key concern about transposing the memory buffer on copying 👍 

I have a few questions regarding the changes for `CopyFromCPUSync` and `CopyFromCPUAsync` (and the corresponding `CopyToX` functions).

QUESTION: What issues are the Sync/Async functions trying to address? The split seems to be around synchronisation of the accelerator stream, but it's not clear when either should be used. e.g. how would someone migrating from the old API to this new API decide which to use?

QUESTION: The previous Copy From/To methods provided the ability to copy small sections. How does that fit into the new API? Is the intention that the Stride iterator will deal with this?

Src/ILGPU/Runtime/ArrayViewExtensions.cs
m4rs-mt: The main "isuse" is that it is possible to create objects that are logically associated with a `CPUAccelerator` instance without explicitly instantiating a `CPUAccelerator` (e.g. an `UnmanagedMemoryBuffer`). Since there is no instance of a parent accelerator, there is no associated stream. What we could do is to use a static singleton instance of a `CPUStream` (since it doesn`t need to perform any special operations anyway) to avoid these null checks and to ensure that there are no edge cases where the `AcceleratorStream` is `null`.

m4rs-mt:
Thank you for your very detailed feedback on this patch. I generally agree that the patch makes `ArrayView` handling a bit more verbose compared to the "original" approach. Therefore, I created additional `MemoryBufferXD` utility classes that reduce the verbosity level. Example code:
```c#
MemoryBuffer<int, ArrayView1D<int, Stride1D.Dense>> x = accelerator.Allocate1D(...);
// can be rewritten to
MemoryBuffer1D<int, Stride1D.Dense> x = accelerator.Allocate1D(...);
// since MemoryBuffer1D<T, TStride> inherits from MemoryBuffer<T, ArrayView1D<T, TStride>>.
```
Please also note that there are implicit conversions between `ArrayView<T>` and `ArrayView1D<T, Stride1D.Dense>` to simplify programming and porting efforts.

However, from now on you must explicitly include all stride information to avoid confusion about the memory layout (and access patterns in your kernels). This is especially important in the case of 2D and 3D buffers and/or views. Also, users can perform all copy/memset/etc. operations directly on views without having to "use" the parent memory buffer they belong to.

> What problems are the sync/async functions trying to solve? The split seems to be around the synchronization of the accelerator stream

The basic idea was to explicitly include synchronization in the scope of API calls. `xxxSync` will always perform synchronization on the current/given `AcceleratorStream`, while `xxxAsync` will not block. Currently, almost all methods that copy to/from CPU memory regions force accelerator stream synchronization for safety reasons. However, it is still possible to shoot yourself in the foot with the new API.... 🤔 

> How would someone moving from the old API to this new API decide which one to use?

Good question. I think the `xxxSync` methods should work fine in most cases. However, I have to admit that there are no additional overloads any more to copy smaller memory chunks between CPU and GPU, which brings us to your second question: 

> The previous Copy From/To methods provided the ability to copy small sections.  How does that fit into the new API? 

My idea was to provide additional overloads that use `Span<T>` to reduce the number of overloads required.

m4rs-mt:
I have included all feedback from the community regarding the new Memory API interface.
The main changes/features of this PR are:
- `Index1`|`Index2`|`Index3` types have been renamed to `Index1D`|`Index2D`|`Index3D` to match the naming scheme of `ArrayViewXD` and `MemoryBufferXD` types.
- `LongIndex1`|`LongIndex2`|`LongIndex3` types have been renamed to `LongIndex1D`|`LongIndex2D`|`LongIndex3D` to match the naming scheme of the `ArrayViewXD` and `MemoryBufferXD` types.
- Separation of concerns between `MemoryBuffer` and `ArrayView` instances:
    - `ArrayView...` structures represent and manage the contents of buffers (or chunks of buffers).
    - `MemoryBuffer...` classes manage the lifetime of allocated memory chunks on a device.
- The `ILGPU.ArrayView` intrinsic structure implements the newly added `IContiguousArrayView` interface that marks contiguous memory sections.
- The `ILGPU.Runtime.MemoryBuffer...` classes implement the newly added `IContiguousArrayView` interface that marks contiguous memory sections.
- Types implementing the `IContiguousArrayView` interface provide extension methods for initializing, copying from and to the memory region (not supported on accelerators).
- This PR adds the notion of `Stride`s. ILGPU contains built-in common strides for 1D, 2D and 3D views.
    - `Stride1D.Dense` represents contiguous chunks of memory that pack elements side by side.
    - `Stride1D.General` represents strides that skip a certain number of elements.
    - `Stride2D.DenseX` represents 2D strides that pack elements side by side in dimension X (transfers from a to views with this stride involve transpose operations).
    - `Stride2D.DenseY` represents 2D strides that pack elements in the Y dimension side by side.
    - `Stride2D.General` represents strides that skip a certain number of elements in the X and Y dimensions.
    - `Stride3D.DenseXY` represents 3D strides that pack elements in the X,Y dimension side by side (transfers from a to views with this stride involve transposition operations).
    - `Stride3D.DenseZY` represents 3D strides that pack elements in the Z,Y dimension side by side.
    - `Stride3D.General` represents strides that omit a certain number of elements in the X, Y and Z dimensions.
- All `ArrayViewXD` types have been moved to the `ILGPU.Runtime` namespace.
- All `ArrayViewXD` types do *not* implement `IContiguousArrayView`, as they support arbitrary stride information.
  Note that the `ArrayView1D<T, Stride1D.Dense>` specialization has an implicit conversion to `ArrayView<T>` (and vice versa) for auxiliary purposes.
- All `CopyFromCPU` and `CopyToCPU` methods are provided with additional hints as to whether they are transposing the input elements or keeping the original layout.
- Note that `GetAsXDArray(...)` always returns elements in .Net standard layout for 1D, 2D and 3D arrays (this may result in transposing the input elements of the buffer on the CPU).
  Use `view.AsContiguous().GetAsArray()` to get the memory layout of the input buffer.

Src/ILGPU/Runtime/ArrayViewExtensions.Generated.tt
```diff
@@ -0,0 +1,800 @@
...
+        /// <remarks>
+<#      if (needsTranspose) { #>
+        /// CAUTION: this method transposes the data on the CPU.
+<#      } #>
+        /// This method is not supported on accelerators.
+        /// </remarks>
+        [NotInsideKernel]
+        public static void CopyFromCPU<T>(
+            this <#= typeName #><T, <#= strideTypeName #>.<#= strideName #>> view,
+            <#= arrayType #> data)
```
MoFtZ, line 550: ISSUE: Should replace the overloads of `CopyFromCPU`/`CopyToCPU` that use `arrayType` with `ReadOnlySpan<T>`.

This will provide support for arrays, as well as spans.

MoFtZ (changes requested):

MoFtZ (commented):
MINOR: Should also run `Code Cleanup` in VS2019.

Src/ILGPU/Runtime/ArrayViewExtensions.Generated.tt
m4rs-mt: What about multidimensional arrays? Should we still emit extension methods taking `Span`s instead?

MoFtZ: ah, true. did not consider that. it would also not work for jagged arrays.

m4rs-mt:
I received an issue report from @Joey9801 that the `Cast` methods are not matching adjusting `Stride`/`Extent` information properly. We had a short discussion offline and came up with the idea to replace the "generic" `Cast` methods with specialized extension methods for individual `ArrayView[2D|3D]<T, Stride>` specializations.

Src/ILGPU.Tests/Generic/TestBase.cs
```diff
@@ -106,6 +106,7 @@ protected TestBase(ITestOutputHelper output, TestContext testContext)
             params object[] arguments)
             where TIndex : struct, IIndex
         {
+            Accelerator.Synchronize();
```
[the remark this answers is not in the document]
m4rs-mt: I added this synchronization point to avoid potential problems related to other operations that use the default accelerator current. These have synchronized before running a test kernel using a different stream 🤓

m4rs-mt:
> I received an issue report from @Joey9801 that the `Cast` methods are not matching adjusting `Stride`/`Extent` information properly. We had a short discussion offline and came up with the idea to replace the "generic" `Cast` methods with specialized extension methods for individual `ArrayView[2D|3D]<T, Stride>` specializations.

I added new `Cast` extension methods to specialized `ArrayView` instances to solve the above problem. I also removed the misleading `Compute...Extent` methods from the `Index` type definitions.

MoFtZ (approved):

PR #423, opened 2021-02-19
Avoid using SuppressMessage on partial methods

marcin-krystianc:
This change is a workaround for the Roslyn issue with processing `SuppressMessage` attributes on partial methods (https://github.com/dotnet/roslyn/issues/50496).
 
 The `SuppressMessage` attribute for `CA1822` diagnostic was necessary because another bug, this time in RoslynAnalyzers (https://github.com/dotnet/roslyn-analyzers/issues/4733). Due to this bug, the `CA1822` diagnostic is generated for any partial method with one or more attributes.

### Fix
- Removing the empty implementation of partial methods in `Release` configuration makes the `[Conditional("DEBUG")]` attribute redundant. Note that it is ok to not provide any implementation for partial method, because in such case [compiler will remove the method entirely](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/partial-method). 
- Removing the `[Conditional("DEBUG")]` attribute makes the `SuppressMessage` redundant

### DIFF
--- a/Src/ILGPU/IR/BasicBlockMapping.cs
+++ b/Src/ILGPU/IR/BasicBlockMapping.cs
@@ -14,6 +14,8 @@ using ILGPU.IR.Analyses.TraversalOrders;
 using System;
 using System.Collections;
 using System.Collections.Generic;
+using System.Diagnostics;
+using System.Diagnostics.CodeAnalysis;
 using System.Runtime.CompilerServices;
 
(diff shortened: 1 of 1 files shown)


m4rs-mt:
@marcin-krystianc Awesome patch 👍

m4rs-mt (approved):

PR #426, opened 2021-02-26
Fixed retrieving field from structure types.

### LINKED ISSUE #418: Struct member variable after fixed array cannot be used in comparison operation.
I have reduce the code from issue #417 into a minimal example of the problem.

When `Cursor` is defined after `Array`, ILGPU will fail to compile. When `Cursor` is defined before `Array`, it compiles without issue.

```CSharp
public unsafe struct ArrayStruct
{
    public fixed int Array[2];
    public int Cursor;
}

private static void FailingKernel(Index1 index, ArrayView<int> view)
{
    var a = new ArrayStruct();
    if (a.Cursor > 0)
    {
        view[index] = index;
    }
}

static void Main()
{
    using var context = new Context();
    using var accelerator = new CudaAccelerator(context);

    accelerator.LoadAutoGroupedStreamKernel<Index1, ArrayView<int>>(FailingKernel);
}
```
The exception thrown is:
`NotSupportedException: Not supported compare or arithmetic value of type 'None'`

MoFtZ:
Fixes #418.

One of the code paths was missing the code for remapping structure field indices.

### DIFF
--- a/Src/ILGPU.Tests/StructureValues.cs
+++ b/Src/ILGPU.Tests/StructureValues.cs
@@ -426,6 +426,52 @@ namespace ILGPU.Tests
             var expected = new int[] { 42 };
             Verify(output, expected);
         }
+
+        public unsafe struct FieldStruct
+        {
+            public fixed int Array[3];
+            public int Cursor;
+        }
+
+        internal static void StructureFixedBufferGetFieldKernel(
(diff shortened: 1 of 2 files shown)


m4rs-mt (approved):
Awesome work 🚀

PR #431, opened 2021-03-12
Use T4.Build to automatically transform T4 text templates during build

jgiannuzzi:
This PR replaces the way that T4 text templates get transformed.

They are currently either transformed manually by clicking around in Visual Studio, or transformed manually by invoking the right MSBuild task from the command line. The latter is what the CI uses, and it also requires installing the `dotnet-t4-fix` tool.

This PR instead uses the [T4.Build](https://github.com/jgiannuzzi/T4.Build) package to automatically transform the T4 text templates during build, only when they have changed, and in parallel (so more CPU cores => faster 🚀 ). The templates can also be cleaned with a regular `dotnet clean`. This also works automatically in Visual Studio, so there is no need to ever click 'Transform All T4 Templates' anymore 💪.

Disclaimer: I am the author of [T4.Build](https://github.com/jgiannuzzi/T4.Build).


m4rs-mt:
@jgiannuzzi I took a look at your PR and this is really cool stuff 😍 After you solve the Windows-related issues, this should be good to go. However, I don't know if we need to adapt this PR to solve the problem. My understanding is that it should be sufficient to solve this problem within the "T4 tool magic" of yours.

Yey007:
This sounds amazing! I would love this.

jgiannuzzi:
> After you solve the Windows-related issues, this should be good to go. However, I don't know if we need to adapt this PR to solve the problem. My understanding is that it should be sufficient to solve this problem within the "T4 tool magic" of yours.

This is actually just a transient error with the GitHub Actions runner that had network connectivity issues - if you re-run the workflow, it should just work.

In the meantime, I have been working on some improvements for T4.Build to solve the issue where the macOS CI job was sometimes 'stuck'. I'll update this PR with a new version once I release it.

m4rs-mt:
@jgiannuzzi Hmm... the problem still seems to exist 😢.

jgiannuzzi:
@m4rs-mt I might be wrong but I really think this is transient — if you look closely at the new run, you'll see that the error is slightly different. Anyway, don't bother re-running it now as I would like to update the PR before you consider merging it 😄

.github/workflows/ci.yml
```diff
@@ -83,13 +83,6 @@ jobs:
           dotnet-version: 3.1.407
       - name: Pin .NET Core SDK
         run: dotnet new globaljson --sdk-version 3.1.407
-      - name: Restore tools
-        run: dotnet tool restore
-        working-directory: ./Src
```
m4rs-mt, line 88: Are you sure that these lines here are not responsible for the current problems we are having on the Windows build machines?

m4rs-mt: I only ask because the Windows CI build is still not able to find dotnet-xunit related things.... only on Windows.

jgiannuzzi: I have to say, I'm starting to think it might be related 😓
I'll investigate in my repo.

jgiannuzzi: I poked around a bit more and managed to reproduce this issue with the current master branch too. It seems to be quite random, but always happening on the Windows runners, and only when the Visual Studio offline NuGet source is enabled (which is the default). Disabling this NuGet source seems to solve the issue permanently, and I would recommend we do that, at least until GitHub fixes the Windows runner (which seems to have other issues in terms of NuGet restore, as it usually takes about 3 times longer than on Linux).

I've added a commit doing just that in this branch, but I'm happy to create a separate PR for it if you'd like.

jgiannuzzi:
Apart from that, the time to compile all text templates on my machine went from 1m10s with the previous manually invoked method down to just 3s with T4.Build v0.2.0 (which uses in-process template compilation through the Roslyn libs installed with the SDK, which have been AOT compiled down to native code).

I have tested this on Linux, Windows, and macOS on both .NET Core 3.1 and .NET 5.0 SDKs and feel pretty confident this can now be merged.

Issue #429, opened 2021-03-11
Bounds checking generates spurious failures in large views


Ruberik:
I'm seeing a bounds checking failure under the following conditions:

* A very large MemoryBuffer (9,813,945,198 bytes in this case, but the specific size matters).
* Getting element (1<<31) - 68 of that MemoryBuffer, using an int index (I don't necessarily expect a predictable pattern to which numbers work and don't work).

This might seem pretty esoteric, but it came up for me in a scenario where I was batch-copying elements from one notionally two-dimensional MemoryBuffer to another, in batches of size slightly under 2^30. The first batch worked fine, but the second batch didn't.

This took a long time to track down, partly because the specific size of your very large MemoryBuffer matters. I get this failure if my MemoryBuffer is size `642*15,286,519 = 9,813,945,198`, but not if it's size `642*17,286,519 = 11,097,945,198` (bigger) or size `642*10,286,519 = 6,603,945,198` (smaller).

Note that the actual *memory access* seems to work fine if I turn off the assertion. Here's the code:

```
        static void Main(string[] args) {
            var contextFlags = Context.DefaultFlags | ContextFlags.EnableAssertions;
            //var contextFlags = Context.DefaultFlags;
            using var AcceleratorContext = new Context(contextFlags, ILGPU.IR.Transformations.OptimizationLevel.O2);
            AcceleratorContext.EnableAlgorithms();
            var Accelerator = new CudaAccelerator(AcceleratorContext);
            var putKernel = Accelerator.LoadStreamKernel<int, int, long, ArrayView2D<byte>>(PutKernel);
            var getKernelNoSubView = Accelerator.LoadStreamKernel<int, int, long, ArrayView<byte>, ArrayView<byte>>(GetKernelNoSubView);

            int subviewYSize = 1672495;
            int subviewXSize = 642;
            int yOffset = 1672495;
            int itemCount = 15286519;

            Console.WriteLine(((subviewYSize + (int)yOffset) - 1) * subviewXSize + subviewXSize - 1);  // int index of the item to set/get
            Console.WriteLine(itemCount);
            using var arr = Accelerator.Allocate<byte>(subviewXSize, itemCount);
            arr.MemSetToZero();
            using var output = Accelerator.Allocate<byte>(1);
            Accelerator.Synchronize();
            putKernel(new KernelConfig(1, 1, SharedMemoryConfig.Empty), subviewYSize, subviewXSize, yOffset, arr);
            Accelerator.Synchronize();
[...]

PR #433, opened 2021-03-15
Fixed bounds check on large views.

MoFtZ:
Fixes https://github.com/m4rs-mt/ILGPU/issues/429.

When injection assertions into Cuda kernels, the `index` should be converted to `long`. Otherwise, if the length of the `ArrayView` is larger than `int.MaxValue`, the upper 32 bits will be truncated.

QUESTION: What should we do about the differences between CPU and Cuda accelerators?
On the CPU, the assertion will check if `ArrayView.Length` is less than `int.MaxValue`. This is not checked on the GPU.

### DIFF
--- a/Src/ILGPU/ArrayView.cs
+++ b/Src/ILGPU/ArrayView.cs
@@ -229,11 +229,7 @@ namespace ILGPU
         {
             [MethodImpl(MethodImplOptions.AggressiveInlining)]
             [ViewIntrinsic(ViewIntrinsicKind.GetViewElementAddressByIndex)]
-            get
-            {
-                IndexTypeExtensions.AssertIntIndex(Length);
-                return ref this[(long)index.X];
-            }
+            get => ref this[(long)index.X];
         }
 
         /// <summary>
@@ -257,11 +253,7 @@ namespace ILGPU
         {
             [MethodImpl(MethodImplOptions.AggressiveInlining)]
             [ViewIntrinsic(ViewIntrinsicKind.GetViewElementAddress)]
-            get
-            {
-                IndexTypeExtensions.AssertIntIndex(Length);
-                return ref this[(long)index];
-            }
+            get => ref this[(long)index];
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


m4rs-mt:
@MoFtZ I don't think we can generally do anything about these differences when using the .Net debuggers; since we want to run the original code. At this stage of the compiler, we can't use the IL front end properly because we are already generating code...

m4rs-mt (approved):

m4rs-mt:
@MoFtZ Should I merge this PR or wait until we have discussed how to proceed regarding the different assertion checks?

MoFtZ:
I'm happy to wait. There is still time before the next release of ILGPU.

I'm wondering if we should make the CPU assertions match the GPU assertions? Rather than using `AssertIntIndexRange`, which causes the CPU code to assert more often, we check that the `index` is within the length of the ArrayView.

m4rs-mt:
I think you are absolutely right and we should reconcile these assertions.

Src/ILGPU/ArrayView.cs
```diff
@@ -229,11 +229,7 @@ public ArrayView(ArrayViewSource source, long index, long length)
         {
             [MethodImpl(MethodImplOptions.AggressiveInlining)]
             [ViewIntrinsic(ViewIntrinsicKind.GetViewElementAddressByIndex)]
-            get
-            {
-                IndexTypeExtensions.AssertIntIndex(Length);
```
m4rs-mt, line 234: QUESTION: Don't we have to adapt these assertions to match the ones you have added to the code generator?

MoFtZ: The assertion already exists in the ArrayView's indexer that accepts a `long` parameter.

https://github.com/m4rs-mt/ILGPU/blob/20342179ec67ffddce8bf9d421339cc72f4bbac2/Src/ILGPU/ArrayView.cs#L264-L277

m4rs-mt (approved):

PR #435, opened 2021-03-16
Gracefully fail when loading symbols from in-memory assemblies.

MoFtZ:
When using Roslyn to compile code at run-time, if the `Assembly` is in-memory, it is not considered to be "dynamic" and has an empty location. This PR gracefully ignores assemblies without a location (skipping symbols), rather than throwing an exception and preventing the kernel from loading.

NOTE: `Assembly.Load` has an overload for passing the PDB bytes, but it is currently unknown how to retrieve the embedded symbols from the assembly.


m4rs-mt (approved):
Nice patch 👍 This allows us to load in-memory assemblies and enable debug symbols at the same time  🚀

Issue #425, opened 2021-02-25
Math.Round(double) causes internal compiler error on .NET 5.0


Ruberik:
Code:
```
        static void Main(string[] args) {
            using var AcceleratorContext = new Context(ILGPU.IR.Transformations.OptimizationLevel.O2);
            AcceleratorContext.EnableAlgorithms();
            using var Accelerator = new CudaAccelerator(AcceleratorContext) as Accelerator;
            var kernel = Accelerator.LoadAutoGroupedStreamKernel<Index1, ArrayView<double>>(MyKernel);
        }

        public static void MyKernel(Index1 index, ArrayView<double> output) {
            output[index] = Math.Round(1.0);
        }
```

Exception:
```
Unhandled exception. ILGPU.InternalCompilerException: 'An internal compiler error has been detected' in method Double Int64BitsToDouble(Int64) declared in type System.BitConverter
 ---> System.NotSupportedException: Not supported runtime method 'get_IsSupported'
   --- End of inner exception stack trace ---
   at ILGPU.Backends.Backend.PreCompileKernelMethod[TBackendHook](EntryPointDescription& entry, TBackendHook backendHook)
   at ILGPU.Backends.Backend.Compile[TBackendHook](EntryPointDescription& entry, KernelSpecialization& specialization, TBackendHook backendHook)
   at ILGPU.Backends.Backend.Compile(EntryPointDescription& entry, KernelSpecialization& specialization)
   at ILGPU.Runtime.Accelerator.CompileKernel(EntryPointDescription& entry, KernelSpecialization& specialization)
   at ILGPU.Runtime.Accelerator.<>c__DisplayClass173_0`1.<LoadGenericKernel>b__0(TKernelLoader& loader, KernelInfo& info)
   at ILGPU.Runtime.Accelerator.LoadAutoGroupedKernel[TDelegate](MethodInfo method, KernelInfo& kernelInfo)
   at ILGPU.Runtime.Accelerator.LoadAutoGroupedKernel[TDelegate](MethodInfo method)
   at ILGPU.Runtime.KernelLoaders.LoadAutoGroupedKernel[TIndex,T1](Accelerator accelerator, Action`2 action)
   at ILGPU.Runtime.KernelLoaders.LoadAutoGroupedStreamKernel[TIndex,T1](Accelerator accelerator, Action`2 action)
   at FurrowUtilities.Misc.Main(String[] args) in C:\main\FurrowUtilities\Misc\Misc.cs:line 69
[...]

PR #436, opened 2021-03-16
Record call stack during compilation for error reporting.

MoFtZ:
Related to #425.

`Math.Round` does not have a Cuda intrinsic, and therefore does not currrently have a special handler in ILGPU.Algorithms. Instead, the code is dissassembled like any other normal function.

In net5.0, `Math.Round` internally calls a method in the `System.Runtime` namespace, which is not supported by ILGPU. This throws a vague exception about the method `get_IsSupported` not being supported in `Int64BitsToDouble`. However, it is not obvious where that is being called from.

This PR tracks the compilation stack to provide more detailed information about what is causing the compilation to fail. An example exception would look like this:
```
ILGPU.InternalCompilerException
  Message=
An internal compiler error has been detected
- in method Boolean get_IsSupported() declared in type System.Runtime.Intrinsics.X86.Sse2+X64
- in method Double Int64BitsToDouble(Int64) declared in type System.BitConverter
- in method Double Round(Double) declared in type System.Math
- in file 'Program.cs', line 11 column 13 to 57
- in method Void MyKernel(ILGPU.Index1, ILGPU.ArrayView`1[System.Double]) declared in type Program

Inner Exception 1:
NotSupportedException: Not supported runtime method 'get_IsSupported'
```

### DIFF
--- a/Src/ILGPU/IR/Location.cs
+++ b/Src/ILGPU/IR/Location.cs
@@ -11,9 +11,11 @@
 
 using ILGPU.Resources;
 using System;
+using System.Collections.Immutable;
 using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
 using System.Runtime.CompilerServices;
+using System.Text;
 
 namespace ILGPU.IR
 {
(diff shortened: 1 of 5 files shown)


Src/ILGPU/IR/Location.cs
```diff
@@ -246,4 +265,101 @@ public class FileLocation : Location
...
     }
+
+    /// <summary>
+    /// A collection of locations than is manipulated as a stack.
+    /// </summary>
+    public class CompilationStackLocation : Location
+    {
+        #region Instance
+
+        private ImmutableStack<Location> Stack { get; }
```
m4rs-mt, line 276: Nit: move properties (even private ones) to a `Properties` region.

m4rs-mt (approved):
Really nice patch 👍 This helps us to output much better error messages when compiling.

Src/ILGPU/IR/Location.cs
MoFtZ: Fixed 👍

m4rs-mt:
@MoFtZ looks like the latest commit causes a warning related to a protected member 😢

MoFtZ:
Yes, I saw the build failure - wanted to fix it, but had some distractions 😄

PR #438, opened 2021-03-17
Implemented CopySign intrinsic.

MoFtZ:
Related to https://github.com/m4rs-mt/ILGPU/issues/425.

`Math.Round` does not have an intrinsic implementation. After adding the `BitConverter` intrinsics from https://github.com/m4rs-mt/ILGPU/pull/437, there is still an additional failure because the `net5.0` implementation of `Math.CopySign` internally also makes a call to the `System.Runtime` namespace.

This PR helps address the issue by providing an intrinsic implementation for `Math.CopySign`.

When the `BitConverter` PR and the `CopySign` PR are combined, they allow `Math.Round` to be compiled by ILGPU like a normal function.

### DIFF
--- a/Src/ILGPU/IntrinsicMath.cs
+++ b/Src/ILGPU/IntrinsicMath.cs
@@ -654,5 +654,43 @@ namespace ILGPU
         public static long ComposeLong(IntegerParts parts) => (long)ComposeULong(parts);
 
         #endregion
+
+        #region CopySign
+
+        /// <summary>
+        /// Returns a value with the magnitude of x and the sign of y.
+        /// </summary>
+        /// <param name="x">A number whose magnitude is used in the result.</param>
+        /// <param name="y">A number whose sign is the used in the result.</param>
+        /// <returns>A value with the magnitude of x and the sign of y.</returns>
+        [MathIntrinsic(MathIntrinsicKind.CopySignF)]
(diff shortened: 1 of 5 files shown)


MoFtZ:
Fixed CI failure on `net47` on Windows runner.

The issue is that `Math.CopySign(double, double)` is not available on `net47`, so ILGPU would crash at startup.

Src/ILGPU/IntrinsicMath.cs
```diff
@@ -654,5 +654,43 @@ public IntegerParts(uint lower, uint upper)
...
+        /// <param name="y">A number whose sign is the used in the result.</param>
+        /// <returns>A value with the magnitude of x and the sign of y.</returns>
+        [MathIntrinsic(MathIntrinsicKind.CopySignF)]
+        public static float CopySign(float x, float y) =>
+#if !NETFRAMEWORK && !NETSTANDARD
+            Math.CopySign(x, y)
+#else
+            // NB: net47 and netstandard2.1 do not support Math.CopySign.
+            Interop.IntAsFloat(
+                (Interop.FloatAsInt(x) & ~(1U << 31)) |
```
m4rs-mt, line 690: The good old bit magic 🥇

m4rs-mt (approved):

Issue #441, opened 2021-03-18
Compiling kernel with constant switch condition generates incorrect OpenCL code.


MoFtZ:
When compiling a kernel with a `switch` statement that has a constant condition, the OpenCL code generator produces incorrect code.

The following is an example kernel:
```CSharp
public static void ConstantSwitchKernel(Index1 index, ArrayView<double> output)
{
    var mode = 3;
    output[index] = mode switch
    {
        1 => 1.0,
        2 => 2.0,
        3 => 3.0,
        _ => 0.0
    };
}
```
The OpenCL code generator ends up producing code similar to:
```CPP
switch (var19)
{
case 0:
    goto L_1065;
case 1:
    goto L_1064;
case 2:
    goto L_1063;
default:
    goto L_1066;
}
```
with a compiler error of:
```
error: use of undeclared identifier 'var19'
        switch (var19)
                ^
```

 In particular, it only affects `OptimizationLevel.Debug`. This is because the other optimisation levels include the `UnreachableCodeElimination` transformation which will detect the constant condition and fold/remove the `switch` statement.

NOTE: The `ConstantSwitchKernel` kernel only fails when there are at least 4 conditions. If there are less conditions, the C# compiler converts the `switch` statement into if/else statements.

m4rs-mt:
@MoFtZ Looks like the constant of the switch condition is not output properly...

MoFtZ:
@m4rs-mt Yes, we already handle this for constant `if` statements, but not for constant `switch` statements. Working on a fix now.

PR #442, opened 2021-03-19
Fixed emitting switch statement with constant condition.

MoFtZ:
Fixes https://github.com/m4rs-mt/ILGPU/issues/441.

Refactored parsing of string format from WriteToOutput into a utility class so that it can be shared.
Also refactored that parsing code to use spans, and also support escapeing brackets. 

Added `FormattableString` to `StatementEmitter`, with support for `Variable`, `BasicBlock` and all basic types.
This was then used to generate a `switch` statement that could handle a constant condition.

NOTE: I have not tried to emit an optimised `switch` statement with a single `goto` branch, like the `if` condition. Should I add this optimisation? or should we require that the ILGPU compilation transforms perform the folding?

### DIFF
--- a/Src/ILGPU/Backends/OpenCL/CLCodeGenerator.Emitter.cs
+++ b/Src/ILGPU/Backends/OpenCL/CLCodeGenerator.Emitter.cs
@@ -12,6 +12,8 @@
 using ILGPU.IR;
 using ILGPU.IR.Types;
 using ILGPU.IR.Values;
+using ILGPU.Resources;
+using ILGPU.Util;
 using System;
 using System.Diagnostics;
 using System.Globalization;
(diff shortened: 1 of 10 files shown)


Src/ILGPU/Backends/OpenCL/CLCodeGenerator.Emitter.cs
```diff
@@ -435,8 +437,114 @@ public void AppendArgumentAddress(Variable argument)
...
+                                AppendConstant((ulong)argument);
+                                break;
+                            case TypeCode.Single:
+                                AppendConstant((float)argument);
+                                break;
+                            case TypeCode.Double:
+                                AppendConstant((double)argument);
+                                break;
+                            case TypeCode.String:
+                                AppendOperation((string)argument);
```
MoFtZ, line 522: TODO: Faster to just call `stringBuilder.Append`.

Src/ILGPU/Backends/OpenCL/CLCodeGenerator.Emitter.cs
```diff
@@ -435,8 +437,114 @@ public void AppendArgumentAddress(Variable argument)
...
+                                break;
+                            default:
+                                if (argument is Variable variable)
+                                {
+                                    Append(variable);
+                                    break;
+                                }
+                                else if (argument is BasicBlock block)
+                                {
+                                    AppendOperation(CodeGenerator.blockLookup[block]);
```
MoFtZ, line 532: TODO: Faster to just call stringBuilder.Append.

m4rs-mt (approved):
Nice refinement of the formatting functions 👍.

PR #452, opened 2021-03-28
Fixed invalid detection of loop bodies.

m4rs-mt:
A community member reported an issue that caused an `InteralCompilerError` when compiling a complex kernel consisting of divergent control flow. It turns out that the current `LoopAnalysis` implementation has a problem when dealing with loops consisting of a single block in combination with input blocks jumping to different loops at the same time.

This PR fixes the invalid detection of loop bodies in more complex control-flow scenarios.

### DIFF
--- a/Src/ILGPU.Tests/BasicLoops.cs
+++ b/Src/ILGPU.Tests/BasicLoops.cs
@@ -488,6 +488,60 @@ namespace ILGPU.Tests
             var expected = Enumerable.Repeat(0, Length).ToArray();
             Verify(buffer, expected);
         }
+
+        internal static void LoopUnrolling_LICM_Kernel(
+            Index1 index,
+            ArrayView<int> source,
+            ArrayView<int> target)
+        {
+            int result = LoopUnrolling_LICM_Kernel_1(source);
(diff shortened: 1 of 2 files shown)


MoFtZ (changes requested):
ISSUE: Something is not right.. when I remove the changes to the Loop analyzer, and run just the new unit test, all tests pass (on CPU/Cuda/OpenCL on net47/netcoreapp3.1).

Could you please double check the unit test?

m4rs-mt:
@MoFtZ I am going to investigate this (potential) issue 👀

m4rs-mt:
I found out what the "problem" is: when we run all tests with the `Debug` build, the code optimizations of the C# compiler are disabled. However, this problem only occurs when we compile all tests in `Release` mode, which is what currently happens in our CI build.

@jgiannuzzi and I have already discussed testing `Debug` and `Release` builds in the CI. Maybe this would be a good time to add such a feature 🤓 ?

MoFtZ:
OK, managed to reproduce the issue by running the unit test in Release configuration using the O1 and O2 optimisation levels.

PR #454, opened 2021-04-07
Restrict list of Cuda devices to only return supported devices.

MoFtZ:
When querying the list of Cuda accelerators, we used to return all devices, even if they were too old to support ILGPU. Added a simple check for whether the Cuda device is compatible with the supported instruction sets.

Also added the new ISA 7.2 instruction set.

~~Also changed the instruction set detection to deal with future/unknown instruction sets - in the same way as the PTX architecture detection.~~


m4rs-mt (approved):
Nice patch 👍 This avoids confusion when using ILGPU in the presence of unsupported devices. However, users can still use `.Cuda(predicate)` to enable unsupported devices.

Issue #341, opened 2020-12-03
CudaException: invalid argument in LoadImplicitlyGroupedStreamKernel()()


Ruberik:
Hi there! First, thank you for this library. It's extremely useful.

I have an issue with a simple reproduction. The short summary is that kernels loaded using LoadImplicitlyGroupedStreamKernel have some surprising limits on the parameters they can be called with. If the parameter:

- Is either an Index2 or an Index3, and
   - Has an entry other than X that's at least 2^16, we get ILGPU.Runtime.Cuda.CudaException: invalid argument.
   - Otherwise, if the product of its entries is at least 2^38, we get System.ArgumentOutOfRangeException: Specified argument was out of the range of valid values. (Parameter 'gridDim')

Test code:
```
using var context = new Context(ILGPU.IR.Transformations.OptimizationLevel.Release);
using var gpu = new CudaAccelerator(context);
void TestKernel1D(Index1 index, int a) { };
void TestKernel2D(Index2 index, int a) { };
void TestKernel3D(Index3 index, int a) { };
var TK1 = gpu.LoadImplicitlyGroupedStreamKernel<Index1, int>(TestKernel1D, 64 * 2);
var TK2 = gpu.LoadImplicitlyGroupedStreamKernel<Index2, int>(TestKernel2D, 64 * 2);
var TK3 = gpu.LoadImplicitlyGroupedStreamKernel<Index3, int>(TestKernel3D, 64 * 2);
TK1(new Index1(1 << 30), 5);           // OK
TK2(new Index2(1 << 30, 1 << 7), 5);   // OK
TK2(new Index2(1, 1 << 16), 5);        // Second index over 2^16: CudaException
TK2(new Index2(1 << 30, 1 << 8), 5);   // Index product over 2^38: ArgumentOutOfRangeException
TK2(new Index2(1 << 30, 1 << 16), 5);  // Both problems: CudaException
TK3(new Index3(1 << 30, 1, 1), 5);     // OK
TK3(new Index3(1, 1 << 16, 1), 5);     // Second index over 2^16: CudaException
TK3(new Index3(1, 1, 1 << 16), 5);     // Third index over 2^16: CudaException
TK3(new Index3(1 << 30, 1 << 3, 1 << 3), 5);  // OK
TK3(new Index3(1 << 30, 1 << 4, 1 << 4), 5);  // Index product over 2^38: ArgumentOutOfRangeException
```

[...]

PR #457, opened 2021-04-09
Added bounds validation to kernel launch delegates.

MoFtZ:
Fixes https://github.com/m4rs-mt/ILGPU/issues/341.

Show a consistent error message when the kernel launch bounds are out-of-range. For implicitly grouped kernels, we verify the bounds before creating the KernelConfig so that the internal exceptions from that class are not triggered, as they are less descriptive.

Reduced max grid size of CPU accelerator to match Cuda accelerators. Also makes it possible to actually test the CPU grid bounds.

Added skipping of unit tests that will always fail such as when the testing size exceeds the max group size.

### DIFF
--- a/Src/ILGPU.Tests/ILGPU.Tests.csproj
+++ b/Src/ILGPU.Tests/ILGPU.Tests.csproj
@@ -36,7 +36,6 @@
 
   <ItemGroup>
     <PackageReference Include="GitHubActionsTestLogger" Version="1.1.0" />
-    <PackageReference Include="FluentAssertions" Version="5.10.3" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.7.1" />
     <PackageReference Include="xunit" Version="2.4.1" />
     <DotNetCliToolReference Include="dotnet-xunit" Version="2.3.1" />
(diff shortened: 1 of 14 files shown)


Src/ILGPU.Tests/ILGPU.Tests.csproj
```diff
@@ -36,11 +36,11 @@
 
   <ItemGroup>
     <PackageReference Include="GitHubActionsTestLogger" Version="1.1.0" />
-    <PackageReference Include="FluentAssertions" Version="5.10.3" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.7.1" />
     <PackageReference Include="xunit" Version="2.4.1" />
     <DotNetCliToolReference Include="dotnet-xunit" Version="2.3.1" />
     <PackageReference Include="T4.Build" Version="0.2.0" PrivateAssets="All" />
+    <PackageReference Include="Xunit.SkippableFact" Version="1.4.13" />
```
m4rs-mt, line 43: Nice feature 🤓 we can also use this feature set to skip several OpenCL tests on different platforms....

m4rs-mt (approved):
Very nice patch. This one resolves a number of misleading exceptions and unifies all kinds of exceptions related to grid and group indices 👍

Issue #382, opened 2021-01-12
Different exception thrown in different graphic cards


AnakinCN:
Hi

I tried writing a sample using structures with ILGPU, and it throws different in  different graphic cards (OpenCL & CUDA), none succeeded.

I believe is something with my code, but how to debug this? There's no detailed information on the exception.

[BlackILGPUnoClass.zip](https://github.com/m4rs-mt/ILGPU/files/5801038/BlackILGPUnoClass.zip)

Thank you in advance.

MoFtZ:
@AnakinCN
The issue is in this bit of code:
```CSharp
public struct EnvelopeItemsStruct
{
    public EnvelopeItem[] Items;
     public int Cursor;
}
```

The member field `Items` is using an array, which is a reference type, and is not supported by ILGPU.

As an alternative, you might be able to use [fixed buffers](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/unsafe-code-pointers/fixed-size-buffers).

And you are correct, the error messages are very generic - we should try to improve this bit of the code.
I get something like this:
```
System.ArgumentOutOfRangeException
  Message=Specified argument was out of the range of valid values. (Parameter 'type')
  StackTrace:
   at ILGPU.Backends.EntryPoints.ArgumentMapper.MapType(Type type)
   at ILGPU.Backends.EntryPoints.ArgumentMapper.MapStructType(Type structType)
   at ILGPU.Backends.EntryPoints.ArgumentMapper.MapType(Type type)
   at ILGPU.Backends.EntryPoints.ArgumentMapper.MapStructType(Type structType)
   at ILGPU.Backends.EntryPoints.ArgumentMapper.MapType(Type type)
   at ILGPU.Backends.EntryPoints.ArgumentMapper.Map[TILEmitter,TMappingHandler](TILEmitter& emitter, TMappingHandler& mappingHandler, ParameterCollection& parameters)
   at ILGPU.Backends.PTX.PTXArgumentMapper.Map[TILEmitter](TILEmitter& emitter, EntryPoint entryPoint)
```

MoFtZ:
> I tried writing a sample using structures with ILGPU, and it throws different in different graphic cards (OpenCL & CUDA), none succeeded.

When I test on OpenCL and Cuda, they both throw the same exception. What different exceptions are you getting?

AnakinCN:
[...]

PR #460, opened 2021-04-13
Improved reporting on invalid kernel arguments.

MoFtZ:
Related to #382.

When a kernel argument uses an unsupported type (e.g. a struct with an array member), the exception thrown is similar to:
```
System.ArgumentOutOfRangeException: Specified argument was out of the range of valid values. (Parameter 'type')
```
This exception is throw for the Cuda and OpenCL accelerators, but not for the CPU accelerator.

This PR adds argument mapping to the CPU accelerator to mimic the behaviour of the other accelerators.
Further, it improves on the exception thrown, to produce something like:
```
System.ArgumentException: 'Not supported kernel-parameter type 'Test.Program+UnsupportedKernelParam''
Inner Exception
NotSupportedException: Class type 'System.Int32[]' is not supported
```

### DIFF
--- a/Src/ILGPU.Tests/KernelEntryPoints.cs
+++ b/Src/ILGPU.Tests/KernelEntryPoints.cs
@@ -555,5 +555,27 @@ namespace ILGPU.Tests
             var expected = Enumerable.Range(0, Length).ToArray();
             Verify(buffer, expected);
         }
+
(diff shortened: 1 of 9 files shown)


m4rs-mt (approved):
Awesome patch 🚀

Issue #416, opened 2021-02-13
Documentation request: add intrinsic example


mikhail-khalizev:
I have a need to multiply two UInt64 numbers and get UInt128. Potentially, CUDA has instructions "mul.lo.u64" and "mul.hi.u64" for this.

Could you supplement the documentation http://www.ilgpu.net/Documentation with an example of how to write intrinsic yourself (for example for "mul.hi.u64")?

MoFtZ:
@mikhail-khalizev Have you seen the [custom intrinsic example](https://github.com/m4rs-mt/ILGPU.Samples/tree/master/Src/CustomIntrinsics) in [ILGPU.Samples](https://github.com/m4rs-mt/ILGPU.Samples)?

mikhail-khalizev:
No, I didn't notice this example, unfortunately.

Now I looked. But there is still the question of how to use `mul.hi.u64` directly, since now the example uses the C # expression:

```csharp
        /// <summary>
        /// The Cuda (PTX) implementation.
        /// </summary>
        public static int ComputePTX(int x) => 2 * x;
```

MoFtZ:
There are two types of intrinsics in that sample code. The first is remappped intrinsics, which points the code to a different method.

The second is code generated intrinsics (`CustomCodeGeneratorIntrinsic`). In the method `GeneratePTXCode`, you create your command, something like this:
```CSharp
using (var command = codeGenerator.BeginCommand("mul.hi.u64"))
{
   command.AppendArgument(.....);
   command.AppendArgument(.....);
}
```

mikhail-khalizev:
Thanks!

PR #467, opened 2021-05-02
Added support for inline PTX assembly instructions.

MoFtZ:
Related to https://github.com/m4rs-mt/ILGPU/issues/416.

Provides a new `CudaAsm` class that can be used within a kernel to embed PTX instructions. This makes it easier to use advanced PTX instructions without the need to create a custom ILGPU intrinsic.

It supports multiple input arguments (arbitrarily capped at 10 in the T4 template), and a single output value per instruction statement.

```CSharp
public static void MyKernel(Index1D index, ArrayView<double> view)
{
    CudaAsm.Emit(
        "{\n\t" +
        "   .reg .f64 t1;\n\t" +
        "   add.f64 t1, %1, %2;\n\t" +
        "   add.f64 %0, t1, %2;\n\t" +
        "}",
        out double result,
        (double)index.X,
        42.0);
    view[index] = result;
}
```
The format string uses `%0` rather than the .NET way of `{0}`. This was to make it match the [inline PTX assembly](https://docs.nvidia.com/cuda/inline-ptx-assembly/index.html) in C++. If you want to use the `%`, then you should escape it with double `%%`.

UPDATE: The selective detection of PTX version, described below has been removed as it was causing issues. May revisit in a separate PR in the future.

It is also possible to selectively detect the PTX version available:
```CSharp
if (CudaAsm.IsPTXVersionAtLeast(CudaInstructionSet.ISA_70)) { }
else if (CudaAsm.IsPTXVersionAtLeast(CudaInstructionSet.ISA_60)) { }
else { }
```

KNOWN ISSUES:
- Using `CudaAsm.IsPTXVersionAtLeast` produces a complex constant expression that the ILGPU compiler is not currently able to fold into an unconditional branch. This means that the PTX compiler may fail if it does not support a newer instruction, that should have been optimised out.  Further work required to improve this area.

### DIFF
--- a/Src/ILGPU/Backends/OpenCL/CLCodeGenerator.Values.cs
+++ b/Src/ILGPU/Backends/OpenCL/CLCodeGenerator.Values.cs
@@ -706,5 +706,10 @@ namespace ILGPU.Backends.OpenCL
         public void GenerateCode(DebugAssertOperation debug) =>
             // Invalid debug node -> should have been removed
             debug.Assert(false);
+
+        /// <summary cref="IBackendCodeGenerator.GenerateCode(LanguageEmitValue)"/>
+        public void GenerateCode(LanguageEmitValue value) =>
+            // Ignore PTX instructions.
+            value.Assert(value.LanguageKind == LanguageKind.PTX);
     }
 }
--- a/Src/ILGPU/Frontend/Intrinsic/LanguageIntrinsics.cs
+++ b/Src/ILGPU/Frontend/Intrinsic/LanguageIntrinsics.cs
@@ -15,6 +15,7 @@ using ILGPU.Resources;
 using ILGPU.Util;
 using System;
 using System.Collections.Immutable;
+using System.Text.RegularExpressions;
 using static ILGPU.Util.FormatString;
 using FormatArray = System.Collections.Immutable.ImmutableArray<
     ILGPU.Util.FormatString.FormatExpression>;
@@ -47,6 +48,13 @@ namespace ILGPU.Frontend.Intrinsic
 
     partial class Intrinsics
     {
+        /// <summary>
+        /// Regex for parsing PTX assembly instructions.
+        /// </summary>
+        private static readonly Regex PTXExpressionRegex =
+            // Escape sequence, %n arguments, singular % detection.
+            new Regex("(%%|%\\d+|%)");
+
         /// <summary>
         /// Handles language operations.
         /// </summary>
@@ -165,59 +173,33 @@ namespace ILGPU.Frontend.Intrinsic
             string ptxExpression,
             out FormatArray expressions)
         {
-            expressions = FormatArray.Empty;
-
             // Search for '%n' format arguments
-            var result = ImmutableArray.CreateBuilder<FormatExpression>(10);
-            var expression = ptxExpression.AsSpan();
+            var parts = PTXExpressionRegex.Split(ptxExpression);
+            var result = ImmutableArray.CreateBuilder<FormatExpression>(parts.Length);
 
-            while (expression.Length > 0)
+            foreach (var part in parts)
             {
-                // Search for next %
-                int foundIndex = expression.IndexOf('%');
-                if (foundIndex < 0)
+                if (part == "%%")
                 {
-                    result.Add(new FormatExpression(expression));
-                    break;
+                    result.Add(new FormatExpression("%"));
                 }
-
-                var escaped =
-                    foundIndex + 1 < expression.Length &&
-                    expression[foundIndex] == expression[foundIndex + 1];
-                if (escaped)
+                else if (part.StartsWith("%"))
                 {
-                    result.Add(new FormatExpression(expression.Slice(0, foundIndex + 1)));
-                    expression = expression.Slice(foundIndex + 2);
+                    // Check whether the argument can be resolved to an integer.
+                    if (int.TryParse(part.Substring(1), out int argument))
+                    {
+                        result.Add(new FormatExpression(argument));
+                    }
+                    else
+                    {
+                        // Singular % or remaining text was not a number.
+                        expressions = FormatArray.Empty;
+                        return false;
+                    }
                 }
-                else
+                else if (part.Length > 0)
                 {
-                    // Append text before new argument
-                    if (foundIndex > 0)
-                    {
-                        result.Add(new FormatExpression(
-                            expression.Slice(0, foundIndex)));
-                    }
-                    expression = expression.Slice(foundIndex + 1);
-
-                    // Retrieve all the numeric text
-                    if (expression.Length == 0)
-                        return false;
-
-                    int endIndex = 0;
-                    while (endIndex < expression.Length &&
-                        char.IsDigit(expression[endIndex]))
-                    {
-                        endIndex++;
-                    }
-
-                    // Check whether the argument can be resolved to an integer
-                    var numericExpr = expression.Slice(0, endIndex);
-                    if (!int.TryParse(numericExpr.ToString(), out int argument))
-                        return false;
-
-                    // Append current argument
-                    result.Add(new FormatExpression(argument));
-                    expression = expression.Slice(endIndex);
+                    result.Add(new FormatExpression(part));
                 }
             }
 
--- a/Src/ILGPU/Frontend/Intrinsic/LanguageIntrinsics.cs
+++ b/Src/ILGPU/Frontend/Intrinsic/LanguageIntrinsics.cs
@@ -109,15 +109,8 @@ namespace ILGPU.Frontend.Intrinsic
             // Valid all argument types
             foreach (var arg in arguments)
             {
-                if (arg.BasicValueType == BasicValueType.Int16 ||
-                    arg.BasicValueType == BasicValueType.Int32 ||
-                    arg.BasicValueType == BasicValueType.Int64 ||
-                    arg.BasicValueType == BasicValueType.Float16 ||
-                    arg.BasicValueType == BasicValueType.Float32 ||
-                    arg.BasicValueType == BasicValueType.Float64)
-                {
+                if (arg.BasicValueType != BasicValueType.None)
                     continue;
-                }
                 throw location.GetNotSupportedException(
                     ErrorMessages.NotSupportedInlinePTXFormatArgumentType,
                     ptxExpression,
--- a/Src/ILGPU/Frontend/Intrinsic/LanguageIntrinsics.cs
+++ b/Src/ILGPU/Frontend/Intrinsic/LanguageIntrinsics.cs
@@ -1,0 +1,235 @@
+﻿// ---------------------------------------------------------------------------------------
+//                                        ILGPU
+//                        Copyright (c) 2016-2020 Marcel Koester
+//                                    www.ilgpu.net
+//
+// File: LanguageIntrinsics.cs
+//
+// This file is part of ILGPU and is distributed under the University of Illinois Open
+// Source License. See LICENSE.txt for details
+// ---------------------------------------------------------------------------------------
+
+using ILGPU.IR;
+using ILGPU.IR.Values;
+using ILGPU.Resources;
+using ILGPU.Util;
+using System;
+using System.Collections.Immutable;
+using static ILGPU.Util.FormatString;
+using FormatArray = System.Collections.Immutable.ImmutableArray<
+    ILGPU.Util.FormatString.FormatExpression>;
+
+namespace ILGPU.Frontend.Intrinsic
+{
+    enum LanguageIntrinsicKind
+    {
+        EmitPTX,
+    }
+
+    /// <summary>
+    /// Marks inline language methods that are built in.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Method, AllowMultiple = false)]
+    sealed class LanguageIntrinsicAttribute : IntrinsicAttribute
+    {
+        public LanguageIntrinsicAttribute(LanguageIntrinsicKind intrinsicKind)
+        {
+            IntrinsicKind = intrinsicKind;
+        }
+
+        public override IntrinsicType Type => IntrinsicType.Language;
+
+        /// <summary>
+        /// Returns the assigned intrinsic kind.
+        /// </summary>
+        public LanguageIntrinsicKind IntrinsicKind { get; }
+    }
+
+    partial class Intrinsics
+    {
+        /// <summary>
+        /// Handles language operations.
+        /// </summary>
+        /// <param name="context">The current invocation context.</param>
+        /// <param name="attribute">The intrinsic attribute.</param>
+        /// <returns>The resulting value.</returns>
+        private static ValueReference HandleLanguageOperation(
+            ref InvocationContext context,
+            LanguageIntrinsicAttribute attribute)
+        {
+            var builder = context.Builder;
+            return attribute.IntrinsicKind switch
+            {
+                LanguageIntrinsicKind.EmitPTX =>
+                    CreateLanguageEmitPTX(ref context),
+                _ => throw context.Location.GetNotSupportedException(
+                    ErrorMessages.NotSupportedLanguageIntrinsic,
+                    attribute.IntrinsicKind.ToString()),
+            };
+        }
+
+        /// <summary>
+        /// Creates a new inline PTX instruction.
+        /// </summary>
+        /// <param name="ptxExpression">The PTX expression string.</param>
+        /// <param name="context">The current invocation context.</param>
+        private static ValueReference CreateLanguageEmitPTX(
+            string ptxExpression,
+            ref InvocationContext context)
+        {
+            // Parse PTX expression and ensure valid argument references
+            var location = context.Location;
+            if (!TryParse(ptxExpression, out var expressions))
+            {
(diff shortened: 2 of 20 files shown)


Src/ILGPU/Runtime/Cuda/CudaAsm.cs
```diff
@@ -0,0 +1,60 @@
...
+
+        /// <summary>
+        /// Determines if the PTX instruction set is at least the specified version.
+        /// </summary>
+        /// <param name="major">The PTX major version.</param>
+        /// <param name="minor">The PTX minor version.</param>
+        /// <returns>Return true if the PTX version is supported.</returns>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static bool IsPTXVersionAtLeast(int major, int minor) =>
+            IsSupported &&
```
m4rs-mt, line 49: I think if we change this `&&` implementation to a bitwise `$`, ILGPU should be able to fold this expression in all cases.

Src/ILGPU/Backends/OpenCL/CLCodeGenerator.Values.cs
```diff
@@ -706,5 +706,10 @@ public void GenerateCode(WarpShuffle shuffle)
         public void GenerateCode(DebugAssertOperation debug) =>
             // Invalid debug node -> should have been removed
             debug.Assert(false);
+
+        /// <summary cref="IBackendCodeGenerator.GenerateCode(LanguageEmitValue)"/>
+        public void GenerateCode(LanguageEmitValue value) =>
+            // Ignore PTX instructions.
+            value.Assert(value.LanguageKind == LanguageKind.PTX);
```
m4rs-mt, line 713: I am not sure whether this assert is correct... 🤓

MoFtZ: Correct. In future, if we add a new `LanguageKind`, this code will need to be modified to either ignore the new language, or emit the correct instructions.

Src/ILGPU/Runtime/Cuda/CudaAsm.cs
MoFtZ: Unfortunately, using bitwise `&` does not help.

When `Accelerator.CurrentType` and `Accelerator.LanguageVersion` are specialised, they replace the placeholder IR node, but no folding it performed.

Using `IsSupported` as an example, the IR node will be `CompareValue(AcceleratorTypeValue, PrimitiveValue)`. When `Accelerator.CurrentType` is specialised, the `AcceleratorTypeValue` node is replaced by a `PrimitiveValue` - making the condition `CompareValue(PrimitiveValue, PrimitiveValue)` Later, when `UnreachableCodeElimination` is run, it finds that the condition is `CompareValue`, which cannot be folded - it only knows how to fold `PrimitiveValue`.

m4rs-mt: You are right 😢 The aggressive constant-propagation pass in 57f3232228f2c4cea282570a300ffd92002fa412 should help to aggressively fold all expressions known at compile time. Perhaps right now is the right time to integrate this pass into ILGPU.

MoFtZ: We originally discussed using the constant-propagation pass in the `O2` optimisation because it was expensive. However, after completing the work for `CudaAsm`, I realised that we need a solution for `O0` too. This is because in `O0` the PTX compiler will see all the PTX instructions, including future instructions that it does not know about.

Src/ILGPU/IR/Values/LanguageValues.cs
```diff
@@ -0,0 +1,216 @@
...
+        protected override string ToPrefixString() => "langVersion";
+
+        #endregion
+    }
+
+    /// <summary>
+    /// Represents an inline lanaguage statement.
+    /// </summary>
+    [ValueKind(ValueKind.LanguageEmit)]
+    public sealed class LanguageEmitValue : Value
```
m4rs-mt, line 87: I guess we should make this class inherit from `MemoryValue` to avoid moving this value around due to possible side effects.

Src/ILGPU/IR/BasicBlock.cs
```diff
@@ -453,8 +453,12 @@ public bool HasSideEffects()
         {
             foreach (Value value in this)
             {
-                if (value is MemoryValue || value is MethodCall)
+                if (value is MemoryValue ||
+                    value is MethodCall ||
+                    value is LanguageEmitValue)
```
m4rs-mt, line 458: If we change the base type to `MemoryValue` we can safely remove this again 🤓

Src/ILGPU/IR/Values/LanguageValues.cs
MoFtZ: Done.

Src/ILGPU/IR/BasicBlock.cs
MoFtZ: Done.

Src/ILGPU/Frontend/Intrinsic/LanguageIntrinsics.cs
```diff
@@ -0,0 +1,238 @@
...
+
+            // Gather all arguments
+            var arguments = InlineList<ValueReference>.Empty;
+            context.Arguments.CopyTo(ref arguments);
+            arguments.RemoveAt(0);
+
+            // Valid all argument types
+            foreach (var arg in arguments)
+            {
+                if (arg.BasicValueType == BasicValueType.Int16 ||
```
m4rs-mt, line 115: May we add a comment here why `Int8` is not supported, since it cannot be represented properly in a non-existing 8-byte register?

Src/ILGPU/Frontend/Intrinsic/LanguageIntrinsics.cs
```diff
@@ -0,0 +1,238 @@
...
+            expressions = FormatArray.Empty;
+
+            // Search for '%n' format arguments
+            var result = ImmutableArray.CreateBuilder<FormatExpression>(10);
+            var expression = ptxExpression.AsSpan();
+
+            while (expression.Length > 0)
+            {
+                // Search for next %
+                int foundIndex = expression.IndexOf('%');
```
m4rs-mt, line 187: Question: May we use a `Regex` instead?

Src/ILGPU/IR/Transformations/AcceleratorSpecializer.cs
```diff
@@ -58,6 +58,12 @@ public class AcceleratorSpecializer : SequentialUnorderedTransformation
             public readonly AcceleratorType AcceleratorType =>
                 Specializer.AcceleratorType;
 
+            /// <summary>
+            /// Returns the inline language version number for the  current accelerator.
+            /// </summary>
+            public readonly int AcceleratorLanguageVersion =>
```
m4rs-mt, line 64: Question: May we introduce an ILGPU-specific accelerator language-version enum?

Src/ILGPU/IR/Values/LanguageValues.cs
```diff
@@ -0,0 +1,216 @@
...
+        internal LanguageEmitValue(
+            in ValueInitializer initializer,
+            LanguageKind languageKind,
+            FormatArray expressions,
+            bool hasOutput,
+            ref ValueList arguments,
+            VoidType voidType)
+            : base(initializer, voidType)
+        {
+            foreach (var argument in arguments)
```
m4rs-mt, line 109: Question: May we put this into `#if DEBUG` to avoid the emission of empty for-loop in release mode?

Src/ILGPU/Runtime/Accelerator.cs
```diff
@@ -87,6 +87,30 @@ public static AcceleratorType CurrentType
...
+        /// <summary>
+        /// Converts the major/minor version into a single language version number.
+        /// </summary>
+        /// <param name="major">The major version.</param>
+        /// <param name="minor">The minor version.</param>
+        /// <returns>The language version number</returns>
+        /// <remarks>
+        /// The format is arbitrary, and assumes there are at most 99 minor versions.
+        /// </remarks>
+        internal static int MakeLanguageVersion(int major, int minor) =>
```
m4rs-mt, line 111: A got it... do you think an "arbitrary" version structure for ILGPU-specific inline language elements make sense (see comment about the `enumeration` above).

Src/ILGPU/Runtime/Cuda/CudaAsm.cs
m4rs-mt: Summary of our discussion from the talk-to-dev session: we came up with the idea of limiting the scope of the (expensive) constant propagation pass to certain basic blocks. This helps to improve performance while being sufficient to optimize the expressions mentioned above.

Src/ILGPU/Frontend/Intrinsic/LanguageIntrinsics.cs
MoFtZ: This restriction is because the [inline PTX assembly](https://docs.nvidia.com/cuda/inline-ptx-assembly/index.html) did not have a register constraint for smaller types. However, on second thought, I think we should allow all primitive types and just deal with it in the PTX emitter.

Src/ILGPU/Frontend/Intrinsic/LanguageIntrinsics.cs
MoFtZ: This parsing scenario is easier than parsing `{0}`, so it could probably be done. I'll give it a try.

Src/ILGPU/Runtime/Accelerator.cs
MoFtZ: I was originally using `int MajorVersion` and `int MinorVersion`. This allowed easy matching in `IsPTXVersionAtLeast(major, minor)`. Later, I changed to a single `int LanguageVersion` so that the accelerator specialisation did not have to do two passes.

What would this enumeration look like? And how would we map `IsPTXVersionAtLeast(major, minor)` to this enumeration?

Src/ILGPU/IR/Values/LanguageValues.cs
MoFtZ: Fixed.

Src/ILGPU/Runtime/Accelerator.cs
m4rs-mt: Sounds good 👍 Maybe we can encapsulate the `int` in a parent structure that can be extended in the future without patching all uses of `int` 🤓

Src/ILGPU/Frontend/Intrinsic/LanguageIntrinsics.cs
MoFtZ: Done.

Src/ILGPU/Runtime/Accelerator.cs
MoFtZ: Changed to an opaque enum (`AcceleratorLanguageVersion`) to make future refactoring easier.

MoFtZ:
Replaced `int` language version with an opaque enum to make future refactoring easier.

Added support for all basic value types - previously did not support `Int1` and `Int8`. The user needs to take care since `Int1` maps to a predicate register, and `Int8` maps to a `Int16` register.

Changed PTX expression parsing to use regular expressions.

Changed the optimiser to deal with the known issue of removing PTX emit commands from branches that will never be reached. This was achieved by adding a new `PartialInlineMode` for performing partial IR rebuilding, and also changing the transformations used in `O0` mode. Also changed `&&` to `&` to avoid an issue with phi nodes that could not be folded, and therefore preventing the removal of some branches.

Src/ILGPU/Frontend/Intrinsic/LanguageIntrinsics.cs
MoFtZ: Added regex parsing of PTX expressions.

Src/ILGPU/IR/Construction/IRRebuilder.cs
```diff
@@ -130,6 +131,111 @@ public interface IMode
...
+                in TRemapper remapper,
+                ref BasicBlockMap<BasicBlock.Builder> mapping)
+                where TRemapper : struct, IDirectTargetRemapper
+            {
+                // Create new blocks if the predicate returns true. Otherwise, use the
+                // existing block.
+                foreach (var block in blocks)
+                {
+                    if (Predicate.Apply(block))
+                    {
```
m4rs-mt, line 170: nit: remove parenthesis 🤓

Src/ILGPU/IR/Transformations/AcceleratorSpecializer.cs
```diff
@@ -158,8 +181,23 @@ public class AcceleratorSpecializer : SequentialUnorderedTransformation
...
+        {
             Specialize(context, value, (int)data.AcceleratorType);
+            data.MarkForRebuild(value.BasicBlock);
+        }
+
+        /// <summary>
+        /// Specializes accelerator inline language major version values.
+        /// </summary>
+        private static void Specialize(
+        RewriterContext context,
```
m4rs-mt, line 194: nit: Identation.

Src/ILGPU/IR/Transformations/AcceleratorSpecializer.cs
```diff
@@ -382,7 +437,47 @@ static AcceleratorSpecializer()
...
+                builder.SourceBlocks);
+
+            // Create an exit return
+            var (exitBlock, exitValue) = rebuilder.Rebuild();
+            exitBlock.CreateReturn(exitValue.Location, exitValue);
+
+            // Merge the entry block
+            var entryBlock = builder.EntryBlockBuilder;
+            entryBlock.Clear();
+            entryBlock.MergeBlock(rebuilder.EntryBlock);
```
m4rs-mt, line 478: nice one, this is correct 👍

Src/ILGPU/IR/Transformations/Optimizer.cs
```diff
@@ -164,16 +160,18 @@ public static class Optimizer
...
             // LowerPointerViews, AcceleratorSpecializer and AddressSpaceSpecializer
-            builder.Add(new LowerStructures());
+            if (level > OptimizationLevel.O0)
+                builder.Add(new LowerStructures());
 
-            // Apply DCE phase in release mode to remove all dead values that
-            // could be created in prior passes
+            // Apply DCE phase to remove all dead values that could be created in prior
+            // passes
+            builder.Add(new UnreachableCodeElimination());
```
m4rs-mt, line 168: As we discussed, I fully agree that we need to apply these transformations in all cases. However, I still believe that the additional "DCE" pass in debug mode is not required. Am I right about that?

Src/ILGPU/IR/Transformations/AcceleratorSpecializer.cs
```diff
@@ -382,7 +437,47 @@ static AcceleratorSpecializer()
...
+                IsInCollectionPredicate<HashSet<BasicBlock>>>(predicate);
+            var rebuilder = builder.CreateRebuilder(
+                builder.Method.CreateParameterMapping(parameterValues),
+                default,
+                partialMode,
+                builder.SourceBlocks);
+
+            // Create an exit return
+            var (exitBlock, exitValue) = rebuilder.Rebuild();
+            exitBlock.CreateReturn(exitValue.Location, exitValue);
```
m4rs-mt, line 473: I think this is correct since we are rebuilding all `affected` blocks. In other words, if we encounter an exit block, this block can perform a `return` operation only. All other blocks should have already been rewired to their successor blocks using  `Branch` terminator. Am I right?

Src/ILGPU/IR/Construction/IRRebuilder.cs
```diff
@@ -460,6 +637,31 @@ private void InitPhiValues(in BlockCollection blocks)
...
+            {
+                var phiRemapper = new PartialBlockRemapper(this);
+
+                foreach (var node in valueMapping)
+                {
+                    var oldNode = node.Key;
+                    var newNode = node.Value;
+
+                    if (oldNode != newNode &&
+                        oldNode.CanBeReplaced &&
```
m4rs-mt, line 653: Should we assert that this value can be replaced? Are there any other cases in which this can happen?

m4rs-mt (changes requested):

Src/ILGPU/IR/Construction/IRRebuilder.cs
MoFtZ: Done.

Src/ILGPU/IR/Transformations/AcceleratorSpecializer.cs
MoFtZ: Done.

Src/ILGPU/IR/Transformations/AcceleratorSpecializer.cs
MoFtZ: Actually, I was not sure if this was always required. The original method should already have a an exit block and return statement, so is it really required? I copied this code from the constant propagation branch.

MoFtZ:
Removed code that supported selectively detecting the PTX version. May revisit in the future, possible when there is lambda support.

m4rs-mt (approved):

PR #468, opened 2021-05-03
Build: Enable TreatWarningsAsErrors by default

mfkl:


MoFtZ:
@mfkl what is the benefit of enabling warnings as errors during normal development?

Currently, it is enabled during the CI build, which ensures that all warnings are resolved. However, during development, I personally prefer not to treat warnings as errors. This is because I do not want to have to resolve warnings such as "warning CS1591: Missing XML comment for publicly visible type or member" until much later.

mfkl:
> @mfkl what is the benefit of enabling warnings as errors during normal development?

Just a matter of taste I guess. Enforcing good practices as early as possible, making sure to fix all the mistakes before opening a PR by having the same build settings locally as the build server (lower feedback loop).

Just a suggestion, feel free to close :-)

Yey007:
@mfkl I think this a good idea in general, but I don't know how much it will help this project specifically. We ignore a lot of warnings in the code for mostly valid reasons, but I guess this will force us to stop the warnings in one way or another, and we may ultimately decide to actually fix it instead of disable it.

However, I agree with @MoFtZ: I don't want my build to fail every time I forget to write a doc comment, especially since I'm often just testing stuff out. Perhaps we could enable this feature for only a certain level of warnings? I'm not sure if that's a thing but I'd be fine with it failing to build on a "legitimate" warning but building if it's something like a doc comment.

mfkl:
> Perhaps we could enable this feature for only a certain level of warnings?

Sure, use `<NoWarn>` https://docs.microsoft.com/en-us/visualstudio/ide/how-to-suppress-compiler-warnings?view=vs-2019

Enabling this was just a suggestion, feel free to close if you think it doesn't add value to ILGPU.

m4rs-mt:
We discussed this PR during two of our talk-to-dev meetings. @mfkl the community currently agreed to keep the current build settings to simplify development and evaluation contributions. However, @mfkl, we will maintain treating of warnings as errors within the CI build. This will ensure that the project can be built without serious compilation issues/warnings if contributors are willing to submit their contribution via a pull request.

PR #471, opened 2021-05-06
Fixed incorrect simplification of ConvertValue when truncating value.

### LINKED ISSUE #470: Regression in debug assertions on array view indexer.
Reported by community member.
Introduced by #433, affecting v0.10.1 and development branch.

The following kernel triggers a debug assertion.

```CSharp
internal static void MyKernel(Index1 index, ArrayView<ulong> input, ArrayView<int> output)
{
    int i = (int)input[index];
    output[i] = i;
}

public static void Main()
{
    using var context = new Context();
    using var accelerator = new CudaAccelerator(context);
    using var inputBuffer = accelerator.Allocate<ulong>(new[] { 0x4000000000U });
    using var outputBuffer = accelerator.Allocate<int>(inputBuffer.Length);

    var kernel = accelerator.LoadAutoGroupedStreamKernel<Index1, ArrayView<ulong>, ArrayView<int>>(MyKernel);
    kernel(inputBuffer.Length, inputBuffer.View, outputBuffer.View);
    accelerator.Synchronize();
}
```

It appears that in the kernel, casting the input value to integer is not truncating.
This is causing `output[i]` to fail with a debug assertion for index out of range.

MoFtZ:
Fixes #470.

For #470, we were converting from Int64 to Int32, then to Int64 (in the debug assertions). However, the conversion was incorrectly simplified into Int64 to Int64, without the truncation to Int32 in the middle.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -283,7 +283,7 @@ Src/ILGPU.Tests/BinaryIntOperations.cs
 Src/ILGPU.Tests/CompareFloatOperations.cs
 Src/ILGPU.Tests/CompareIntOperations.cs
 Src/ILGPU.Tests/ConvertFloatOperations.cs
-Src/ILGPU.Tests/ConvertIntOperations.cs
+Src/ILGPU.Tests/ConvertIntOperations.Generated.cs
 Src/ILGPU.Tests/FixedBuffers.cs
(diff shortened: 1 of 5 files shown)


m4rs-mt (approved):
Awesome work 🚀

PR #478, opened 2021-05-19
Updated InlineList implementation to reduce memory consumption.

m4rs-mt:
The `InlineList` data structure is one of the most fundamental parts of the ILGPU compiler infrastructure. This PR addresses a "problem" where the `InlineList` reserves too much memory in certain cases. Furthermore, it fixes a critical issue in which the element count is not setup properly.

Note that this PR does not contain any test cases, since the `InlineList` is implicitly tested in all available test cases.

### DIFF
--- a/Src/ILGPU/Util/InlineList.cs
+++ b/Src/ILGPU/Util/InlineList.cs
@@ -387,8 +387,9 @@ namespace ILGPU.Util
             int count,
             ref InlineList<T> list)
         {
-            list.Reserve(Count + count);
+            list.Reserve(Math.Max(list.Capacity, count));
             Array.Copy(items, startIndex, list.items, 0, count);
+            list.Count = Math.Max(list.Count, count);
         }
 
         /// <summary>


Src/ILGPU/Util/InlineList.cs
```diff
@@ -387,7 +388,7 @@ public void MoveTo(ref InlineList<T> list)
             int count,
             ref InlineList<T> list)
         {
-            list.Reserve(Count + count);
+            list.Reserve(Math.Max(list.Capacity, count));
             Array.Copy(items, startIndex, list.items, 0, count);
```
MoFtZ, line 392: ISSUE: At this point, `list.Count` may be incorrect. It should be fixed here before returning.

```suggestion
            Array.Copy(items, startIndex, list.items, 0, count);
            list.Count = Max.Math(list.Count, count);
```

Src/ILGPU/Util/InlineList.cs
```diff
@@ -373,6 +373,7 @@ public void MoveTo(ref InlineList<T> list)
         {
             var result = Create(count);
             SliceTo(startIndex, count, ref result);
+            result.Count = count;
```
MoFtZ, line 376: ISSUE: This should be fixed in `SliceTo`, because `SliceTo` is a public method.
```suggestion
```

MoFtZ (changes requested):

Src/ILGPU/Util/InlineList.cs
m4rs-mt: Definitely 👍 !

MoFtZ (approved):

PR #480, opened 2021-05-20
Added page locking on pinned memory.

### LINKED ISSUE #472: Feature request: performance improvement for already pinned 1Dim arrays?
From . Net5 onwards it is possible to pin arrays via GC.AllocateArray(). Unfortunately, as far as I know, there is no option to check whether the array is pinned. 
Is it possibile to generate an IntPtr for the already pinned memory using Fixed{} and thus to avoid one copy operation for integral and floating-point numeric type arrays? 
I see that ArrayViews are currently under development in #421, so perhaps that would be an additional option, although an unsafe one.
Congrats on that great project!

MoFtZ:
Fixes #472.

This is a proposed API for dealing with page-locking memory on Cuda. On other accelerators, the page locking functionality throws a not supported exception.
QUESTION: Should this be a no-op instead?

Added a new `accelerator.CreatePageLockFromPinned` function that returns a `PageLockScope`. This currently accepts 1D arrays, or an `IntPtr` with `long` length. Pinning memory could take time, so the scope can be created once and retained. It should be disposed when no longer required.

Also added new `CopyFromPageLockedAsync` and `CopyToPageLockedAsync` methods that copy from `PageLockScope` without synchronising with the stream.

QUESTION: My device returns `false` for `CU_DEVICE_ATTRIBUTE_CAN_USE_HOST_POINTER_FOR_REGISTERED_MEM`. So, I registered the host memory using `CU_MEMHOSTREGISTER_DEVICEMAP`, and then retrieved the device pointer using `cuMemHostGetDevicePointer`. Unfortunately, that fails with `invalid device context`. However, using the host pointer works just fine.. any thoughts?

### DIFF
--- a/Src/ILGPU/Runtime/Cuda/CudaNativeMethods.cs
+++ b/Src/ILGPU/Runtime/Cuda/CudaNativeMethods.cs
@@ -272,6 +272,33 @@ namespace ILGPU.Runtime.Cuda
         CU_EVENT_INTERPROCESS = 4,
     }
 
+    /// <summary>
+    /// Represents flags used to register host memory with the CUDA.
+    /// </summary>
+    [Flags]
+    public enum MemHostRegisterFlags
+    {
+        /// <summary>
+        /// Host memory is portable between CUDA contexts.
+        /// </summary>
+        CU_MEMHOSTREGISTER_PORTABLE = 1,
+
+        /// <summary>
+        /// Host memory is mapped into CUDA address space.
+        /// </summary>
+        CU_MEMHOSTREGISTER_DEVICEMAP = 2,
+
+        /// <summary>
+        /// Memory pointer is treated as pointing to some memory-mapped I/O space.
+        /// </summary>
+        CU_MEMHOSTREGISTER_IOMEMORY = 4,
+
+        /// <summary>
+        /// Memory pointer is treated as pointing to memory that is considered read-only.
+        /// </summary>
+        CU_MEMHOSTREGISTER_READ_ONLY = 8,
+    }
+
     #endregion
 }
 
--- a/Src/ILGPU.Tests/Configurations.txt
+++ b/Src/ILGPU.Tests/Configurations.txt
@@ -9,6 +9,7 @@ Indices: Debug, Release, O2
 KernelEntryPoints: Debug, Release, O2
 MemoryBufferOperations : Debug, Release, O2
 MemoryCacheOperations : Debug, Release, O2
+PageLockedMemory: Debug, Release, O2
 ProfilingMarkers: Debug, Release, O2
 SharedMemory: Debug, Release, O2
 SizeOfValues: Debug, Release, O2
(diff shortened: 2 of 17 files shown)


m4rs-mt:
@MoFtZ Great work 👍 I think we should "map" unsupported page-locking functionality to a `Trace.WriteLine` + "`nop`" operation. This allows users to still write fully portable programs with respect to the current high-level API 🤓. 

> QUESTION: My device returns false for CU_DEVICE_ATTRIBUTE_CAN_USE_HOST_POINTER_FOR_REGISTERED_MEM.

Oh, I will try your PR on my NVIDIA workstation.

MoFtZ:
> Update: My workstation supports this feature which simplifies the Cuda pagelocking interop. I think we need to distinguish between scenarios where this flag is supported or not.

This could be a problem. My GTX 1070 reports `CU_DEVICE_ATTRIBUTE_CAN_USE_HOST_POINTER_FOR_REGISTERED_MEM` = false, even though it works. And the same device does not accept calls to `cuMemHostGetDevicePointer`, complaining about `invalid device context`.

Anyway, I have updated the PR to check for this flag. Let me know how you go.

Src/ILGPU/Runtime/Cuda/CudaNativeMethods.cs
```diff
@@ -248,6 +248,49 @@ enum Peer2PeerAttribute
...
     }
 
+    /// <summary>
+    /// Represents flags used to register host memory with the CUDA.
+    /// </summary>
+    [Flags]
+    public enum MemHostRegisterFlags
+    {
+        /// <summary>
+        /// If set, host memory is portable between CUDA contexts.
```
m4rs-mt, line 258: I am not sure if we are allowed to include the driver API documentation here... I would feel better if we just briefly summarize in one line roughly what the content is.

m4rs-mt (approved):

MoFtZ:
Resolved conflicts:
- `Src\ILGPU\Runtime\Cuda\CudaNativeMethods.cs`
- `Src\ILGPU.Tests\Configurations.txt`

MoFtZ:
ISSUE: Does not work on my GTX 1070.

MoFtZ:
Tested on:
- RTX 2080
- RTX 2070
- RTX 2060 Super
- GTX 1070

All using the latest Cuda drivers (v11.3).
All report that `CU_DEVICE_ATTRIBUTE_CAN_USE_HOST_POINTER_FOR_REGISTERED_MEM` is not supported.
All fail when calling `cuMemHostGetDevicePointer`.

All work fine when directly using the host pointer.

Based on the Cuda Driver API documentation for [Unified Addressing](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__UNIFIED.html#group__CUDA__UNIFIED), I believe that using the host pointer will work with ILGPU because:
- we are using a 64-bit process, which uses Unified Addressing
- and we are using the host pointer for copying to/from memory

> Either the host or the unified device pointer value may be used to refer to this memory through cuMemcpy() and similar functions using the CU_MEMORYTYPE_UNIFIED memory type.

MoFtZ:
Attempting to replicate the issue using C++ has shown that there exists `cuMemHostRegister_v2` and `cuMemHostGetDevicePointer_v2` functions with identical signatures. After switching the PR to use these functions, I no longer get 'invalid device context' errors, and the PR just works.

MoFtZ:
Fixed incorrectly unregistering Device Pointer rather than Host Pointer, when the device does not support using the host pointer for registered memory.

m4rs-mt (approved):

PR #481, opened 2021-05-21
Fixed Group helper methods for multi-dimensional kernels.

### LINKED ISSUE #473: Group.IsFirstThread bug
I have a kernel which processes 2D arrays. The simplified version computes a sum of the 2D array elements as follows:

```c#
void MyKernel(ArrayView2D<float> xMat, ArrayView2D<float> oMat)
{
	var shared = SharedMemory.GetDynamic<float>();
	shared[0] = 0;
	Group.Barrier();
	
	Atomic.Add(ref shared[0], xMat[Grid.GlobalIndex);
	
	Group.Barrier();
	
	if (Group.IsFirstThread)
		oMat[0, 0] = shared[0];
}
```

Here is the kernel config code:

```c#
private KernelConfig ComputeConfig(MemoryBuffer3D<float> xMat)
{
	Index3 groupDim = new Index3((int)xMat.Width, (int)xMat.Height, 1);
	Index3 groupNum = new Index3(1, 1, 1);

	var memConfig = SharedMemoryConfig.RequestDynamic<float>(1);
	return new KernelConfig(groupNum, groupDim, memConfig);
}
```

I invoke this kernel with _2x2_ array and it returns 2x sum what is incorrect. The problem is that `Group.IsFirstThread` returns _TRUE_ for _index = [0,0]_ and _[0,1]_, while I would expect it would return _TRUE_ only for `[0,0]` index.

It looks like the `Group.IsFirstThread` returns _TRUE_ if the first _X dimension == 0_.
Is it a bug or feature? :-)

Thanks, Csaba

MoFtZ:
Fixes #473.

### DIFF
--- a/Src/ILGPU/Group.cs
+++ b/Src/ILGPU/Group.cs
@@ -160,12 +160,16 @@ namespace ILGPU
         /// <summary>
         /// Returns true if the current thread is the first in the group.
         /// </summary>
-        public static bool IsFirstThread => IdxX == 0;
+        public static bool IsFirstThread =>
+            IdxX == 0 & IdxY == 0 & IdxZ == 0;
 
         /// <summary>
         /// Returns true if the current thread is the last in the group.
         /// </summary>
-        public static bool IsLastThread => IdxX == DimX - 1;
+        public static bool IsLastThread =>
+           (IdxX == DimX - 1) &
+           (IdxY == DimY - 1) &
+           (IdxZ == DimZ - 1);
 
         #endregion
 


m4rs-mt (changes requested):
I think we should change `&&` to `&` in this context to simplify the code structure so that we also benefit from all optimizations in `debug` builds 🤓.

MoFtZ:
Changed `&&` to `&`.
Also put brackets around `IsLastThread` to ensure correct operator precedence.

m4rs-mt (approved):
🚀

PR #482, opened 2021-05-24
Added stream profiling markers.

### LINKED ISSUE #399: Feature request: add the ability to accurately measure kernel runtime
Cuda (as well as opencl) supports events that measure kernel runtime.
And it would be useful to add the ability to use them in ILGPU.

Cuda example:
```cuda
  // Allocate CUDA events that we'll use for timing
  cudaEvent_t start, stop;
  checkCudaErrors(cudaEventCreate(&start));
  checkCudaErrors(cudaEventCreate(&stop));

  // Record the start event
  checkCudaErrors(cudaEventRecord(start, stream));

  MatrixMulCUDA<32><<<grid, threads, 0, stream>>>(d_C, d_A, d_B, dimsA.x, dimsB.x);

  // Record the stop event
  checkCudaErrors(cudaEventRecord(stop, stream));

  // Wait for the stop event to complete
  checkCudaErrors(cudaEventSynchronize(stop));

  float msecTotal = 0.0f;
  checkCudaErrors(cudaEventElapsedTime(&msecTotal, start, stop));
```

MoFtZ:
Fixes #399.

This is a proposed PR to add high resolution timers for Cuda and OpenCL accelerators.

It introduces the concept of a `ProfilingMarker` that can be added to an `AcceleratorStream`.
For OpenCL, the user is required to enable profiling - `contextBuilder.Profiling()`.

Example usage:
```CSharp
using var start = accelerator.DefaultStream.AddProfilingMarker();
launchKernelToBeProfiled();
using var end = accelerator.DefaultStream.AddProfilingMarker();

TimeSpan elapsedTime = end - start; // or end.MeasureFrom(start);
```

NOTE: On the CPU accelerator, it currently uses `DateTime.UtcNow`, which apparently has approximately 15 millisecond accuracy. If this is not acceptable, we will need to investigate a cross-platform high-resolution timer.

NOTE: There appears to be an issue with OpenCL. A workaround was required where the OpenCL event had to be awaited immediately, otherwise the clock timer did not appear to start. Unsure if this is a driver bug on my Intel HD 630.

### DIFF
--- a/Src/ILGPU/Runtime/AcceleratorStream.cs
+++ b/Src/ILGPU/Runtime/AcceleratorStream.cs
@@ -9,8 +9,10 @@
 // Source License. See LICENSE.txt for details
 // ---------------------------------------------------------------------------------------
 
+using ILGPU.Resources;
 using System;
 using System.Diagnostics.CodeAnalysis;
+using System.Runtime.CompilerServices;
 using System.Threading.Tasks;
 
 namespace ILGPU.Runtime
@@ -68,6 +70,23 @@ namespace ILGPU.Runtime
         /// </summary>
         /// <returns>A scoped binding object.</returns>
         public ScopedAcceleratorBinding BindScoped() => Accelerator.BindScoped();
+
+        /// <summary>
+        /// Adds a profiling marker into the stream.
+        /// </summary>
+        /// <returns>The profiling marker.</returns>
+        public ProfilingMarker AddProfilingMarker() =>
+            Accelerator.Context.Properties.EnableProfiling
+            ? AddProfilingMarkerInternal()
+            : throw new NotSupportedException(
+                RuntimeErrorMessages.NotSupportedProfilingMarker);
+
+        /// <summary>
+        /// Adds a profiling marker into the stream.
+        /// </summary>
+        /// <returns>The profiling marker.</returns>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        protected abstract ProfilingMarker AddProfilingMarkerInternal();
 
         #endregion
     }
--- a/Src/ILGPU/Runtime/Cuda/CudaNativeMethods.cs
+++ b/Src/ILGPU/Runtime/Cuda/CudaNativeMethods.cs
@@ -248,6 +248,30 @@ namespace ILGPU.Runtime.Cuda
         CU_DEVICE_P2P_ATTRIBUTE_NATIVE_ATOMIC_SUPPORTED = 3
     }
 
+    [Flags]
+    public enum CudaEventFlags
+    {
+        /// <summary>
+        /// Default event creation flag.
+        /// </summary>
+        CU_EVENT_DEFAULT = 0,
+
+        /// <summary>
+        /// The created event should use blocking synchronization.
+        /// </summary>
+        CU_EVENT_BLOCKING_SYNC = 1,
+
+        /// <summary>
+        /// The created event does not need to record timing data.
+        /// </summary>
+        CU_EVENT_DISABLE_TIMING = 2,
+
+        /// <summary>
+        /// The created event may be used as an interprocess event.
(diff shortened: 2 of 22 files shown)


Src/ILGPU/Runtime/AcceleratorStream.cs
```diff
@@ -69,6 +69,12 @@ protected AcceleratorStream(Accelerator accelerator)
         /// <returns>A scoped binding object.</returns>
         public ScopedAcceleratorBinding BindScoped() => Accelerator.BindScoped();
 
+        /// <summary>
+        /// Adds a profiling marker into the stream.
+        /// </summary>
+        /// <returns>The profiling marker.</returns>
+        public abstract ProfilingMarker AddProfilingMarker();
```
m4rs-mt, line 76: Question: Should we also provide an extension/instance method similar to `Synchronize` that adds a marker to the default stream? However, this might be a bad idea as this functionality will only apply to the default stream. @MoFtZ what do you think?

Src/ILGPU/Runtime/OpenCL/CLAPI.xml
```diff
@@ -65,7 +65,7 @@
             ReturnType="IntPtr">
         <Parameter Name="context" Type="IntPtr" />
         <Parameter Name="device" Type="IntPtr" />
-        <Parameter Name="properties" Type="IntPtr" />
+        <Parameter Name="properties" Type="CLCommandQueueProperties" />
```
m4rs-mt, line 68: Question: Is this change correct? If so, was it wrong before?

Src/ILGPU/Runtime/OpenCL/CLAccelerator.cs
```diff
@@ -318,7 +318,7 @@ protected override CLKernel CreateKernel(CLCompiledKernel compiledKernel)
 
         /// <inheritdoc/>
         protected override AcceleratorStream CreateStreamInternal() =>
-            new CLStream(this);
+            new CLStream(this, Context.Properties.EnableProfiling);
```
m4rs-mt, line 321: Question: can we check this flag in all cases to ensure consistent behavior for all GPUs?

Src/ILGPU/Runtime/OpenCL/CLAPI.xml
MoFtZ: It was fine - previously, we always used `IntPtr.Zero`, which is the valid default.

Src/ILGPU/Runtime/OpenCL/CLAccelerator.cs
MoFtZ: OK, will update.

Src/ILGPU/Runtime/AcceleratorStream.cs
MoFtZ: It probably makes sense to add.

MoFtZ:
Updated PR.

Adding a profiling marker without enabling profiling will throw an exception.
Added `AddProfilingMarker` extension method for `Accelerator`.

Src/ILGPU/Runtime/Cuda/CudaNativeMethods.cs
```diff
@@ -248,6 +248,37 @@ enum Peer2PeerAttribute
...
+    [Flags]
+    public enum CudaEventFlags
+    {
+        /// <summary>
+        /// Default event creation flag.
+        /// </summary>
+        CU_EVENT_DEFAULT = 0,
+
+        /// <summary>
+        /// Specifies that the created event should use blocking synchronization. A CPU
```
m4rs-mt, line 260: I am not sure if we are allowed to include the driver API documentation here... I would feel better if we just briefly summarize in one line roughly what the content is 🤓

m4rs-mt (approved):

PR #491, opened 2021-05-27
Added GetPreferredDevice and GetPreferredDevices to Context

NullandKale:
This adds 2 functions to Context which provide a simple method for getting the "best" device.

I currently select the best device based on memory size, ignoring the CPU devices. 

There are 3 scenarios I can think of for the possible hardware config:

- 1 GPU computers (like mine)
- 1 dGPU and 1 iGPU (laptops)
- N dGPU and possibly 1iGPU (servers)

and 4 scenarios I can think of for Device usage

- 1 CPU for debugging. (for all hardware configs)
- 1 GPU in use (for all hardware configs)
- N matched GPUs (for the N dGPU and possibly 1iGPU hardware config)
- N unmatched GPUs in use (for 1 dGPU and 1 iGPU or the N dGPU and possibly 1iGPU hardware config)

I think this could be done with 2 or 3 functions, I implemented the 2 function method.
 
- Device GetBestDevice(bool PreferCPU = false)
- List<Device> GetBestDevices(bool PreferCPU = false, bool MatchingDevicesOnly = false)

or

- Device GetBestDevice(bool PreferCPU = false)
- List<Device> GetBestMatchingDevices(bool PreferCPU = false)
- List<Device> GetBestDevices(bool PreferCPU = false)

I am interested in peoples input on both how to pick the best device and other sets of devices that people may want to be returned.

### DIFF
--- a/Src/ILGPU/Context.cs
+++ b/Src/ILGPU/Context.cs
@@ -26,6 +26,7 @@ using System.Collections;
 using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Diagnostics;
+using System.Linq;
 using System.Reflection;
 using System.Reflection.Emit;
 using System.Runtime.CompilerServices;
@@ -382,6 +383,77 @@ namespace ILGPU
             return deviceMapping.TryGetValue(type, out var devices)
                 ? new DeviceCollection<TDevice>(devices)
                 : new DeviceCollection<TDevice>(new List<Device>());
+        }
+
+        /// <summary>
+        /// Attempts to return the most optimal single device.
+        /// </summary>
+        /// <param name="PreferCPU">Always returns CPU device 0.</param>
+        /// <returns>Selected device.</returns>
+        public Device GetBestDevice(bool PreferCPU = false)
+        {
+            if(PreferCPU)
+            {
+                return deviceMapping.TryGetValue(AcceleratorType.CPU, out var devices)
+                    ? devices.First()
+                    : throw new NotSupportedException(
+                            RuntimeErrorMessages.NotSupportedTargetAccelerator);
+            }
+
+            var sorted = Devices.Sort((d1, d2) => d1.MemorySize.CompareTo(d2.MemorySize))
+                .Where(d => d.AcceleratorType != AcceleratorType.CPU);
+
+            if(sorted.Count() < 0)
+            {
+                return deviceMapping.TryGetValue(AcceleratorType.CPU, out var devices)
+                    ? devices.First()
+                    : throw new NotSupportedException(
+                            RuntimeErrorMessages.NotSupportedTargetAccelerator);
+            }
+            else
+            {
+                return sorted.First();
+            }
+        }
+
+        /// <summary>
+        /// Attempts to return the most optimal set of devices.
+        /// </summary>
+        /// <param name="PreferCPU">Always returns first CPU device.</param>
+        /// <param name="MatchingDevicesOnly">Only returns matching devices.</param>
+        /// <returns>Selected devices.</returns>
+        public IEnumerable<Device> GetBestDevices(bool PreferCPU = false, bool MatchingDevicesOnly = false)
+        {
+            if (PreferCPU)
+            {
+                return deviceMapping.TryGetValue(AcceleratorType.CPU, out var devices)
+                    ? devices
+                    : throw new NotSupportedException(
+                            RuntimeErrorMessages.NotSupportedTargetAccelerator);
+            }
+
+            var sorted = Devices.Sort((d1, d2) => d1.MemorySize.CompareTo(d2.MemorySize))
+                                .Where(d => d.AcceleratorType != AcceleratorType.CPU);
+
+            if (sorted.Count() < 0)
+            {
+                return deviceMapping.TryGetValue(AcceleratorType.CPU, out var devices)
+                    ? devices
+                    : throw new NotSupportedException(
+                            RuntimeErrorMessages.NotSupportedTargetAccelerator);
+            }
+            else
+            {
+                if(MatchingDevicesOnly)
+                {
+                    return sorted.Where(
+                        d => d.AcceleratorType == sorted.First().AcceleratorType);
+                }
+                else
+                {
+                    return sorted;
+                }
+            }
         }
 
         /// <summary>
--- a/Src/ILGPU/Context.cs
+++ b/Src/ILGPU/Context.cs
@@ -390,19 +390,8 @@ namespace ILGPU
         /// </summary>
         /// <param name="preferCPU">Always returns CPU device 0.</param>
         /// <returns>Selected device.</returns>
-        public Device GetPreferredDevice(bool preferCPU)
-        {
-            if (preferCPU)
-            {
-                return GetDevice<CPUDevice>(0);
-            }
-
-            var sorted = Devices
-                .OrderByDescending(d => d.MemorySize)
-                .Where(d => d.AcceleratorType != AcceleratorType.CPU);
-
-            return sorted.FirstOrDefault() ?? GetDevice<CPUDevice>(0);
-        }
+        public Device GetPreferredDevice(bool preferCPU) =>
+            GetPreferredDevices(preferCPU, matchingDevicesOnly: false).First();
 
         /// <summary>
         /// Attempts to return the most optimal set of devices.
--- a/Src/ILGPU/Context.cs
+++ b/Src/ILGPU/Context.cs
@@ -400,7 +400,8 @@ namespace ILGPU
         /// <param name="matchingDevicesOnly">Only returns matching devices.</param>
         /// <returns>Selected devices.</returns>
         public IEnumerable<Device> GetPreferredDevices(
-            bool preferCPU, bool matchingDevicesOnly)
+            bool preferCPU, 
+            bool matchingDevicesOnly)
         {
             if (preferCPU)
             {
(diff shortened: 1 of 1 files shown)


Src/ILGPU/Context.cs
```diff
@@ -384,6 +385,77 @@ public DeviceCollection<TDevice> GetDevices<TDevice>()
                 : new DeviceCollection<TDevice>(new List<Device>());
         }
 
+        /// <summary>
+        /// Attempts to return the most optimal single device.
+        /// </summary>
+        /// <param name="PreferCPU">Always returns CPU device 0.</param>
+        /// <returns>Selected device.</returns>
+        public Device GetBestDevice(bool PreferCPU = false)
```
m4rs-mt, line 393: nit: we do not use default parameters, since they can break ABI compatibility. We have to create additional overloads that specify the parameter values. Moreover, all parameters name should be written in `CamelCase`.

Src/ILGPU/Context.cs
```diff
@@ -384,6 +385,77 @@ public DeviceCollection<TDevice> GetDevices<TDevice>()
...
+        {
+            if(PreferCPU)
+            {
+                return deviceMapping.TryGetValue(AcceleratorType.CPU, out var devices)
```
m4rs-mt, line 397: Question: may we cache this lookup operation in the beginning?

Src/ILGPU/Context.cs
```diff
@@ -384,6 +385,77 @@ public DeviceCollection<TDevice> GetDevices<TDevice>()
...
+                    ? devices.First()
+                    : throw new NotSupportedException(
+                            RuntimeErrorMessages.NotSupportedTargetAccelerator);
+            }
+
+            var sorted = Devices.Sort((d1, d2) => d1.MemorySize.CompareTo(d2.MemorySize))
```
m4rs-mt, line 403: nit: Coding style, break on `.Sort`

m4rs-mt:
@NullandKale Thanks for your PR 🚀 I just left a few small comments 🤓.

Src/ILGPU/Context.cs
MoFtZ: QUESTION: Could this be replaced with `GetCPUDevice(0)`?

Src/ILGPU/Context.cs
MoFtZ: SUGGESTION: It might fit on the line if replaced by `.OrderByDescending(d => d.MemorySize)`

Src/ILGPU/Context.cs
```diff
@@ -384,6 +385,77 @@ public DeviceCollection<TDevice> GetDevices<TDevice>()
...
+                .Where(d => d.AcceleratorType != AcceleratorType.CPU);
+
+            if(sorted.Count() < 0)
```
MoFtZ, line 406: ISSUE: When the `sorted.Count() == 0`, we are going to call `sorted.First()` which will throw an exception.

Also, not optimal to call `Count()` if you don't need the number.
Maybe just `return sorted.FirstOrDefault() ?? GetCPUDevice(0);`

MoFtZ (changes requested):
MINOR: I noticed the formatting is inconsistent e.g. space after `if`. Maybe run the code formatted in VS2019 (Ctrl+K, Ctrl+D).

Src/ILGPU/Context.cs
NullandKale: We could cache it against subsequent calls of GetPreferredDevice but caching it in the function for the call at the bottom doesn't matter. I did not add caching because I don't expect this function to be called a lot.

NullandKale: I did replace it with GetCPUDevice(0) it was the same code effectively.

Src/ILGPU/Context.cs
NullandKale: Changed to OrderByDecending I don't know what I used sort

Src/ILGPU/Context.cs
NullandKale: I replaced sorted.Count with sorted,Any() in the GetPreferredDevices, in GetPreferredDevice I used the return sorted.FirstOrDefault() ?? GetCPUDevice(0);

Src/ILGPU/Context.cs
```diff
@@ -384,6 +385,70 @@ public DeviceCollection<TDevice> GetDevices<TDevice>()
...
+        /// Attempts to return the most optimal set of devices.
+        /// </summary>
+        /// <param name="preferCPU">Always returns first CPU device.</param>
+        /// <param name="matchingDevicesOnly">Only returns matching devices.</param>
+        /// <returns>Selected devices.</returns>
+        public IEnumerable<Device> GetPreferredDevices(bool preferCPU, bool matchingDevicesOnly)
+        {
+            if (preferCPU)
+            {
+                return deviceMapping.TryGetValue(AcceleratorType.CPU, out var devices)
```
MoFtZ, line 417: This looks like `GetDevice<CPUDevice>(0)` again.

Src/ILGPU/Context.cs
```diff
@@ -384,6 +385,70 @@ public DeviceCollection<TDevice> GetDevices<TDevice>()
...
+                        d.MemorySize == toMatch.MemorySize);
+                }
+                else
+                {
+                    return sorted;
+                }
+            }
+            else
+            {
+                return deviceMapping.TryGetValue(AcceleratorType.CPU, out var devices)
```
MoFtZ, line 445: This looks like GetDevice<CPUDevice>(0) too.

Src/ILGPU/Context.cs
```diff
@@ -384,6 +385,70 @@ public DeviceCollection<TDevice> GetDevices<TDevice>()
...
+        /// Attempts to return the most optimal single device.
+        /// </summary>
+        /// <param name="preferCPU">Always returns CPU device 0.</param>
+        /// <returns>Selected device.</returns>
+        public Device GetPreferredDevice(bool preferCPU)
+        {
+            if (preferCPU)
+            {
+                return GetDevice<CPUDevice>(0);
+            }
```
MoFtZ, line 398: MINOR: The style convention in ILGPU is for bracketless single-line `if` statements.
```suggestion
            if (preferCPU)
                return GetDevice<CPUDevice>(0);
```

Src/ILGPU/Context.cs
```diff
@@ -384,6 +385,70 @@ public DeviceCollection<TDevice> GetDevices<TDevice>()
                 : new DeviceCollection<TDevice>(new List<Device>());
         }
 
+        /// <summary>
+        /// Attempts to return the most optimal single device.
+        /// </summary>
+        /// <param name="preferCPU">Always returns CPU device 0.</param>
+        /// <returns>Selected device.</returns>
+        public Device GetPreferredDevice(bool preferCPU)
```
MoFtZ, line 393: SUGGESTION: Replace the implementation of this function, and call `GetPreferredDevices` instead.

```CSharp
public Device GetPreferredDevice(bool preferCPU) =>
    GetPreferredDevices(preferCPU, matchingDevicesOnly: false).First();
```

Less code to maintain.

MoFtZ (changes requested):

Src/ILGPU/Context.cs
NullandKale: This returns the entire list, GetDevice<CPUDevice>(0) only returns 1.

MoFtZ: ok, so maybe use `GetCPUDevices()` then.

m4rs-mt: Is there a consensus on this?

Src/ILGPU/Context.cs
m4rs-mt: Same here... :nerd

Src/ILGPU/Context.cs
MoFtZ: The PR is correct - `GetCPUDevices` returns `DeviceCollection`, which does not implement `IEnumerable<>T`.

Src/ILGPU/Context.cs
MoFtZ: The PR is correct - `GetCPUDevices` returns `DeviceCollection`, which does not implement `IEnumerable<>T`.

Src/ILGPU/Context.cs
```diff
@@ -384,6 +385,59 @@ public DeviceCollection<TDevice> GetDevices<TDevice>()
...
+            GetPreferredDevices(preferCPU, matchingDevicesOnly: false).First();
+
+        /// <summary>
+        /// Attempts to return the most optimal set of devices.
+        /// </summary>
+        /// <param name="preferCPU">Always returns first CPU device.</param>
+        /// <param name="matchingDevicesOnly">Only returns matching devices.</param>
+        /// <returns>Selected devices.</returns>
+        public IEnumerable<Device> GetPreferredDevices(
+            bool preferCPU, bool matchingDevicesOnly)
```
m4rs-mt, line 403: nit: please break after the first parameter.

m4rs-mt (changes requested):
After you have fixed the last formatting nits (tiny remarks 🤓), this PR should be ready to land.

Src/ILGPU/Context.cs
NullandKale: Fixed, is a formatting guide linked anywhere for ILGPU?

m4rs-mt: There is no explicit documentation. What I recommend, however, is to always look at the surrounding code (e.g. https://github.com/m4rs-mt/ILGPU/blob/c26dd77731ea2779c5c33ae9d91a74f85917978a/Src/ILGPU/Context.cs#L224-L228)

m4rs-mt (approved):

PR #498, opened 2021-06-06
Extended page locking features, introduced PageLockedND arrays and removed ExchangeBuffer.

m4rs-mt:
This PR removes the `ExchangeBuffer` class in favor of newly implemented `PageLockedArrayND` types based on the `PageLockScope` (added in PR #480). These multi-dimensional array wrappers automatically pin and lock .Net arrays in memory to improve transfer speed.

This PR also adds a `PageLockMode` enumeration to automatically enable (and even force) page locking of implicitly instantiated memory buffers during transfer operations.

At the same time, this PR also addresses community feedback regarding the inconsistent naming of `GetAsNDArray` vs. `ArrayViewND` names by renaming `GetAsNDArray` to `GetAsArrayND` methods.

Sample:
```c#
// Specify automatic page locking during transfer operations
var context = Context.Create(builder => ... builder.PageLocking(PageLockingMode.Auto));
...

// Rely on disabled (default) or automatic page locking behavior
var array = buffer.GetAsArray1D();
// vs. an explicitly page locked version.
var pagedArray = buffer.GetAsPagedLockedArray1D();

// Allocate a page locked .Net array
var array = accelerator.AllocatePageLockedArray1D<long>(...);
```

Note that there are no extension methods for copying to/from `PageLockedArrayND` instances for all `ArrayViewND<Type, Stride>` specializations. This is because some transfer operations involve transposes, which in turn involve temporary transpose buffers.


Src/ILGPU/Runtime/Cuda/CudaPageLockScope.cs
```diff
@@ -32,25 +32,38 @@ public sealed class CudaPageLockScope<T> : PageLockScope<T>
...
-            if (accelerator.Device.SupportsUsingHostPointerForRegisteredMemory)
+
+            // Perform the memory registration.
+            var cudaError = CurrentAPI.MemHostRegister(
+                hostPtr,
+                new IntPtr(LengthInBytes),
+                flags);
+
+            // Check whether we have already registered this memory pointer.
+            Locked = cudaError != CudaError.CUDA_ERROR_HOST_MEMORY_ALREADY_REGISTERED;
```
MoFtZ, line 60: ISSUE: When the memory address is already registered by another scope, there is no guarantee that the other scope will keep the registration longer than this scope.

MoFtZ (changes requested):

Src/ILGPU/Runtime/Cuda/CudaPageLockScope.cs
m4rs-mt: @MoFtZ Yes, you are right 🤓 I think we should revert these changes.

MoFtZ: Maybe we could introduce a reference-counting system?

m4rs-mt: @MoFtZ I'm not sure it's worth introducing reference counting in this context.

MoFtZ (approved):

PR #499, opened 2021-06-08
Fixed emitting IL for specialized kernel launcher with many arguments.

MoFtZ:
A member of the community raised an issue when using a kernel with a large number of arguments of `SpecializedValue`.

When the kernel uses 6 `SpecializedValue` arguments, everything worked fine. However, when the kernel uses 7 `SpecializedValue` arguments, ILGPU would throw the exception: `System.NotSupportedException: Illegal one-byte branch at position: 23. Requested branch was: 150.`

It has been identified as an issue in generating the specialized kernel launcher - if there are too many specialized fields, the IL generated in the kernel launcher would incorrectly use a short (1-byte) branch instruction. This PR attempts to address this issue by changing to use a long (4-byte) branch based on the number of specialized fields.

NOTE: It has been noticed that the IL emitted in `net5.0` is shorter than `net47` and `netcoreapp3.1`. `net5.0` uses 17 IL bytes per field, compared to 22 bytes per field. This means that `net5.0` could fit 8 specialized fields before switching to a long branch, compared to 7 specialized fields for `net47` and `netcoreapp3.1`.


m4rs-mt (approved):
Nice one 👍

PR #504, opened 2021-06-12
Fixed regression copying from/to OpenCL accelerators.

MoFtZ:
Bug introduced as part of https://github.com/m4rs-mt/ILGPU/pull/498.

When copying between CPU and Cuda, the code is the same regardless of the type of the source and target buffers. However, OpenCL has different code paths, depending on if the source/target buffers are CPU or OpenCL. The issue that is occurring is that both the source and target buffers are being treated as OpenCL, rather than one being CPU.


m4rs-mt (approved):

MoFtZ:
Fixed another issue in `CLMemoryBufferOperations`, `CopyFromToCPU_IntXX` unit tests. Looks like it's been there since at least ILGPU v1.0-beta1.

m4rs-mt (approved):

PR #510, opened 2021-06-21
Improve Index Hash Codes

Nnelg:
Introduce bit-shifting into hash code algorithm, so that the order of values matters. (Under the old method, (1, 2, 3) would collide with (1, 3, 2) etc.)


Src/ILGPU/IndexTypes.tt
```diff
@@ -690,4 +695,4 @@ namespace ILGPU
     }
 
 <#  } #>
-}
```
MoFtZ, line 693: ISSUE: Please remove this change.

We currently enforce that T4 should not end with a newline on its own. This is because there are known issues with transforming T4 templates on other platforms.

Src/ILGPU/IndexTypes.tt
```diff
@@ -362,8 +362,13 @@ namespace ILGPU
...
         /// <returns>The hash code of this index.</returns>
-        public readonly override int GetHashCode() =>
-            <#= def.Expression(" ^ ", p => $"{p}.GetHashCode()") #>;
+        public readonly override int GetHashCode()
+        {
+            int hash = <#= def.Expression(
+                $"{Environment.NewLine}            hash = (hash << 1) ^ (hash >> 1) ^ ",
+                p => $"{p}.GetHashCode();") #>
+            return hash;
+        }
```
MoFtZ, line 371: SUGGESTION: What about using a ValueTuple? i.e. `(X, Y, Z).ToHashCode()`.
Or will this not work on the GPU because it is using `System.HashCode`?

```CSharp
<# if (dimension == 1) { #>
            <#= def.Expression(", ", p => $"{p}") #>.GetHashCode();
<# } else { #>
            (<#= def.Expression(", ", p => $"{p}") #>).GetHashCode();
<# } #>
```

MoFtZ (changes requested):

Src/ILGPU/IndexTypes.tt
Nnelg: I couldn't find anything about how `ValueTuple.GetHashCode()` is implemented beyond it being guaranteed that equal values will have equal hash codes, so I have to assume it _won't_ solve the issue.

No idea about what'll go on inside the GPU, my concern was solely how it's used on the CPU. I've got a use-case where I've got a dictionary that's keyed on 2D grid locations inside an algorithm that heavily utilizes the GPU, and I didn't want to write a redundant implementation of basically just (int x, int y) for it.

MoFtZ: Internally, `ValueTuple` it is using `System.HashCode`:
https://github.com/dotnet/runtime/blob/v5.0.7/src/libraries/System.Private.CoreLib/src/System/ValueTuple.cs#L566

Based on the source code, `System.HashCode` is using xxHash32:
https://github.com/dotnet/runtime/blob/v5.0.7/src/libraries/System.Private.CoreLib/src/System/HashCode.cs#L51

https://github.com/Cyan4973/xxHash

> xxHash is an Extremely fast Hash algorithm, running at RAM speed limits. It successfully completes the SMHasher test suite which evaluates collision, dispersion and randomness qualities of hash functions. Code is highly portable, and hashes are identical across all platforms (little / big endian).

Src/ILGPU/IndexTypes.tt
MoFtZ: This is probably your text editor. It has been configured to save and ensure there is a blank newline. Try disabling the setting, or using a different text editor to make the change. Alternatively, you can use git to remove the change.

Nnelg: I'm using github's web editor. I looked for a setting, but couldn't find anything. I don't know how to use a different editor on github, or git itself.

Src/ILGPU/IndexTypes.tt
Nnelg: Okay, looking deeper into that, it seems like using the ValueTuple might be a better approach. (Something about spreading the hashcodes out more...)

Src/ILGPU/IndexTypes.tt
MoFtZ: I couldn't find any solution when using Github's web editor.

I've raised a PR into your fork of ILGPU:
https://github.com/Nnelg/ILGPU/pull/1

If you accept the PR, it should fix the newline issue.

Nnelg: I want to hold off on this until we're sure the rest of the code is good, because any time I edit it'll be put right back in...

MoFtZ (approved):

m4rs-mt (approved):
Awesome work 🥇 Thanks a lot for this extension!

PR #513, opened 2021-06-28
Refined processing of arrays within ILGPU.

### LINKED ISSUE #493: Regression in writing to field of structure causing Cuda misaligned address in v1.0-beta1
A community member has reported a regression in v1.0-beta1.
Previously worked in v0.10.0.

The issue is that assigning a value to `Fifth` in the data structure is generating `CudaException: 'misaligned address'` at runtime.

The reduced version is below. Uncommenting `Fourth` alters the structure layout, and the misalignment does not occur.
```CSharp
namespace Test
{
    public class Program
    {
        public struct MyData
        {
            public byte[] First;
            public int Second;
            public int Third;
            //public int Fourth;
            public long Fifth;
        }

        public static void MyKernel(Index1D index, ArrayView<long> view)
        {
            var d = new MyData();
            d.Fifth = 42;   // <-- misaligned address here
            view[index] = d.Fifth;
        }

        static unsafe void Main()
        {
            using var context = Context.Create(x => x.Default());
            using var buffer = accelerator.Allocate1D<long>(1024);

            var kernel = accelerator.LoadAutoGroupedStreamKernel<Index1D, ArrayView<long>>(MyKernel);
            kernel((int)buffer.Length, buffer.View);
            accelerator.Synchronize();
        }
    }
}
```

This appears related to the recent changes for [multi-dimensional arrays and constant arrays](#479 ).

m4rs-mt:
This PR fixes #493 and further issues that have been introduced in the scope of #479. It solves them by adding new array values to the IR (namely `NewArray`, `LoadArrayElementAddress` and `GetArrayLength`). Furthermore, it also introduces the notion of a multi-dimensional `ArrayType` in the scope of the internal type system. These features allow us to represent and preserve high-level array operations as long as possible within the ILGPU IR.

In order to convert arrays to data structures compatible with the current backends, this PR adds a new `LowerArrays` transformation. Its purpose is to lower IR array values to raw buffer allocations + view-based values. Since the internal `SSAStructureConstruction` pass is already able to convert such buffer allocations into register values (if possible), only slight adjustments to the optimization pipeline were required to apply the full optimization potential to the new array IR values.

Also included in this PR is the ability to convert 1D, 2D and 3D .Net arrays into views (in the scope of ILGPU kernels only!)
Sample:
```c#
var array = new int[32];
ArrayView1D<int, Stride1D.Dense> view = array.AsArrayView();
```


MoFtZ (changes requested):
ISSUE: Multiple unit tests are failing on my test machine.

The failures are affecting `Release` and `O2` configurations on Cuda and OpenCL.
```
e.g.
ILGPU.Tests.Cuda.CudaArrays_Release
- ArrayBounds (10 failed)
- ArrayLength (10 failed)
- ArrayLongLength (10 failed)
- MultiDimArrayBounds (5 failed)
- MultiDimArrayLength (5 failed)
```

The failures are related to the unit tests using `TestStruct`, and appear to be failing an [internal assertion](https://github.com/m4rs-mt/ILGPU/blob/arrays/Src/ILGPU/IR/Construction/Structures.cs#L506).

MoFtZ:
QUESTION: Are we able to add debug assertions when indexing .Net arrays?

```CSharp
        public static void MyKernel(Index1D index, ArrayView<int> buffer)
        {
            var y = new int[2];
            buffer[index] = index + y[y.Length];  // Can we assert here?
        }
```

Note, if we convert the .Net array to an ArrayView, the indexer correctly asserts.
```CSharp
        public static void MyKernel(Index1D index, ArrayView<int> buffer)
        {
            var y = new int[2].AsArrayView();
            buffer[index] = index + y[y.Length];  // Correctly asserts here
        }
```

m4rs-mt:
@MoFtZ I could reproduce the problem in `Release` and `O2` builds. I already know what the problem is and I will prepare a fix.

m4rs-mt:
@MoFtZ The bounds checking issue was related to invalid generation of debug assertions that checked for `<=` instead of `<`. The additional issues related to the failed tests were related to some SSA conversion issues in the `SSAStructureConstruction` pass.

MoFtZ (approved):

PR #520, opened 2021-07-01
Workflows: Create release on tag

76creates:
### Goal
Create release when tag is pushed and upload built artefacts to it, it is required that tag starts with `v`.

### Bonus side quests
- mark the release as a draft
- mark as a pre-release when there is a suffix in the tag (e.g. v1.0.0-beta2)
- set the name to Release $tag (e.g. Release v1.0.0-beta2)


.github/workflows/ci.yml
```diff
@@ -173,6 +173,22 @@ jobs:
...
 
+  release:
+    if: startsWith(github.ref, 'refs/tags/v')
+    needs: [check-style, build-and-test]
+    runs-on: ubuntu-latest
+    steps:
+      - name: Download NuGet package artifact
+        uses: actions/download-artifact@v2
+        with:
+          name: nuget-package
```
ljubon, line 184: could you please check if this should be `nuget-package` or `nuget-packages`? 💭 
per current `ci.yml` they are uploading [nuget-packages](https://github.com/m4rs-mt/ILGPU/blob/master/.github/workflows/ci.yml#L192) so you would receive here empty file? 🤔

jgiannuzzi:
Thanks for you PR, @76creates!

It seems that you did not rebase it on the latest `master`, in which the CI pipeline changed significantly to merge testing, packaging, and publishing of both ILGPU and ILGPU.Algorithms. Could you please apply the same logic on top of the latest `master`?

.github/workflows/ci.yml
76creates: nice catch, it was updated meanwhile, will adapt, ty ⚡

76creates:
Merged, resolved conflicts

.github/workflows/ci.yml
```diff
@@ -192,6 +192,22 @@ jobs:
           name: nuget-packages
           path: ./Bin/Release/ILGPU*.${{ steps.package.outputs.version }}.*nupkg
 
+  release:
+    if: startsWith(github.ref, 'refs/tags/v')
+    needs: [check-style, build-and-test]
```
MoFtZ, line 197: ISSUE: Missing reference to `package`.
```suggestion
    needs: [check-style, build-and-test, package]
```

MoFtZ (changes requested):

jgiannuzzi:
@76creates Could you please also
* mark the release as a draft, so that @m4rs-mt can then complete the release notes
* mark as a pre-release when there is a suffix in the tag (e.g. `v1.0.0-beta2`)
* set the name to `Release $tag` (e.g. `Release v1.0.0-beta2`)

jgiannuzzi:
As this job is pretty similar to the existing `publish` job (same condition, same dependencies, and same first step), maybe we could merge them together?

76creates:
> As this job is pretty similar to the existing `publish` job (same condition, same dependencies, and same first step), maybe we could merge them together?

keeping them separate might enhance readability, observing that `download-artifact` is a cheap action I don't see much time saving taking place, if you still thing it should go shoot me a 🚀 emoji :D

jgiannuzzi:
> keeping them separate might enhance readability, observing that `download-artifact` is a cheap action I don't see much time saving taking place, if you still thing it should go shoot me a 🚀 emoji :D

I kinda see this as just an extra step of the same job that is responsible for publishing the release (maybe `publish` should become `publish-release` to make that clear?). I think merging them would be more a matter of maintainability rather than any kind of optimisation. We won't risk having the condition or dependencies getting out of sync during a future refactoring.

@MoFtZ @m4rs-mt @ljubon what are your thoughts?

ljubon:
> > keeping them separate might enhance readability, observing that `download-artifact` is a cheap action I don't see much time saving taking place, if you still thing it should go shoot me a 🚀 emoji :D
> 
> I kinda see this as just an extra step of the same job that is responsible for publishing the release (maybe `publish` should become `publish-release` to make that clear?). I think merging them would be more a matter of maintainability rather than any kind of optimisation. We won't risk having the condition or dependencies getting out of sync during a future refactoring.
> 
> @MoFtZ @m4rs-mt @ljubon what are your thoughts?

hey guys, with first look I see `publish-nuget` as a step in the `release` job
imho, the release job/stage should be the last one in this CI/CD process....upload/publishing/creating/tagging should be part of the `release` job/stage 🚀 
p.s @76creates  any particular reason for using `windows-latest` agent for publishing nuggets? 💭

MoFtZ:
Merging the `publish` and `release` jobs makes sense to me - maintainability, as well as flow of steps.

However, one thought I had was, how do we recover if one of the steps fail? The order of publish/release might be important because there is no rollback if the second step fails - the first step has already applied its change.

If the package was uploaded to Nuget successfully, and the Github release step fails, can we re-run the failed job? i.e. will Nuget accept the second upload, or do we have to login to Nuget and delete the upload? Similarly, if the order was reversed, can we create the same Github release? or do we have to manually delete the created release before re-running the job?

Which brought me back full circle to, if we keep the jobs separated in the CI pipeline, we can just re-run the job that failed.. 🤯 But this seems to be optimising for the unlikely event of failure.

jgiannuzzi:
Good point @MoFtZ!

Unfortunately there is currently no way to re-run a single step or job with GitHub Actions - only the entire workflow can be re-run. And there is no way to delete a published package on nuget.org - it can only be hidden and a package with a different version number must be uploaded.

Deleting a release on GitHub is always possible though. As we want to create it as draft (so that the release notes can be manually completed), I would suggest to have this step run before the NuGet publication step. That way we always have a fail safe and can just delete the release and re-run the whole workflow in case of problem with the NuGet publication.

MoFtZ:
> I would suggest to have this step (GitHub release) run before the NuGet publication step.

Sounds good to me 👍 Even more reason to combine the publish and release steps.

76creates:
> > > keeping them separate might enhance readability, observing that `download-artifact` is a cheap action I don't see much time saving taking place, if you still thing it should go shoot me a 🚀 emoji :D
> > 
> > 
> > I kinda see this as just an extra step of the same job that is responsible for publishing the release (maybe `publish` should become `publish-release` to make that clear?). I think merging them would be more a matter of maintainability rather than any kind of optimisation. We won't risk having the condition or dependencies getting out of sync during a future refactoring.
> > @MoFtZ @m4rs-mt @ljubon what are your thoughts?
> 
> hey guys, with first look I see `publish-nuget` as a step in the `release` job
> imho, the release job/stage should be the last one in this CI/CD process....upload/publishing/creating/tagging should be part of the `release` job/stage 🚀
> p.s @76creates any particular reason for using `windows-latest` agent for publishing nuggets? 💭

`windows-latest` is not being used, i guess you saw that on #521, there its fixed as well now ⚡

76creates:
Merged two actions, added logic for marking release an 'prerelease', marked as draft by default ⚡

.github/workflows/ci.yml
```diff
@@ -133,6 +133,7 @@ jobs:
           }
           
           echo "::set-output name=version::$main_version"
+          echo "::set-output name=tag::$tag"
```
jgiannuzzi, line 136: `version` and `tag` will have the same value here, so we should only output one of them (I'd argue for `version`).

.github/workflows/ci.yml
```diff
@@ -192,7 +193,10 @@ jobs:
           name: nuget-packages
           path: ./Bin/Release/ILGPU*.${{ steps.package.outputs.version }}.*nupkg
 
-  publish:
+    outputs:
+      tag: ${{ steps.version.outputs.tag }}
```
jgiannuzzi, line 197: As per the comment above, we should probably just use `version` here instead of the new variable `tag`.

.github/workflows/ci.yml
```diff
@@ -201,5 +205,28 @@ jobs:
...
+          ;else
+            echo "::set-output name=prerelease::false" 
+          ;fi
+
+      - name: Create release
+        uses: softprops/action-gh-release@v1
+        with:
+          name: Release v${{ needs.package.outputs.tag }}
+          draft: true
+          prerelease: ${{ steps.prerelease-logic.outputs.prerelease }}
```
jgiannuzzi, line 226: The whole prerelease condition could probably be simplified by using [`contains`](https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#contains)

jgiannuzzi (commented):
Great stuff, @76creates!

I have a few remarks below, but this is mostly about polishing stuff 🤓

.github/workflows/ci.yml
76creates: nice, on the closer inspection I can see what you mean ✅

.github/workflows/ci.yml
76creates: would be nice, tho I don't see regex is supported, we could search for `-`, and if you think its good enough lets do it ⚡

jgiannuzzi: I _think_ that searching for `-` should be enough for determining it's a prerelease.

@MoFtZ @m4rs-mt do you confirm?

MoFtZ: I believe that is correct.

According to the [Nuget documentation](https://docs.microsoft.com/en-us/nuget/concepts/package-versioning#semantic-versioning-200) and the Semantic Versioning spec, [clause 9](https://semver.org/spec/v2.0.0.html#spec-item-9), says it is a hyphen.

In theory, clause 10 also talks about build metadata, which starts with +, but may contain hyphens, but it shouldn't appear in our usage.

MoFtZ: @76creates I think @jgiannuzzi was referring to removing the `prerelease-logic` step, and using `contains` as part of the `softprops/action-gh-release@v1` step.

```suggestion
          prerelease: ${{ contains(needs.package.outputs.version, '-') }}
```

MoFtZ (changes requested):

.github/workflows/ci.yml
76creates: makes total sense, I failed to see that, updated

jgiannuzzi (approved):
LGTM, thanks @76creates!

.github/workflows/ci.yml
```diff
@@ -218,5 +221,18 @@ jobs:
         uses: actions/download-artifact@v2
         with:
           name: nuget-packages
+
+      # if tag cointains suffix "-" treat it as pre-release
```
MoFtZ, line 225: MINOR: Spelling of `cointains`. Also `-` is not a suffix.

```suggestion
      # if tag contains "-" treat it as pre-release
```

MoFtZ (changes requested):
Sorry @76creates, one last change.

MoFtZ (approved):

jgiannuzzi (approved):

.github/workflows/ci.yml
jgiannuzzi: you read my mind @MoFtZ 😅
I'll use the suggestion feature next time 😄

PR #521, opened 2021-07-01
Workflows: Push pre-release packages to separate feed

76creates:
### Goal
We want to push pre-release nugets to separate feed, we should do this on master branch only.
Feed is `https://f.feedz.io/ilgpu/preview/nuget/index.json`, new secret being used is `FEEDZIO_API_KEY`, it should be created.

### Side quest
Change how we define suffix for the package name, use date-time instead of hash due to how it's sorted on the feed side.


.github/workflows/ci.yml
```diff
@@ -173,6 +173,19 @@ jobs:
...
 
+  publish-nuget-pre-release:
+    if: github.ref == 'refs/heads/master'
+    needs: [check-style, build-and-test]
+    runs-on: windows-latest
+    steps:
+      - name: Download NuGet package artifact
+        uses: actions/download-artifact@v2
+        with:
+          name: nuget-package
```
ljubon, line 184: same as the previous PR, `nuget-package` artifact doesn't exist on `m4rs-mt:master` 👀

jgiannuzzi:
Thanks for you PR, @76creates!

It seems that you did not rebase it on the latest `master`, in which the CI pipeline changed significantly to merge testing, packaging, and publishing of both ILGPU and ILGPU.Algorithms. Could you please apply the same logic on top of the latest `master`?

.github/workflows/ci.yml
76creates: updated this one, nice catch, ty ⚡

76creates:
Merged, resolved conflicts

.github/workflows/ci.yml
```diff
@@ -192,6 +192,19 @@ jobs:
           name: nuget-packages
           path: ./Bin/Release/ILGPU*.${{ steps.package.outputs.version }}.*nupkg
 
+  publish-nuget-pre-release:
+    if: github.ref == 'refs/heads/master'
+    needs: [check-style, build-and-test]
```
MoFtZ, line 197: ISSUE: Missing reference to `package`.
```suggestion
    needs: [check-style, build-and-test, package]
```

.github/workflows/ci.yml
```diff
@@ -192,6 +192,19 @@ jobs:
...
+    runs-on: windows-latest
+    steps:
+      - name: Download NuGet package artifact
+        uses: actions/download-artifact@v2
+        with:
+          name: nuget-packages
+      - name: Publish to FeedzIO
+        shell: bash
```
MoFtZ, line 205: SUGGESTION: Instead of running on `windows-latest` and using `bash` shell, just run on `ubuntu-latest`.

MoFtZ (changes requested):

jgiannuzzi:
When enabling preview versions for [Consul.NET](https://github.com/G-Research/consuldotnet), we realised that the ordering  could be wrong when just using the commit SHA as the suffix. We thus introduced a datetime in the suffix. I think we should do the same here. See https://github.com/G-Research/consuldotnet/pull/102 for more details.

jgiannuzzi (approved):
LGTM 👍

m4rs-mt (approved):

PR #522, opened 2021-07-04
Refined test infrastructure to improve performance.

m4rs-mt:
This PR refines the current test infrastructure to improve performance. It uses the ILGPU-internal caching concept and increasing the number of parallel test threads at the same time.

This PR introduces two environment variables:
```c#
ILGPU_TEST_PARALLELIZATION_LEVEL = 4 // Degree of parallelism: total num threads = 4 * 3 (num optimization levels)
```
and
```c#
ILGPU_CLEAN_TESTS = "" // Non-empty string to enable "clean tests"
```
Turning on "clean tests", all caches will be cleared after each test to test them without possible interference from other tests.


m4rs-mt:
@jgiannuzzi @76creates I think we should further parallelize the `Windows` runner tests to use different machines for the `net47` and `netcore3.1` targets. Also, it would be really great if we could group the tests into CPU and GPU tests to have a better understanding of the type of platform actually being tested.

MoFtZ (changes requested):
ISSUE: Please update `.gitignore` and remove entries for:
- `Src/ILGPU.Tests.CPU/TestContexts.cs`
- `Src/ILGPU.Tests.Cuda/TestContexts.cs`
- `Src/ILGPU.Tests.OpenCL/TestContexts.cs`

My local checkout was failing to compile because one of these generated files existed - had to manually delete the file.

m4rs-mt:
@MoFtZ nice catch 👍

MoFtZ (approved):

jgiannuzzi:
> Also, it would be really great if we could group the tests into CPU and GPU tests to have a better understanding of the type of platform actually being tested.

I'm not sure I understand what you mean by that - the GPU tests are already running separately on the `cuda` runners

m4rs-mt:
@jgiannuzzi I was referring to the build-matrix view:
image

m4rs-mt:
@jgiannuzzi I also think that we should limit the parallelization level to 2 for the GPU tests. Alternatively, we should reduce the number of concurrent Cuda runners on our build machine. The reason are possible `OutOfMemory` issues when too many contexts and tests and instantiated at the same time.

PR #523, opened 2021-07-04
Fixed invalid copies from and to sub views.

m4rs-mt:
The current implementation of `CopyFromCPU` and `CopyToCPU`, which do not require reorder operations, perform invalid data copy operations on partial sub views. This PR fixes invalid copies from and to CPU memory operating on partial sub views by adding additional checks.


MoFtZ (approved):

MoFtZ (commented):
SUGGESTION: Can we add some unit tests for the inefficiency copy exception?

m4rs-mt:
@MoFtZ Definitely! Makes sense to add 😄

MoFtZ (approved):

PR #530, opened 2021-07-07
Selective builds on macOS, master and tags

76creates:
## Preface
MacOS ILGPU builds are taking bit too long to complete, since we suspect there is a little to no one using ILGPU on MacOS we want to limit these builds to tags/master branch when doing a push.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -69,8 +69,18 @@ jobs:
       - name: Setup OS matrix
         id: setup-os-matrix
         run: |
-          os='["ubuntu-latest","windows-latest","macos-latest"'
+          os='["ubuntu-latest","windows-latest"'
+
+          # runs on macOS only if there is a push to master, or a tag is 
+          # pushed, we do this since macOS builds last too long and ILGPU 
+          # is rarely used on a macOS
+          [ "${{ github.ref }}" == "refs/heads/master" ] ||
+          [[ "${{ github.ref }}" =~ "refs/tags/v" ]] && 
+          [ "${{ github.event_name }}" == "push" ] &&
+          os="$os,\"macos-latest\""
+
           [ "${{ github.event.repository.fork }}" == "false" ] && os="$os,\"cuda\""
+
           os="$os]"
           echo "::set-output name=os::$os"
     outputs:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -69,20 +69,22 @@ jobs:
       - name: Setup OS matrix
         id: setup-os-matrix
         run: |
-          os='["ubuntu-latest","windows-latest"'
+          os=("ubuntu-latest" "windows-latest")
 
           # runs on macOS only if there is a push to master, or a tag is 
           # pushed, we do this since macOS builds last too long and ILGPU 
           # is rarely used on a macOS
-          [ "${{ github.ref }}" == "refs/heads/master" ] ||
-          [[ "${{ github.ref }}" =~ "refs/tags/v" ]] && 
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -69,8 +69,18 @@ jobs:
...
         run: |
-          os='["ubuntu-latest","windows-latest","macos-latest"'
+          os='["ubuntu-latest","windows-latest"'
+
+          # runs on macOS only if there is a push to master, or a tag is 
+          # pushed, we do this since macOS builds last too long and ILGPU 
+          # is rarely used on a macOS
+          [ "${{ github.ref }}" == "refs/heads/master" ] ||
+          [[ "${{ github.ref }}" =~ "refs/tags/v" ]] && 
+          [ "${{ github.event_name }}" == "push" ] &&
```
76creates, line 79: is this necessary?

ljubon: i think yes, since we also want to avoid `pull_request` events, so I guess it can stay

.github/workflows/ci.yml
```diff
@@ -69,8 +69,18 @@ jobs:
...
+          os="$os,\"macos-latest\""
```
ljubon, line 80: should we also add here other platforms? if all conditions are met, we want to run for all platforms (previous state), not just macOS? 🤔

76creates: ye, it basically appends to initial `os='["ubuntu-latest","windows-latest"'` so in case this part gets evaluated it would be `os='["ubuntu-latest","windows-latest","macos-latest"'`

.github/workflows/ci.yml
jgiannuzzi: good point @ljubon - we should actually do the same in the other conditions about pushes to master/tag, i.e. the publish/release jobs (talking about GHA conditions here, not shell)

jgiannuzzi: also I would reorganise and group the conditions to something like:
```bash
          [ "${{ github.event_name }}" == "push" ] &&
          ([ "${{ github.ref }}" == "refs/heads/master" ] ||
          [[ "${{ github.ref }}" =~ "refs/tags/v" ]])
```

jgiannuzzi: also also we might want to switch to using https://github.com/actions/github-script so we can do proper JSON processing (I know no one here likes JavaScript 😅) OR we could use PowerShell and its `ConvertTo-Json -Compress` cmdlet (that might fit this project better 😉)

.github/workflows/ci.yml
ljubon: oh, my bad i forgot ` os="$os,\"macos-latest\"` at beginning 
i saw onyl `macos-latest` 🤣 
looks good 👌

.github/workflows/ci.yml
76creates: json would naturally be easiest with the java script, so it would be my pick(even tho I am not a huge fan :D)
tho if we are going for a shell I would prefer bash over powerShell, i believe its a more common shell between the two, with `jq` its something like:
```
OSES="[]"
OSES=$(echo "example-os" | jq -R --argjson oses "$OSES" '$oses + [.]')
```

jgiannuzzi: I really like your second proposal! ⚡
Maybe with variable name `os` instead of `_OSES` for readability?

Something like:
```bash
os=("os1" "os2")
os+=("os3")
echo "::set-output name=os::$(jq -cn '$ARGS.positional' --args ${os[@]})"
```

jgiannuzzi (approved):

PR #532, opened 2021-07-08
Added missing struct type constraints.

MoFtZ:
Fixed missing constraint for `IControlFlowDirection` and `LowerThreadIntrinsics.ILoweringImplementation` - other uses of these interfaces already include the `struct` type constraint.

Added constraint to `IRRebuilder.IMode` - all implementations are currently structs.

These instances were found in the initial investigation into enabling nullable references types.


m4rs-mt (approved):
Nice catch 👍

PR #535, opened 2021-07-13
Changes InvalidEntryPointIndexParameterOfWrongType to be more descriptive

NullandKale:
`InvalidEntryPointIndexParameterOfWrongType` error did not help the user understand what was wrong and how to fix it, now it does!

### DIFF
--- a/Src/ILGPU/Resources/ErrorMessages.resx
+++ b/Src/ILGPU/Resources/ErrorMessages.resx
@@ -124,7 +124,7 @@
     <value>An entry point must receive at least one index parameter</value>
   </data>
   <data name="InvalidEntryPointIndexParameterOfWrongType" xml:space="preserve">
-    <value>An implicitly grouped entry point must receive an int index type as first parameter, long indecies will not work.</value>
+    <value>An implicitly grouped entry point must receive an int index type as first parameter, long indices will not work.</value>
   </data>
   <data name="InvalidEntryPointInstanceKernelMethod" xml:space="preserve">
     <value>Only static methods and non-capturing lambdas are supported at the moment</value>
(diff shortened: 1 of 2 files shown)


Src/ILGPU/Resources/ErrorMessages.resx
```diff
@@ -124,7 +124,7 @@
     <value>An entry point must receive at least one index parameter</value>
   </data>
   <data name="InvalidEntryPointIndexParameterOfWrongType" xml:space="preserve">
-    <value>An entry point must receive a compatible index type as first parameter</value>
+    <value>An implicitly grouped entry point must receive an int index type as first parameter, long indecies will not work.</value>
```
m4rs-mt, line 127: Question: typo "indecies" => "indices"?

NullandKale: Totally missed that, I will fix it when I get home in ~ 90 minutes.

Src/ILGPU/Resources/ErrorMessages.resx
```diff
@@ -124,7 +124,7 @@
     <value>An entry point must receive at least one index parameter</value>
   </data>
   <data name="InvalidEntryPointIndexParameterOfWrongType" xml:space="preserve">
-    <value>An entry point must receive a compatible index type as first parameter</value>
+    <value>An implicitly grouped entry point must receive an int index type as first parameter, long indices will not work.</value>
```
MoFtZ, line 127: MINOR: Should we use more formal language?

```suggestion
    <value>An implicitly grouped entry point must receive an int index type as first parameter, long indices are not supported.</value>
```

m4rs-mt (approved):

PR #536, opened 2021-07-14
Promote .NET 5 to a default target framework

jgiannuzzi:
This PR promotes .NET 5 to a default target framework, instead of being hidden behind the experimental `Channel` property. We now require the .NET 5 SDK to build ILGPU, and will run only the .NET 5 tests, unless running in the CI.

The CI has also been adapted (credits to @ljubon!) to run all the tests for the various target frameworks in parallel (so no more double time to run the tests on Windows because of `net47`!). The .NET SDKs are not pinned anymore but we instead always use the latest stable release. To avoid PRs failing randomly after a new release of .NET, we run this workflow daily, so that the maintainers will get a notification in case of failure.

This PR also introduces a virtual job named `all-required-checks-done` that depends on all the various jobs we want to pass before merging a PR. The branch protection settings will need to be adapted to _only_ require this one.

---

Please remember to add the following line to the commit log when merging this PR so that @ljubon gets proper credits:
```
Co-authored-by: Ljubo Nikolic <ljubonikolic@gmail.com>
```


.github/workflows/ci.yml
```diff
@@ -88,36 +95,69 @@ jobs:
...
-        if: matrix.os != 'cuda'
-        run: dotnet test --logger GitHubActions --no-build --configuration=Release Src/${{ matrix.library }}.Tests.CPU
-        env:
-          ILGPU_CLEAN_TESTS: ${{ github.event_name == 'push' && (github.ref == 'refs/heads/master' || startsWith(github.ref, 'refs/tags/v')) }}
-      - name: CUDA Tests
-        if: matrix.os == 'cuda'
-        run: dotnet test --logger GitHubActions --no-build --configuration=Release Src/${{ matrix.library }}.Tests.Cuda
+          dotnet-version: 5.0.x
+      
+      - name: Setup the latest .NET Core 3.1 SDK
```
MoFtZ, line 144: QUESTION: What is the purpose of installing the .NET Core 3.1 SDK _**and**_ the .NET 5.0 SDK, when we have pinned to .NET 5.0.x ?

jgiannuzzi: We need to install .NET Core 3.1 SDK to have the latest .NET Core 3.1 _runtime_ available for running the tests. Hence the condition for this step.

.github/workflows/ci.yml
```diff
@@ -3,11 +3,17 @@ name: CI
 on:
   push:
   pull_request:
+  schedule:
+    # Run daily at 00:00 so we get notified if CI is broken before a pull request
+    # is submitted.
+    - cron:  '0 0 * * *'
```
MoFtZ, line 9: QUESTION: How does this interact with the conditional check `if: github.event_name == 'push' || github.event.pull_request.head.repo.id != github.event.pull_request.base.repo.id` ? Do we need another condition?

.github/workflows/ci.yml
MoFtZ: Thanks for the clarification 👍

.github/workflows/ci.yml
ljubon: nice catch! ⚡  I'm not sure anymore about this implementation with mentioned conditions
I'll test conditions separately and provide feedback asap

m4rs-mt:
@jgiannuzzi @ljubon I will include `Co-authored-by: Ljubo Nikolic <ljubonikolic@gmail.com>` when merging this PR 🎉

.github/workflows/ci.yml
jgiannuzzi: Good catch! - I think we need another condition like
```
if: github.event_name == 'push' || github.event_name == 'schedule' || github.event.pull_request.head.repo.id != github.event.pull_request.base.repo.id
```

jgiannuzzi: I have adapted all the conditions to make sure we run all the jobs, with `ILGPU_CLEAN_TESTS=true`, and on macOS too.

ljubon: thanks for covering `schedule` conditions @jgiannuzzi ⚡

.github/workflows/ci.yml
```diff
@@ -88,52 +98,85 @@ jobs:
...
+          dotnet-version: 5.0.x
+      
+      - name: Setup the latest .NET Core 3.1 SDK
+        if: matrix.framework == 'netcoreapp3.1'
+        uses: actions/setup-dotnet@v1
+        with:
+          dotnet-version: 3.1.x
+
+      - name: Set test flavor
+        id: test-flavor
```
ljubon, line 154: maybe i'm missing something here, but I'm interested in this step how/why we are using its outputs? 🤔

jgiannuzzi: This enables us to combine Cuda and CPU tests into one - see https://github.com/m4rs-mt/ILGPU/pull/536/commits/c1d96b938d659f87d876212b735db2d9f21828fb

ljubon: thanks 🙏

.github/workflows/ci.yml
```diff
@@ -212,9 +255,19 @@ jobs:
...
+  all-required-checks-done:
+    needs: [check-style, build, test, package]
+    runs-on: ubuntu-latest
+    steps:
+      - run: echo "All required checks done"
+
+  # Publish a preview version of the NuGet packages on feedz.io when there is
+  # a push to master.
+  publish-preview:
     if: github.ref == 'refs/heads/master' && github.event.repository.fork == 'false'
```
MoFtZ, line 269: QUESTION: I think this `if` condition will cause the scheduled job to publish a preview nuget package every night. Should we disable it for scheduled jobs?

MoFtZ (approved):

.github/workflows/ci.yml
jgiannuzzi: Agreed, thanks for catching it @MoFtZ!

jgiannuzzi: Fixed in https://github.com/m4rs-mt/ILGPU/pull/536/commits/da40b7195110fa6f16604548477ba32555ce0503

MoFtZ (approved):

PR #538, opened 2021-07-14
Merged Samples into ILGPU repository.

m4rs-mt:
This PR merges all ILGPU Samples into the main repository. However, we still need to adapt all sample programs to make them compatible with the `v1.X` version.

@MoFtZ, @MPSQUARK, @Yey007, @faruknane, @NullandKale: This requires additional work on all sample programs 😃


faruknane:
I couldn't find a chance to contribute to ILGPU other than just thinking about ideas. I've got a new job 😃 and I work very much since it's a startup. Therefore, I am truly sorry for not being able to help. However, this repo / ILGPU become important for me because I respect this valuable work. I am considering doing a graduation project and I want to use ILGPU in my project if my professors accept. Thank you again @m4rs-mt !

MoFtZ:
@m4rs-mt Should we revert the change for using ILGPU as a project reference, back to the original Nuget reference? then it will pass the CI builds and we can merge this PR. We can then update the samples later.

Might need to make a change to the Mandelbrot sample to ignore net47 on non-Windows.

Yey007:
@m4rs-mt  I can probably help out. Are there any per-section issues I can check out? Or should I just pick a sample and begin?

m4rs-mt:
In our weekly meeting, we basically agreed to merge the examples as they are (for now). That is, they still point to the official `nuget` releases to be able to compile them properly (again... for now). @MoFtZ suggested that we port the examples one by one.

Yey007:
Sounds good!

PR #541, opened 2021-07-17
Fixed invalid register/variable swaps.

m4rs-mt:
This PR fixes a critical problem in the current `OpenCL` and `PTX` backends. When swapping views inside loops and then accessing them, it could happen that the actual swap operations were not compiled correctly. Although the IR itself represented the operations in a legal way, the backends swapped the registers/variables involved without introducing temporary swap targets. This PR fixes this critical problem by introducing "Phi intermediate values". This concept allows us to reason about Phi values that need to be stored in temporary locations without overwriting their values.

### DIFF
--- a/Src/ILGPU.Tests/BasicLoops.cs
+++ b/Src/ILGPU.Tests/BasicLoops.cs
@@ -1,4 +1,6 @@
 ﻿using ILGPU.Runtime;
+using ILGPU.Util;
+using System;
 using System.Linq;
 using Xunit;
 using Xunit.Abstractions;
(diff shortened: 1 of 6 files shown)


MoFtZ (changes requested):
ISSUE: It looks like the new BasicLoop unit tests are failing on OpenCL.

The generated code fails to compile, complaining about undeclared variables. My guess is that the intermediate phi nodes are not being declared.

m4rs-mt:
@MoFtZ This was my bad as this PR was intended to be an initial draft, since the OpenCL version does not work at the moment 😃

MoFtZ (changes requested):
ISSUE: The new BasicLoop unit tests are failing on OpenCL for O2.

Looks like an existing issue with `GenerateCode(AddressSpaceCast value)`.

m4rs-mt:
@MoFtZ In fact, this patch caused another issue that was already present in the code generation parts related to `AddressSpaceCast` values in the `CLBackend`.

MoFtZ (approved):

PR #546, opened 2021-07-19
Added updated documentation for 0.10.1

NullandKale:
This is the documentation for 0.10.1. I am almost finished with the 1.0.0 docs, I just need to update a small section that I missed in Tutorial_03.md


NullandKale:
In my instance of the docs I kept the ReadMe.md/Home.md file and just renamed it to OldHome.md. In this instance I removed it completely. I did not remove any of the other old docs because some of it covers more advanced topics that my docs do not cover. 


Maybe we should remove any of the old docs that my tutorial covers and leave what is not covered by my docs?

m4rs-mt:
@NullandKale awesome work 🚀 Should we refer to the not-covered "more advanced" docs as "Advanced Resources" in the scope of the documentation?

NullandKale:
@m4rs-mt I made the suggested change and removed the few old docs that were completely covered by my docs. Once I write an explicit grouping tutorial Kernels.md can be removed.

m4rs-mt (approved):

PR #547, opened 2021-07-21
Guard against zero-length CUDA allocations

Joey9801:
From the CUDA API docs for [cuMemAlloc](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__MEM.html#group__CUDA__MEM_1gb82d2a09844a58dd9e744dc31e8aa467):
> If bytesize is 0, cuMemAlloc() returns CUDA_ERROR_INVALID_VALUE.

Since there is already no valid index that can be dereferenced in a zero-length array, I claim that setting `NativePtr` to null is safe to do.


m4rs-mt:
@Joey9801 Nice patch! I think we should also try to keep all runtime implementations "in sync" in terms of available `allocation` features. @MoFtZ @NullandKale @Yey007 You already mentioned specific extensions that might be needed for the OpenCL API, right? I currently think that we need to ensure a minimum buffer "length" of 1 element for OpenCL. Otherwise, the kernel launcher may throw a `NullReferenceException`.

MoFtZ:
To support zero-length allocations, we will need to address some potential issues:
- [ ] Keep the Cuda, OpenCL and CPU implementations consistent.
   - Update the CPU and OpenCL implementations to accept `0` length.
- [ ] Ensure that copy from/to operations deal with zero-length allocations
    - Do the Cuda/OpenCL copy APIs accept `IntPtr.Zero` as a source/destination?
- [ ] Ensure that we can pass zero-length array view to kernel launchers.
    - Do the Cuda/OpenCL launch APIs accept `IntPtr.Zero` as an argument?

Also, if we need to implement zero-length allocations as a 1 element allocation, then we need to ensure that we consistently "lie" whenever someone asks for the length of the buffer. I am unsure if there is any scenario where we would need to tell the "truth" about its length - possibly when talking to the OpenCL launch API?

Joey9801:
@MoFtZ: I've had a check around and attempted to answer all of questions. 
	
- Copy[To|From]CPUUnsafeAsync:
	- From what I can tell, all host <-> device copies eventually end up here
	- These [already](https://github.com/m4rs-mt/ILGPU/blob/master/Src/ILGPU/Runtime/ArrayViewExtensions.cs#L527) explicitly work with zero length views (early successfull return if length is zero).
- CUDA:
	- [API doesn't accept zero length allocations](https://www.khronos.org/registry/OpenCL/sdk/1.2/docs/man/xhtml/clCreateBuffer.html)
	- Can just not call it for zero length allocations, and set the pointer to zero
	- Launch API does accept null pointer / zero length views (confirmed experimentally).
- OpenCL:
	- Like CUDA, doesn't like zero length alloations.
	- Can use the same guard for allocations
	- ?? Launch API accepts null pointer/zero length buffers ??
- CPU buffers:
	- `PointerSourceBuffer`:
		- Trivially ok
	- `UnmanagedMemoryBuffer`:
		- `Marshal.AllocHGlobal` is happy to allocate zero bytes
	- `PageLockedMemoryBuffer`:
		- Inherits from `UnmanagedMemoryBuffer`, so the allocation is OK
		- CUDA:
			- `cuMemHostRegister` complains about both null pointers and zero length (experimentally found, docs say nothing)
			- Easy to add guard, and just return `NullPageLockScope` for such buffers
		- OpenCL:
			- Page locked memory not supported (at least by ILGPU)
	- ArraySourceBuffer:
		- `GCHandle.Alloc` is happy to accept a zero length array

I think that the only unaswered point is whether OpenCL is happy to supply null pointers / zero length views to its kernels. I don't have an OpenCL environent set up on my machine, so haven't been able to test that myself.

MoFtZ:
Thanks @Joey9801. Could you please add some unit tests for the various scenarios?

I'll do some investigation into the unanswered OpenCL questions.

m4rs-mt:
@Joey9801 Thanks for the summary of the current state of support for zero-length allocations. @MoFtZ @Joey9801 I have been playing around with the OpenCL API and apparently there are *huge* differences between the various hardware vendors. The only actual "solution" I have found so far is to allocate at least a single element to create an actual buffer handle that can be passed to the driver.

MoFtZ:
@m4rs-mt That is disappointing, as the [clSetKernelArg](https://www.khronos.org/registry/OpenCL/sdk/2.0/docs/man/xhtml/clSetKernelArg.html) documentation seemed to suggest that it was possible to use `NULL` as a buffer argument.

Are we able to apply this single element buffer at the last minute? e.g. Allocate it just to pass to the kernel, and then dispose immediately afterwards?

m4rs-mt:
@MoFtZ This seems to work perfectly fine when specifying the size of a dynamically-allocated shared-memory array (we are already doing that [here](https://github.com/m4rs-mt/ILGPU/blob/master/Src/ILGPU/Runtime/OpenCL/CLAPI.cs#L78)).

Src/ILGPU/Runtime/OpenCL/CLMemoryBuffer.cs
```diff
@@ -145,14 +145,21 @@ public sealed class CLMemoryBuffer : MemoryBuffer
...
-                CurrentAPI.CreateBuffer(
-                    accelerator.NativePtr,
-                    CLBufferFlags.CL_MEM_READ_WRITE,
-                    new IntPtr(LengthInBytes),
-                    IntPtr.Zero,
-                    out IntPtr resultPtr));
-            NativePtr = resultPtr;
+            if (LengthInBytes == 0)
+            {
+                NativePtr = IntPtr.Zero;
```
m4rs-mt, line 150: I have experimented with a lot (and I really mean it 😄) hardware configurations using our OpenCL runtime. It turns out that on some ARM and AMD platforms, the `ReleaseBuffer` operation fails if its being called with a `null` pointer. If we add some extra checks to the `DisposeAcceleratorObject` method, this patch should work fine in practice with the latest drivers on all platforms.

m4rs-mt:
@MoFtZ Regarding `clSetKernelArg`: It actually seems to work to use `null` pointers in practice. The issue was related to a different problem affecting my AMD driver setup 😞

m4rs-mt:
@Joey9801 @MoFtZ minor: should we also add a trivial test case for testing zero-length allocations?

MoFtZ:
> @Joey9801 @MoFtZ minor: should we also add a trivial test case for testing zero-length allocations?

I am thinking tests for:
- zero allocation
- zero copy from/to
- launch kernel with zero-length buffer

m4rs-mt:
@MoFtZ @Joey9801 I have created a new PR containing your changes @Joey9801 + additional guards for the OpenCL `MemoryBuffer` class.

m4rs-mt:
I had a brief discussion with @Joey9801 and he is fine with tracking his changes under the updated PR #610.

PR #549, opened 2021-07-22
Fixed constant folding when converting unsigned integers.

### LINKED ISSUE #544: Switch . . . Case not matching integers correctly.
I have a simple loop with a case statement in it.  The case statements aren't matching with the variables in the loop which is causing some wildly different behavior than executing the same view CPU.

```c#
for(int i=0; i< 24;i++) {
                switch (i)
                {
                    case 0:
                        cheatLong = 0x0000000000000001;
                        break;
                    case 1:
                        cheatLong = 0x0000000000008082;
                        break;
.
.
.
                    case 23:
                        cheatLong = 0x8000000080008008;
                        break;
                    default:
                        cheatLong = 0x0000000000000001;
                        break;
                }
}
```

In this case you would expect that the default statement would never be hit, but I am finding that it quite often does.

MoFtZ:
Fixes #544.

The following new unit tests fail with the existing code:
- `ConvertOperation_UInt32_Int64_MaxValue`
- `ConvertOperation_UInt32_UInt64_MaxValue`

From the reproduction steps in #544, the C# compiler recognizes `0x0000000080000001` as `0x80000001`, which fits as UInt32. It then generates a cast from UInt32 to UInt64.

The issue occurs during generation of the IR for this constant value. The cast from UInt32 to UInt64 is incorrectly folded, and produces the constant `0xffffffff80000001`.

### DIFF
--- a/Src/ILGPU.Tests/ConvertIntOperations.Generated.tt
+++ b/Src/ILGPU.Tests/ConvertIntOperations.Generated.tt
@@ -54,6 +54,33 @@ namespace ILGPU.Tests
             Verify(b.View, reference);
         }
 
+        internal static void <#= kernelName #>_MaxValue(
+            Index1D index,
+            ArrayView1D<<#= targetType.Type #>, Stride1D.Dense> output)
+        {
+            unchecked
+            {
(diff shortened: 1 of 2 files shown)


m4rs-mt (approved):
Nice catch 👍

PR #550, opened 2021-07-23
Fix zero-length subview behaviour

Joey9801:
Currently if you attempt to create a subview that has zero length in at least one non-dense dimension, an exception gets thrown complaining about an index being out of range. It turns out that this comes from `ComputeBufferLength` returning a negative number.

The existing calculation in `ComputeBufferLength` is something like `stride.ComputeElementIndex(extent - IndexND.One) + 1`, ie it computes a value one past the end of the last element. When the extent is zero in at least one dimension however this behaves oddly, as the subtraction is non-saturating:
- If the only zero is in a dense dimension, it will return a positive non-zero number (not dangerously wrong, but more than required)
- If there is a zero in a non-dense dimension, it will return a negative number.  This is clearly wrong, and gets caught shortly afterwards when trying to create a subview of the BaseView.

This PR:
- Adds unit tests demonstrating the issue
- Adds some guards to the various `ComputeBufferLength` methods to fix this behaviour.
- Normalizes the implementation of `Stride2D.DenseX` such that it defers to the templated `ComputeBufferLength` methods like the other `Stride[2, 3]D.DenseAB` structs.

### DIFF
--- a/Src/ILGPU.Tests/ArrayViews.cs
+++ b/Src/ILGPU.Tests/ArrayViews.cs
@@ -249,7 +249,8 @@ namespace ILGPU.Tests
         [Fact]
         public void ArrayViewZeroLengthSubView2D()
         {
-            void Check<TStride>(ArrayView2D<int, TStride> view) where TStride : struct, IStride2D
+            void Check<TStride>(ArrayView2D<int, TStride> view)
+                where TStride : struct, IStride2D
             {
                 var extents = new[]
                 {
@@ -278,7 +279,8 @@ namespace ILGPU.Tests
         [Fact]
         public void ArrayViewZeroLengthSubView3D()
         {
-            void Check<TStride>(ArrayView3D<int, TStride> view) where TStride : struct, IStride3D
+            void Check<TStride>(ArrayView3D<int, TStride> view)
+                where TStride : struct, IStride3D
             {
                 var extents = new[]
                 {
--- a/Src/ILGPU.Tests/ArrayViews.cs
+++ b/Src/ILGPU.Tests/ArrayViews.cs
@@ -231,7 +231,7 @@ namespace ILGPU.Tests
         }
 
         [Fact]
-        public void ArrayViewZeroLengthSubview1d()
+        public void ArrayViewZeroLengthSubView1D()
         {
             using var buffer = Accelerator.Allocate1D<int>(128);
 
@@ -247,7 +247,7 @@ namespace ILGPU.Tests
         }
 
         [Fact]
-        public void ArrayViewZeroLengthSubview2d()
+        public void ArrayViewZeroLengthSubView2D()
         {
             void Check<TStride>(ArrayView2D<int, TStride> view) where TStride : struct, IStride2D
             {
@@ -276,7 +276,7 @@ namespace ILGPU.Tests
         }
 
         [Fact]
-        public void ArrayViewZeroLengthSubview3d()
+        public void ArrayViewZeroLengthSubView3D()
         {
             void Check<TStride>(ArrayView3D<int, TStride> view) where TStride : struct, IStride3D
             {
--- a/Src/ILGPU.Tests/ArrayViews.cs
+++ b/Src/ILGPU.Tests/ArrayViews.cs
@@ -234,9 +234,16 @@ namespace ILGPU.Tests
         public void ArrayViewZeroLengthSubview1d()
         {
             using var buffer = Accelerator.Allocate1D<int>(128);
-            var zero = buffer.View.SubView(64, 0);
-            Assert.Equal(0, zero.Length);
-            Assert.Equal(0, zero.LengthInBytes);
+
+            var subView1 = buffer.View.SubView(64, 0);
+            Assert.Equal(0, subView1.Length);
+            Assert.Equal(0, subView1.LengthInBytes);
+            Assert.Equal(0, subView1.Extent.X);
+
+            var subView2 = buffer.View.AsGeneral().SubView(64, 0);
+            Assert.Equal(0, subView2.Length);
+            Assert.Equal(0, subView2.LengthInBytes);
+            Assert.Equal(0, subView2.Extent.X);
         }
 
         [Fact]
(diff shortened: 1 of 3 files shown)


Src/ILGPU/StrideTypes.tt
```diff
@@ -298,7 +298,7 @@ namespace ILGPU
             /// Computes the 32-bit length of a required allocation.
             /// </summary>
             /// <param name="extent">The extent to allocate.</param>
-            /// <returns>The 3264-bit length of a required allocation.</returns>
```
m4rs-mt, line 301: Nice catch 🤓

Src/ILGPU/StrideTypes.tt
```diff
@@ -378,13 +378,20 @@ namespace ILGPU
...
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public static int ComputeBufferLength<TStride>(
             this TStride stride,
             <#= indexName #> extent)
-            where TStride : <#= strideName #> =>
-            stride.ComputeElementIndex(extent - <#= indexName #>.One) + 1;
+            where TStride : <#= strideName #> {
+<#      for (int i = 0; i < def.Dimension; ++i) { #>
+<#          var iName = IndexDimensions[i].PropertyName; #>
+            if (extent.<#= iName #> == 0) {
```
m4rs-mt, line 389: nit: Please removed the braces around the if statement.

Src/ILGPU/StrideTypes.tt
```diff
@@ -396,8 +403,16 @@ namespace ILGPU
         public static long ComputeBufferLength<TStride>(
             this TStride stride,
             <#= longIndexName #> extent)
-            where TStride : <#= strideName #> =>
-            stride.ComputeElementIndex(extent - <#= longIndexName #>.One) + 1L;
+            where TStride : <#= strideName #> {
+<#      for (int i = 0; i < def.Dimension; ++i) { #>
+<#          var iName = IndexDimensions[i].PropertyName; #>
+            if (extent.<#= iName #> == 0L) {
```
m4rs-mt, line 409: nit: same here 🤓

m4rs-mt (changes requested):
Nice patch 👍 Yes, you are right that support for zero-length views has not been implemented properly. I think that this PR is a great extension to the current (sub)view API.

Src/ILGPU.Tests/ArrayViews.cs
```diff
@@ -230,6 +230,82 @@ public void ArrayViewLengthInBytes(int length)
             Verify(buffer.View, expected);
         }
 
+        [Fact]
+        public void ArrayViewZeroLengthSubview1d()
```
MoFtZ, line 234: MINOR: Follow existing naming convention - `SubView` and `1D`.
```suggestion
        public void ArrayViewZeroLengthSubView1D()
```

Src/ILGPU.Tests/ArrayViews.cs
```diff
@@ -230,6 +230,82 @@ public void ArrayViewLengthInBytes(int length)
...
+            Assert.Equal(0, subView1.Extent.X);
+
+            var subView2 = buffer.View.AsGeneral().SubView(64, 0);
+            Assert.Equal(0, subView2.Length);
+            Assert.Equal(0, subView2.LengthInBytes);
+            Assert.Equal(0, subView2.Extent.X);
+        }
+
+        [Fact]
+        public void ArrayViewZeroLengthSubview2d()
```
MoFtZ, line 250:
```suggestion
        public void ArrayViewZeroLengthSubView2D()
```

Src/ILGPU.Tests/ArrayViews.cs
```diff
@@ -230,6 +230,82 @@ public void ArrayViewLengthInBytes(int length)
...
+            using var buff1 = Accelerator.Allocate2DDenseY<int>((10, 10));
+            using var buff2 = Accelerator.Allocate2DDenseX<int>((10, 10));
+            Check<Stride2D.DenseY>(buff1.View);
+            Check<Stride2D.General>(buff1.View.AsGeneral());
+            Check<Stride2D.DenseX>(buff2.View);
+            Check<Stride2D.General>(buff2.View.AsGeneral());
+        }
+
+        [Fact]
+        public void ArrayViewZeroLengthSubview3d()
```
MoFtZ, line 279:
```suggestion
        public void ArrayViewZeroLengthSubView3D()
```

MoFtZ (commented):
QUESTION: Should we also deal with negative input values?

Joey9801:
> QUESTION: Should we also deal with negative input values?

Aftter thinking about it for a bit, there are a few things that could be done for negative subview extents:
1. Alter the subview API to take unsigned extents, such that negative numbers are unrepresentable in the first place.
2. Throw runtime exceptions if there is a negative extent component
3. Actually try and support them in some meaningful way

There are a few interpretations of what 3. could mean:
- A negative extent component just shifts the start of the view backwards. Eg `[0, 1, 2, 3, 4, 5].SubView(3, -2) == [1, 2]`
- A negative extent component actually reverses the view in that axis. Eg `[0, 1, 2, 3, 4, 5].SubView(3, -2) == [2, 1]`. Could be trivially implemented with negative strides.

I think that of those options I would most favour 1., especially as v1.0 is still in beta so the public API is still mutable to a certain extent. To the best of my knowledge nobody is asking for any features that look like 3. Additionally it looks like implementing it could be fraught with odd edge cases and off-by-one errors, so it's by far my least favourite option.

MoFtZ:
> 1. Alter the subview API to take unsigned extents, such that negative numbers are unrepresentable in the first place.

This will be for @m4rs-mt to decide. Most of the APIs use signed integers, I'm guessing because it's CLS-compliant (?)

> 2. Throw runtime exceptions if there is a negative extent component

ILGPU does not currently support exceptions within the kernel. However, we can assert, which gets mapped to the accelerator-specific assert.

> 3. Actually try and support them in some meaningful way

I was thinking more of option 2 when I raised the question, but if we can come up with a meaningful API, then great. And yes, a separate PR sounds good to me.

MoFtZ (approved):

m4rs-mt:
> Most of the APIs use signed integers, I'm guessing because it's CLS-compliant (?)

@MoFtZ @Joey9801 Yes, you are right. This is to be CLS-compliant. I would prefer adding assertion checks to the code as @MoFtZ suggested.

m4rs-mt (approved):

PR #559, opened 2021-07-30
Fixed naming of available static field modes.

MoFtZ:
This looks like a misnamed enum from the change to use context builders.


m4rs-mt:
@MoFtZ Looks good to me 👍

m4rs-mt (approved):

PR #566, opened 2021-07-31
Added Cuda ISA 7.3 and 7.4.

MoFtZ:


m4rs-mt:
@MoFtZ do we need to adapt some instructions in order to make them compatible with `ISA >= 7.3`?

MoFtZ:
Not that I could see from the release notes:
https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#changes-in-ptx-isa-version-7-4

If we wanted to be cautious, we could skip updating `SupportedInstructionSets`. What do you think?

m4rs-mt (approved):
@MoFtZ looks good to me 👍.

PR #569, opened 2021-08-02
Added UCE transformation to the backend optimization passes.

### LINKED ISSUE #567: Regression when allocating 2D/3D SharedMemory inside kernel since ILGPU v1.0.0-beta1
When compiling a kernel using `SharedMemory` to allocate 2D or 3D memory, ILGPU is throwing a `NotSupportedException` with the message `The allocation size of type 'Int32' must be statically known and not a dynamic value`.

The `SharedMemory.Allocate2D` and `SharedMemory.Allocated3D` functions were changed to accept `Index2D/Index3D` in the new Memory API, which appears to be causing an issue with the expectation of a constant value.

```CSharp
public class Program
{
    public const int DIM_X = 3;
    public const int DIM_Y = 5;

    public static void MyKernel(ArrayView<int> buffer)
    {
        var sharedMemory = SharedMemory.Allocate2D<int, Stride2D.DenseX>(new Index2D(DIM_X, DIM_Y), new Stride2D.DenseX(DIM_Y));
    }

    static unsafe void Main()
    {

        var kernel = accelerator.LoadStreamKernel<ArrayView<int>>(MyKernel);
    }
}
```

m4rs-mt:
This PR adds our `UnreachableCodeElimination` pass to the backend optimization passes. Due to several refinement steps in our transformation pipeline (compared to `v0.10.X` and earlier versions) this could cause problems (part of #567) and can easily lead to suboptimal program code being generated.

### DIFF
--- a/Src/ILGPU/IR/Transformations/Optimizer.cs
+++ b/Src/ILGPU/IR/Transformations/Optimizer.cs
@@ -151,6 +151,11 @@ namespace ILGPU.IR.Transformations
             if (inliningMode != InliningMode.Disabled)
                 builder.Add(new Inliner());
 
+            // Apply UCE and DCE passes to avoid dead branches and fold conditionals that
+            // do not affect the actual code being executed
+            builder.Add(new UnreachableCodeElimination());
+            builder.Add(new DeadCodeElimination());
+
             // Skip further optimizations in debug mode
             if (level < OptimizationLevel.O1)
                 return;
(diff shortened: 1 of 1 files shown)


MoFtZ (changes requested):
ISSUE: The `MatrixMultiply` sample is still failing at run-time.

m4rs-mt:
@MoFtZ It turns out that the `CPUBackend` does not perform backend specific transformations as they are not usually not required (at the moment...). However, this causes a runtime check related to the size of an allocation node to fail.

MoFtZ (changes requested):
Can we please add a unit test for `SharedMemory.Allocate2D` to ensure that future changes to the optimizer does not break.

m4rs-mt:
@MoFtZ I think we should add this test to #574, otherwise the CPU tests will fail.

MoFtZ:
If we add the unit tests to #574, wont the Cuda unit tests fail without this PR?

MoFtZ (approved):

m4rs-mt:
@MoFtZ This means that we can merge #569 and #574 now and you will open  another PR that adds those test cases?

PR #574, opened 2021-08-02
Enabled Backend optimizations for CPU backend.

### LINKED ISSUE #567: Regression when allocating 2D/3D SharedMemory inside kernel since ILGPU v1.0.0-beta1

m4rs-mt:
Based on #569 and #567, we decided to enable all backend-specific operations in CPU mode as well to ensure the same "experience" and behavior of ILGPU in CPU and GPU accelerated modes.

Fixes #567 when both, #569 and this PR are merged.

### DIFF
--- a/Src/ILGPU.Tests/SharedMemory.cs
+++ b/Src/ILGPU.Tests/SharedMemory.cs
@@ -210,5 +210,53 @@ namespace ILGPU.Tests
             Assert.Throws<InternalCompilerException>(() =>
                 Execute(buffer.Length, buffer.View));
         }
+
+        internal static void MultiDimensionalSharedMemoryKernel2D(
+            ArrayView1D<int, Stride1D.Dense> output)
+        {
+            var sharedMemory = ILGPU.SharedMemory.Allocate2D<int, Stride2D.DenseX>(
+                new Index2D(5, 7),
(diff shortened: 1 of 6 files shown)


MoFtZ (approved):

m4rs-mt:
As discussed in #569, this PR requires additional test cases to represent the issue(s) from #567.

PR #570, opened 2021-08-02
Updated ILGPU samples CI pipeline.

MoFtZ:
Part of #562.

Updated ILGPU samples:
- Fixed warnings.
- Updated to C# 8.
- Simplified `using` statements.

Updated CI pipeline:
- Build ILGPU samples.
- Package samples into ZIP artifact
  - Replaced ILGPU project references with package references.

PR #575, opened 2021-08-03
Fix release creation in CI pipeline

jgiannuzzi:
This PR aims at fixing a bug I introduced during the CI refactor in #536 😓

The output of the `package` job was not accessible anymore to the `publish-release` job, which resulted in the wrong Release name, and the `prerelease` flag never being set.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -253,7 +253,7 @@ jobs:
           path: ./Bin/Release/ILGPU*.${{ steps.package.outputs.version }}.*nupkg
 
     outputs:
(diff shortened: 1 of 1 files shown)


MoFtZ:
@jgiannuzzi I fixed this in #570, but wanted to confirm for this PR - the `publish-release` job uses `needs.package.outputs.version`, but the output of the `package` job is `tag`. Should that `tag` be changed to `version`?

jgiannuzzi:
Correct, thanks @MoFtZ! I've updated this PR accordingly.

I will try to review #570 as soon as I can (might not be today though).

MoFtZ (approved):

PR #585, opened 2021-08-05
Allow zero-length subviews of zero-length views

Joey9801:
#550 made it possible to make a zero-length subview of a non-zero-length view. It turns out that it was still not possible to make a zero-length subview of a zero-length subview.

For example, prior to this PR the following does not work:
```c#
var a = accelerator.Allocate2DDenseY((10, 10));
var b = a.View.SubView(index: (0, 0), extent: (5, 0));
var c = b.View.SubView(index: (0, 0), extent: (3, 0)); // <-- Fails
```

The only problem here was pair of slightly too conservative `Trace.Assert`s. This PR extends the unit tests to cover this case, and amends those asserts.

### DIFF
--- a/Src/ILGPU/Runtime/ArrayViews.tt
+++ b/Src/ILGPU/Runtime/ArrayViews.tt
@@ -248,7 +248,9 @@ namespace ILGPU.Runtime
 <#      for (int i = 0; i < dim.Dimension; ++i) { #>
 <#          var iName = IndexDimensions[i].PropertyName; #>
             Trace.Assert(
-                index.<#= iName #> >= 0 & index.<#= iName #> < Extent.<#= iName #>,
+                index.<#= iName #> >= 0 & (
+                    index.<#= iName #> < Extent.<#= iName #>
+                    || index.<#= iName #> == 0),
                 "Index <#= iName #> out of bounds");
 <#      } #>
             return Stride.ComputeElementIndex(index);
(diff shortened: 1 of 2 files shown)


Src/ILGPU/Runtime/ArrayViews.tt
```diff
@@ -248,7 +248,9 @@ namespace ILGPU.Runtime
 <#      for (int i = 0; i < dim.Dimension; ++i) { #>
 <#          var iName = IndexDimensions[i].PropertyName; #>
             Trace.Assert(
-                index.<#= iName #> >= 0 & index.<#= iName #> < Extent.<#= iName #>,
+                index.<#= iName #> >= 0 & (
+                    index.<#= iName #> < Extent.<#= iName #>
+                    || index.<#= iName #> == 0),
```
m4rs-mt, line 253: nit: can we change this to a logic `|` operation in order to improve compilation speed in debug builds?

Src/ILGPU/Runtime/ArrayViews.tt
```diff
@@ -268,10 +270,11 @@ namespace ILGPU.Runtime
         {
 <#      for (int i = 0; i < dim.Dimension; ++i) { #>
 <#          var iName = IndexDimensions[i].PropertyName; #>
+<#          var thisExtent = $"({baseType})Extent.{iName}"; #>
             Trace.Assert(
-                index.<#= iName #> >= 0 &
-                index.<#= iName #> + extent.<#= iName #> <=
-                    (<#= baseType #>)Extent.<#= iName #>,
+                index.<#= iName #> >= 0 &&
```
m4rs-mt, line 275: nit: can we change this to a logic `|` operation in order to improve compilation speed in debug builds?

Src/ILGPU/Runtime/ArrayViews.tt
```diff
@@ -268,10 +270,11 @@ namespace ILGPU.Runtime
...
+                (index.<#= iName #> + extent.<#= iName #> <= <#= thisExtent #>
+                 || (index.<#= iName #> == 0 && extent.<#= iName #> == 0)),
```
m4rs-mt, line 277: nit: can we change this to a logic `|` operation in order to improve compilation speed in debug builds?

m4rs-mt (changes requested):
Thanks a lot @Joey9801 for these very useful extensions. This makes the "zero-length subview support" even "more complete" 🚀

Joey9801:
I've made the requested changes.

I wasn't previously aware of a compilation speed difference between the logical operators (`&`, `|`, etc...) and the conditional logical operators (`&&`, `||`, etc..). Is this the case for general C# programs, or are you referring to the ILGPU JIT performance?

m4rs-mt:
@Joey9801 Yes, I was referring to the ILGPU JIT performance. `&&` and `||` operators involve the generation of multiple branches and basic blocks that increase the runtime of the compilation pipeline in debug builds 🤓

m4rs-mt (approved):

Joey9801:
That makes sense, thanks!

Issue #597, opened 2021-08-17
Arrays in structs?


madsami:
I need to have structs with the layout e.g.,
```
MyStruct {
  public double[] data;
}
```

But when I try to declare it like so: _public ArrayView<MyStruct> population_ I get the error **'MyStruct' must be a non-nullable value type, along with all fields at any level of nesting, in order to use it as parameter 'T' in the generic type or method 'ArrayView<T>'.**

I tried replacing the double[] with ArrayView<double>, but it's the same. How can I realize a design like this?
The size of the array is set upon creation. Basically what I really need is the ability to have a list/array of objects that also contain further lists or arrays.

MoFtZ:
hi @madsami. As you noticed, ILGPU only supports "value types". Arrays in .Net are considered a reference type, and the struct just holds a reference to another object (the array).

If your data is a fixed length, you could try using [fixed sized buffers](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/unsafe-code#fixed-size-buffers), which are supported in ILGPU:
```CSharp
unsafe struct MyStruct
{
    public fixed double data[128];
}
```
The above defines a struct with 128 elements. As a fixed sized buffer, the struct itself holds all 128 elements, rather than a reference to another object/array.

Otherwise, you will need to arrange your data differently. For example, you could try and flatten your data. One `ArrayView<double>` containing all the `data` elements, and another `ArrayView<int>` than contains indices into `data`.

PR #600, opened 2021-08-19
Added additional sample projects.

MoFtZ:
Added fixed size buffer example for #597.

QUESTION: Can you think of any other sample projects we should add?

### DIFF
--- a/Src/ILGPU/Runtime/ArrayViewExtensions.cs
+++ b/Src/ILGPU/Runtime/ArrayViewExtensions.cs
@@ -1209,6 +1209,40 @@ namespace ILGPU.Runtime
             }
         }
 
+        /// <summary>
+        /// Copies from the source view into the given CPU data array using the default
+        /// stream.
+        /// </summary>
+        /// <typeparam name="T">The element type.</typeparam>
+        /// <typeparam name="TView">The view type.</typeparam>
+        /// <param name="source">The source view instance.</param>
+        /// <param name="span">The CPU data target.</param>
+        /// <remarks>This method is not supported on accelerators.</remarks>
+        [NotInsideKernel]
+        public static unsafe void CopyToCPU<T, TView>(
+            this TView source,
+            in Span<T> span)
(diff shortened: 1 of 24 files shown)


m4rs-mt (approved):

m4rs-mt:
@MoFtZ Maybe we can add an additional sample demonstrating the use of our `Random` functionality:

- Using the high-level `Random` API to fill a buffer with uniformly distributed floats (can also include the `CuRand` features)
- Using the kernel `Random` API while accessing a data view inside a GPU kernel to generate random numbers on the fly

MoFtZ:
@m4rs-mt I'm trying to use the `RNGView` within a kernel, but it raises an exception on the CPU accelerator inside `GetRandomProvider`. Did you have any suggestions for how to fix the issue?

MoFtZ:
Fixed issue with `RNGView` on the CPU accelerator. On the CPU accelerator, the group size was not always the same as the warp size. And even on Cuda accelerators, my group size was 30, compared to a warp size of 32, so it also had potential issues.

Also updated the `SimpleViews` sample to show offset copying to fix https://github.com/m4rs-mt/ILGPU/issues/604.

Src/ILGPU/Runtime/ArrayViewExtensions.cs
```diff
@@ -1209,6 +1209,40 @@ public static void MemSetToZero<TView>(this TView view)
...
+        /// <summary>
+        /// Copies from the source view into the given CPU data array using the default
+        /// stream.
+        /// </summary>
+        /// <typeparam name="T">The element type.</typeparam>
+        /// <typeparam name="TView">The view type.</typeparam>
+        /// <param name="source">The source view instance.</param>
+        /// <param name="span">The CPU data target.</param>
+        /// <remarks>This method is not supported on accelerators.</remarks>
+        [NotInsideKernel]
```
m4rs-mt, line 1221: nice extension methods 👍

m4rs-mt (approved):

PR #607, opened 2021-09-01
Fixed invalid detection and processing of loops consisting of a single entry block.

m4rs-mt:
This PR fixes invalid detection (and invalid processing) of loops consisting of a single block (which is also the entry block of a function that has no predecessors).

### DIFF
--- a/Src/ILGPU.Tests/BasicLoops.cs
+++ b/Src/ILGPU.Tests/BasicLoops.cs
@@ -611,6 +611,63 @@ namespace ILGPU.Tests
                 (expected2[0], expected2[1]), Length).ToArray();
             Verify(target2.View, expectedData2);
         }
+
+        /// <summary>
+        /// Wrapper view required by <see cref="DoLoopWithoutEntryBlockKernel(Index1D,
+        /// ArrayView{int}, ArrayView{int})"/>.
+        /// </summary>
+        private readonly struct WrapperView
+        {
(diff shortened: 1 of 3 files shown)


MoFtZ (changes requested):
ISSUE: This appears to produce a kernel that never returns.

When I edit the `Empty` sample project with `<Optimize>true</Optimize>`, and run the following code in Debug configuration:

```CSharp
static void Main()
{
    // Every application needs an instantiated global ILGPU context
    using var context = Context.Create(x => x.Cuda());

    {
        var length = 7;
        Console.WriteLine($"Performing operations on {accelerator}");

        using var target = accelerator.Allocate1D<int>(length);
        using var source = accelerator.Allocate1D<int>(length);
        source.MemSetToZero();
        accelerator.Synchronize();

        var kernel = accelerator.LoadAutoGroupedStreamKernel<Index1D, ArrayView<int>, ArrayView<int>>(DoLoopWithoutEntryBlockKernel);
        kernel(length, target.View.AsContiguous(), source.View.AsContiguous());

        var expected = Enumerable.Repeat(0, length).ToArray();
        accelerator.Synchronize();
    }
}
```

m4rs-mt:
It was not sufficient to update the loop analysis. Instead, it was required to fix an issue in the IL frontend code generator to wire loops consisting of a single block (which is *also* the entry block of a function).

MoFtZ (approved):

PR #611, opened 2021-09-08
SPIR-V Builder (1/6 SPIR-V Backend PRs)

Yey007:
This is the first of six planned PRs for the current state of the SPIR-V backend. It adds the two SPIR-V builder classes, `StringSPIRVBuilder` and `BinarySPIRVBuilder`, which provide an interface to generate SPIR-V in string representation and runnable binary respectively.

Note: I added tests for the builder utilities for my own convenience, if they should be removed I can do that.

Question: I added JetBrains Rider's cache folder to the .gitignore, as I personally use the education version. Should I remove it? It would be a little annoying to delete the cache each time I want to commit, but I understand that it's not an IDE that many use.

Question: What should we do about the massive SPIRV grammar file? I assume we don't want to just keep that laying around.


Yey007:
Looks like I have a few tests to fix...

Yey007:
Hey, is it alright to ignore some warnings for generated code? For example enum members having the same value? Or CLS compliance? I could fix them but considering the code is generated it doesn't seem very worthwhile, except maybe the warnings about conflicting language keywords.

Src/ILGPU/Backends/SPIRV/BinarySPIRVBuilder.tt
```diff
@@ -0,0 +1,131 @@
...
+<#
+    PushIndent(standardIndent);
+#>
+/// <summary>
+/// Defines utility methods to generate SPIRV operations
+/// </summary>
+[CLSCompliant(false)]
+public class BinarySPIRVBuilder : ISPIRVBuilder {
+
+    private List<uint> _instructions = new List<uint>();
```
m4rs-mt, line 72: nit: make field readonly.

Src/ILGPU.Tests.SPIRV/SPIRVBuilderUtilsTests.cs
```diff
@@ -0,0 +1,64 @@
...
+using Xunit;
+
+namespace ILGPU.Tests.SPIRV
+{
+    public class SPIRVBuilderUtilsTests
+    {
+        [Fact]
+        public void ToUintListUint()
+        {
+            var result = SPIRVBuilderUtils.ToUintList((uint)0);
```
m4rs-mt, line 12: General question: should we abstract from the actual `uint` type by introducing a new `SpirVInstruction` structure? This would make the code easier to understand and all types will be `CLSCompliant`.

Src/ILGPU/Backends/SPIRV/BinarySPIRVBuilder.tt
```diff
@@ -0,0 +1,131 @@
...
+
+    public byte[] ToByteArray() {
```
m4rs-mt, line 74: Performance question: should be change to interface to return a `ReadOnlySpan<byte>`?

Src/ILGPU/Backends/SPIRV/BinarySPIRVBuilder.tt
```diff
@@ -0,0 +1,131 @@
...
+namespace ILGPU.Backends.SPIRV {
+
+<#
+    PushIndent(standardIndent);
+#>
+/// <summary>
+/// Defines utility methods to generate SPIRV operations
+/// </summary>
+[CLSCompliant(false)]
+public class BinarySPIRVBuilder : ISPIRVBuilder {
```
m4rs-mt, line 70: nit: please adapt the coding convention regarding the braces

Src/ILGPU/Backends/SPIRV/BinarySPIRVBuilder.tt
```diff
@@ -0,0 +1,131 @@
...
+
+    public byte[] ToByteArray() {
+        uint[] uintArray = _instructions.ToArray();
+        byte[] byteArray = new byte[uintArray.Length * 4];
+        Buffer.BlockCopy(uintArray, 0, byteArray, 0, uintArray.Length * 4);
+        return byteArray;
+    }
+
+    public void AddMetadata(
+        uint magic,
```
m4rs-mt, line 82: remark: if we introduce a `SpirVInstruction`, this function will have nicer signature. I also think about introducing enumeration/static field values for version and genMagic parameters. What do you think?

Src/ILGPU/Backends/SPIRV/BinarySPIRVBuilder.tt
```diff
@@ -0,0 +1,131 @@
...
+        uint schema)
+    {
+        _instructions.Add(magic);
+        _instructions.Add(version);
+        _instructions.Add(genMagic);
+        _instructions.Add(bound);
+        _instructions.Add(schema);
+    }
+
+    public void Merge(ISPIRVBuilder other) {
```
m4rs-mt, line 95: nit: adapt coding convention regarding braces

Src/ILGPU/Backends/SPIRV/BinarySPIRVBuilder.tt
```diff
@@ -0,0 +1,131 @@
...
+        if(other == null) {
```
m4rs-mt, line 96: nit: no braces needed

Src/ILGPU/Backends/SPIRV/BinarySPIRVBuilder.tt
```diff
@@ -0,0 +1,131 @@
...
+            throw new ArgumentNullException();
+        }
+
+        var otherBinary = other as BinarySPIRVBuilder;
```
m4rs-mt, line 100: nit: change to other is BinarySPIRVBuilder to avoid sophisticated checks

Src/ILGPU/Backends/SPIRV/BuilderCommon.ttinclude
```diff
@@ -0,0 +1,89 @@
...
+    var typeMap = LoadTypeMap();
+    const string standardIndent = "    ";
+
+    bool IsInstructionCLSCompliant(SPIRVOp inst)
+    {
+        var list = inst.Operands?.Operands;
+        if (list != null)
+        {
+            foreach (var operand in list)
+            {
```
m4rs-mt, line 18: nit: no braces needed

Src/ILGPU/Backends/SPIRV/BuilderCommon.ttinclude
```diff
@@ -0,0 +1,89 @@
...
+                if (!IsCLSCompliant(operand.Type))
+                {
```
m4rs-mt, line 20: nit: no braces needed

Src/ILGPU/Backends/SPIRV/BuilderCommon.ttinclude
```diff
@@ -0,0 +1,89 @@
...
+        }
+        return true;
+    }
+
+    string GenerateParamName(SPIRVOperand operand, int fallbackIndex)
+    {
+        string paramName = operand.Name ?? $"param{fallbackIndex}";
+
+        if (operand.Type == "IdResult")
+        {
```
m4rs-mt, line 34: nit: code formatting

Src/ILGPU/Backends/SPIRV/BuilderCommon.ttinclude
```diff
@@ -0,0 +1,89 @@
...
+
+        return paramName;
+    }
+
+    string GenerateFullParam(SPIRVOperand operand, string paramName)
+    {
+        var type = typeMap[operand.Type];
+        var fullParam = $"{type} {paramName}";
+        if (operand.Quantifier == "*")
+        {
```
m4rs-mt, line 49: nit: no braces needed

Src/ILGPU/Backends/SPIRV/BuilderCommon.ttinclude
```diff
@@ -0,0 +1,89 @@
...
+
+        return pairs;
+    }
+
+    void CreateHeader(string opName, List<(SPIRVOperand operand, string name)> paramInfos)
+    {
+#>public void Generate<#= opName #>(<#
+
+        var names = paramInfos.Select(x => GenerateFullParam(x.operand, x.name));
+        string joined = string.Join(", ", names);
```
m4rs-mt, line 84: nit: can be simplified to `<# string.Join(", ", names) #>`

Src/ILGPU/Backends/SPIRV/ISPIRVBuilder.tt
```diff
@@ -0,0 +1,58 @@
...
+<#
+    PushIndent(standardIndent);
+#>
+/// <summary>
+/// Defines utility methods to generate SPIRV operations
+/// </summary>
+[CLSCompliant(false)]
+public interface ISPIRVBuilder
+{
+
```
m4rs-mt, line 24: nit: remove empty line.
Question: do we really need to declare the whole interface in the T4 template? We can refine the code to define a common base interface as a partial interface in another C# file.

Src/ILGPU/Backends/SPIRV/ISPIRVBuilder.tt
```diff
@@ -0,0 +1,58 @@
...
+#><#@ assembly name="System.Xml" #>
+<#@ assembly name="System.Core"#>
+<#@ import namespace="System.Linq"#>
+using System;
+using System.Collections.Generic;
+
+#nullable enable
+#pragma warning disable 1591
+
+namespace ILGPU.Backends.SPIRV {
```
m4rs-mt, line 13: nit: coding convention regarding braces

Src/ILGPU/Backends/SPIRV/ISPIRVBuilder.tt
```diff
@@ -0,0 +1,58 @@
...
+    // deal with it
+    void Merge(ISPIRVBuilder other);
+
+<#
+    PushIndent(standardIndent);
+    foreach (var instruction in grammar.Instructions.Instructions)
+    {
+        if (!IsInstructionCLSCompliant(instruction))
+        {
+#>[CLSCompliant(false)]
```
m4rs-mt, line 41: Remark: if we change the interface to work on `SpirVInstructions`, we can skip all of these attributes.

Src/ILGPU/Backends/SPIRV/SPIRVBuilderUtils.cs
```diff
@@ -0,0 +1,85 @@
...
+    /// </summary>
+    public static class SPIRVBuilderUtils
+    {
+        /// <summary>
+        /// Attempts to convert an object to a SPIRV compatible representation
+        /// </summary>
+        /// <param name="obj">The object to convert</param>
+        /// <returns>A list of words (uints) representing the object</returns>
+        /// <exception cref="NotImplementedException"></exception>
+        public static List<uint> ToUintList(object obj)
```
m4rs-mt, line 20: I am not sure about this function in general, as it seems to be dangerous to use. Can we split/refine this function to avoid using objects?

Src/ILGPU/Backends/SPIRV/SPIRVXMLReader.ttinclude
```diff
@@ -0,0 +1,142 @@
...
+        {
+            return _grammarCache;
+        }
+
+        var xml = new XmlSerializer(typeof(SPIRVGrammar));
+        using var file = File.Open(Path.Combine(Host.ResolvePath("."), "spirv.grammar.xml"), FileMode.Open);
+
+        var grammar = (SPIRVGrammar) xml.Deserialize(file);
+
+        //Sanitize operand names
```
m4rs-mt, line 20: nit: space missing between `//` and `S`

Src/ILGPU.Tests.SPIRV/SPIRVBuilderUtilsTests.cs
Yey007: What exactly would you want to include in this structure? Basically the only thing that all ops share is an OpCode and WordCount. Perhaps I'm misunderstanding...

Src/ILGPU/Backends/SPIRV/BinarySPIRVBuilder.tt
Yey007: I'm not sure if it will change the amount of copies we have to do in that function, but it will probably help somewhere else. I'll change it.

Src/ILGPU/Backends/SPIRV/ISPIRVBuilder.tt
Yey007: So like move this section
```csharp
byte[] ToByteArray();

void AddMetadata(uint magic, uint version, uint genMagic, uint bound, uint schema);

// This is the best way I could come up with to
// handle trying to merge different builders
// Implementing classes will kinda just have to
// deal with it
void Merge(ISPIRVBuilder other);
```
to its own file? I guess we could do that but I'm not sure if it helps that much. If we do go forward with it we should do the same with all the other classes.

Src/ILGPU/Backends/SPIRV/SPIRVBuilderUtils.cs
Yey007: I agree. This function is not very cohesive and prone to error. I'll split it up to be more type safe.

Question: There will still be some amount of danger as I don't see a way to get around reflection. Is this okay? Can you think of a way to get around it?

Yey007:
@m4rs-mt @MoFtZ don't mean to be pushy but is there any progress on this? If there are things that need to be discussed I'm open to that. Don't be afraid to tear the PR apart if the implementation is not up to standard!

Thanks

MoFtZ:
@Yey007 I'm not familiar with SPIR-V, so I don't think I'll be much help in reviewing that side of things.

One thing we need to consider is how to handle the file `spirv.grammar.xml`. Does this file need to live in a separate repository?

We originally had issues with processing the original JSON, and that's why we converted it to XML. Should we try to get the JSON processing working? That would potentially allow us to use a direct link to the original JSON. Alternatively, should we be linking to the JSON and converting to XML on-the-fly?

m4rs-mt:
@Yey007 Thank very much you for your work and efforts you have put into this work. As @MoFtZ pointed out, we discussed several ways to integrate SPIR-V support into ILGPU. We concluded that it would be best to try to avoid importing the grammar files directly. As an alternative, we considered that we could fork the grammar repository and add a custom "builder generator" to that repository that would create all the bindings outside of the ILGPU repository. ILGPU can then use the pre-built bindings instead of regenerating them during the build process.

m4rs-mt:
I am closing this PR in favor of a new one.

PR #616, opened 2021-09-11
Allow building `net471` target without Windows

lostmsu:


MoFtZ (changes requested):
QUESTION: `Microsoft.NETFramework.ReferenceAssemblies` is an interesting Nuget package - first time I have heard of it. Should we be using the more specific `Microsoft.NETFramework.ReferenceAssemblies.net471`?

Also, if we plan on using this Nuget package, we should apply it to all the other locations in ILGPU too. e.g. unit test projects.

lostmsu:
For the reference, [official doc](https://github.com/Microsoft/dotnet/tree/master/releases/reference-assemblies).

AFAIK Microsoft recommends using the "meta-package", however, I used specific version too without any issues.

Only the projects, that target .NET Framework need this. It just allows .NET SDK to build `net4xx` target on non-Windows OS and on Windows OS where SDK targeting pack is not installed. I noticed now, that test projects actually have `net471` too, but only for GitHub actions. I don't think this change is needed for GitHub actions, because unless you want to add Mono to your test matrix, there's no point in building `net471` test target on non-Windows.

MoFtZ (approved):

lostmsu:
@MoFtZ  do you want me to rebase this?

MoFtZ:
This is unlikely to cause conflicts, so no need to rebase. Just wait for @m4rs-mt to review.

m4rs-mt (approved):
Looks also good to me 👍 Thanks a lot for this simplification!

Issue #614, opened 2021-09-09
`AccessViolationException` trying to run a specialized kernel on CPU


lostmsu:
Sorry I did not narrow the issue down further.

There's the gist demonstrating the problem (about 100 lines total):
https://gist.github.com/lostmsu/debe2a5f4565189f9e722f1dfeb6050f

This is on `1.0.0-beta3`. I have not tested earlier versions or other accelerator implementations.

OS: Windows 10

lostmsu:
```csharp
Kernel<Add, N2<bool>>(11,
    leftData.View, rightData.View,
    result,
    cpu.Allocate1D(new long[] { 4, 3 }).View,
    cpu.Allocate1D(new long[] { 3, 1 }).View,
    SpecializedValue.New(EquatableFixedArray.New(new FixedArray<N2<bool>, bool> {
        [0] = false, [1] = true,
    })),
    SpecializedValue.New(EquatableFixedArray.New(new FixedArray<N2<bool>, bool> {
        [0] = true, [1] = false,
    })));
```

lostmsu:
In case it is the culprit, implementation of `UnsafeRef`:
https://github.com/losttech/TypeNum/blob/f28befa00b82d2d154cc947b529520f77db2c5a6/src/FixedArray.cs#L39

MoFtZ:
hi @lostmsu. OK, I've managed to reduce the problem down to a small example:
```CSharp
public class Program
{
    public struct MyStruct : IEquatable<MyStruct>
    {
        public int Data;

        public MyStruct(int data)
        {
            Data = data;
        }

        public bool Equals(MyStruct other) =>
            Data.Equals(other.Data);

        public override bool Equals(object? obj)
            => obj is MyStruct other && Equals(other);
    }

    static void Main()
    {
        var compute = Context.CreateDefaultAutoDebug();
        var cpu = CPUDevice.Default.CreateCPUAccelerator(compute);

        var kernel = cpu.LoadAutoGroupedStreamKernel<
            Index1D,
            SpecializedValue<MyStruct>,
            SpecializedValue<MyStruct>>(MyKernel);
        kernel(12,
            SpecializedValue.New(new MyStruct(1)),
            SpecializedValue.New(new MyStruct(2)));
    }

    static void MyKernel(
        Index1D index,
        SpecializedValue<MyStruct> first,
        SpecializedValue<MyStruct> second)
    {
    }
}
```
[...]

Issue #415, opened 2021-02-12
Feature request: add support of runtime compiled parts of kernel  


mikhail-khalizev:
Faced the following problem. There is a common kernel logic. And there are additional methods for transformation numbers that are generated dynamically before starting the kernel (as in the example below, which fails).

I propose to add support for executing ```Action``` and ```Func``` through a mechanism like ```SpecializedValue```.

```csharp
using ILGPU;
using ILGPU.Algorithms;
using ILGPU.Runtime;
using System;
using System.Linq;
using System.Linq.Expressions;

namespace SimpleKernel
{
    public class Program
    {
        private static Func<int, int> _transform;

        static unsafe void MyKernel(ArrayView<int> dataView)
        {
            if (Grid.GlobalIndex.X == 0)
                dataView[0] = _transform(10);
        }

        static unsafe void Main()
        {
            var arg = Expression.Parameter(typeof(int), "arg");

            // Runtime generated transformation of number.
            _transform = Expression.Lambda<Func<int, int>>(
                Expression.Add(
                    arg,
                    Expression.Constant(1)),
                arg).Compile();


            using var context = new Context();
            
            foreach (var acceleratorId in Accelerator.Accelerators)
            {
                using var accelerator = Accelerator.Create(context, acceleratorId);

                Console.WriteLine($"Performing operations on {accelerator}");
                        
                var kernel = accelerator.LoadStreamKernel<ArrayView<int>>(MyKernel);
                var config = accelerator.ComputeGridStrideLoopExtent(1);

                using (var buffer = accelerator.Allocate<int>(1))
                {
                    buffer.MemSetToZero();

                    kernel(config, buffer);
                    accelerator.Synchronize();
                    var data = buffer.GetAsArray()[0];
                            
                    Console.WriteLine($"Data: {data}");
                }

                Console.WriteLine();
            }
        }
    }
}
```

faruknane:
@mikhail-khalizev I am working on a library that allows you to write dynamic functions and to use/call them in statically created kernels. In a few days, I will try to share two examples of how you can do what you wanted to do.

faruknane:
[...]

PR #617, opened 2021-09-12
Fixed generated call to `ValueType.GetHashCode`

lostmsu:
The bug (https://github.com/m4rs-mt/ILGPU/issues/614) occured when type `T` inside `SpecializedValue<T>` does not implement `GetHashCode`, and instead inherits one from `ValueType`.

Because `ValueType` itself is not a struct, to call its `GetHashCode` the first parameter must be an object reference. Therefore the object of type `T` (which `SpecializedValue<T>` requires to be a `struct`) must be boxed before calling `GetHashCode`.

### DIFF
--- a/Src/ILGPU/Backends/IL/ILEmitterExtensions.cs
+++ b/Src/ILGPU/Backends/IL/ILEmitterExtensions.cs
@@ -67,10 +67,26 @@ namespace ILGPU.Backends.IL
                 emitter.Emit(LocalOperation.Load, result);
 
                 emitter.Emit(OpCodes.Ldarg_0);
-                emitter.Emit(OpCodes.Ldflda, field);
-                emitter.EmitCall(field.FieldType.GetMethod(
+
+                var fieldHashCode = field.FieldType.GetMethod(
                     GetHashCodeInfo.Name,
-                    BindingFlags.Public | BindingFlags.Instance));
+                    BindingFlags.Public | BindingFlags.Instance);
+                if (!field.FieldType.IsValueType)
+                {
+                    emitter.Emit(OpCodes.Ldfld, field);
+                }
+                else if (fieldHashCode.DeclaringType != field.FieldType)
+                {
+                    // to call GetHashCode inherited from ValueType, struct must be boxed
+                    emitter.Emit(OpCodes.Ldfld, field);
+                    emitter.Emit(OpCodes.Box, field.FieldType);
+                }
+                else
+                {
+                    emitter.Emit(OpCodes.Ldflda, field);
+                }
+
+                emitter.EmitCall(fieldHashCode);
 
                 emitter.Emit(OpCodes.Xor);
                 emitter.Emit(LocalOperation.Store, result);
--- a/Src/ILGPU.Tests/Generic/TestData.cs
+++ b/Src/ILGPU.Tests/Generic/TestData.cs
@@ -83,6 +83,22 @@ namespace ILGPU.Tests
         public override int GetHashCode() => 0;
     }
 
+    [Serializable]
+    // warning disabled intentionally for testing this scenario
+    #pragma warning disable CS0659 // Type does not override Object.GetHashCode()
+    public struct NoHashCodeStruct : IXunitSerializable, IEquatable<NoHashCodeStruct>
+    #pragma warning restore CS0659 // Type does not override Object.GetHashCode()
+    {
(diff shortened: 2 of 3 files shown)


Src/ILGPU/Backends/IL/ILEmitterExtensions.cs
```diff
@@ -67,10 +67,21 @@ public static class ILEmitterExtensions
...
                     GetHashCodeInfo.Name,
-                    BindingFlags.Public | BindingFlags.Instance));
+                    BindingFlags.Public | BindingFlags.Instance);
+                if (field.FieldType.IsValueType &&
+                    fieldHashCode.DeclaringType != field.FieldType)
+                {
+                    emitter.Emit(OpCodes.Ldfld, field);
+                    emitter.Emit(OpCodes.Box, field.FieldType);
+                } else
+                {
```
MoFtZ, line 80: MINOR: Coding convention.
```suggestion
                }
                else
                {
```

Src/ILGPU/Backends/IL/ILEmitterExtensions.cs
```diff
@@ -67,10 +67,21 @@ public static class ILEmitterExtensions
...
                 emitter.Emit(OpCodes.Ldarg_0);
-                emitter.Emit(OpCodes.Ldflda, field);
-                emitter.EmitCall(field.FieldType.GetMethod(
+
+                var fieldHashCode = field.FieldType.GetMethod(
                     GetHashCodeInfo.Name,
-                    BindingFlags.Public | BindingFlags.Instance));
+                    BindingFlags.Public | BindingFlags.Instance);
+                if (field.FieldType.IsValueType &&
+                    fieldHashCode.DeclaringType != field.FieldType)
```
MoFtZ, line 75: MINOR: This condition can probably be simplified since `SpecializedValue` only accepts `unmanaged` types.
```suggestion
                if (fieldHashCode.DeclaringType != field.FieldType)
```

MoFtZ (changes requested):
Nice work @lostmsu, appears to behave as expected.

Could you please also add a unit test for this condition? Update `Src/ILGPU.Tests/SpecializedKernels.cs` based on the minimal example.

Src/ILGPU/Backends/IL/ILEmitterExtensions.cs
lostmsu: @MoFtZ I was looking forward to it supporting other types as requested in https://github.com/m4rs-mt/ILGPU/issues/415 and https://github.com/m4rs-mt/ILGPU/issues/415

MoFtZ: We can add support later.

`SpecializedValue` applies the `unmanaged` restriction, and `Action<>` and `Func<>` are reference types, so they cannot be used unless we change `SpecializedValue`, introduce a new type like `SpecalizedFuncValue`, or some other API change.

Also, the original condition would apply boxing if "it is a ValueType and does not match the declaring type". However, wouldn't we want to apply boxing to reference types as well? Easier to deal with it later when we have defined an API, and unit tests to ensure it all works.

lostmsu: @MoFtZ AFAIK reference type values are "boxes". Which means the version I made does not handle them correctly. They only need `Ldfld` (and not `Ldflda`). If neither of us will be the one to implement specialized lambdas, the person who would will likely stumble upon this error and get puzzled too.

I will implement this code for reference types, but won't add test coverage.

lostmsu:
Added support for reference types.

Added test for struct without explicit `GetHashCode`

Src/ILGPU/Backends/IL/ILEmitterExtensions.cs
lostmsu: This should be fixed now.

lostmsu:
disabled warning for test struct

MoFtZ (approved):

lostmsu:
Do you need anything from me to complete this change?

MoFtZ:
@lostmsu I do not have permissions on this repository. Have to wait for @m4rs-mt to review this PR.

m4rs-mt (approved):
Looks good to me and thanks a lot for your patch 🚀 Would you also mind to agree on the copyright transfer on #598 before we merge this PR?

lostmsu:
@m4rs-mt done

Issue #623, opened 2021-10-01
Memory error using blas with multiple accelerators


Ruberik:
### The very short summary
The following series of function calls causes a crash:
```
accelerator0.Synchronize();
accelerator1.Synchronize();
blas0.Axpy(...)
blas1.Axpy(...)
accelerator0.Synchronize();
accelerator1.Synchronize();  // The crash happens here
```
I would have expected this set of function calls to execute the Axpy operations on each GPU without crashing. 

### The plot thickens
The following series of function calls _does not_ cause a crash:
```
accelerator0.Synchronize();
accelerator1.Synchronize();
blas0.Axpy(...)
accelerator1.Synchronize();  // This line was added. Note that nothing has happened on accelerator1 since the last Synchronize() call.
blas1.Axpy(...)
accelerator0.Synchronize();
accelerator1.Synchronize(); 
```

### The full repro
```
        static void Main(string[] args) {
            using (var context = Context.Create().Cuda().EnableAlgorithms().ToContext()) {
                var acceleratorCount = context.GetCudaDevices().Count;
                var accs = Enumerable.Range(0, acceleratorCount).Select(i => context.CreateCudaAccelerator(i)).ToArray();
                var blases = accs.Select(acc => new CuBlas(acc, CuBlasAPIVersion.V11)).ToArray();
                try {
                    Console.WriteLine($"Calling the one we expect to succeed");
                    AxPy0(accs, blases, true);
                    Console.WriteLine($"Calling the one we expect to crash");
                    AxPy0(accs, blases, false);
                } finally {
                    foreach (var b in blases) b.Dispose();
                    foreach (var acc in accs) acc.Dispose();
                }
            }
        }

        private static void AxPy0(CudaAccelerator[] accelerators, CuBlas[] blases, bool synchronizeBeforeCall) {
            int size = 1024;
            var derivs = accelerators.Select(x => x.Allocate1D<float>(size)).ToArray();
            var values = accelerators.Select(x => x.Allocate1D<float>(size)).ToArray();
            foreach (var acc in accelerators) acc.Synchronize();
            try {
                for (int i = 0; i < accelerators.Length; i++) {
                    Console.WriteLine($"  Sending calls to accelerator {i}");
                    var acc = accelerators[i];
                    var blas = blases[i];
                    if (synchronizeBeforeCall) acc.Synchronize();  // If false, crash. If true, no crash.
[...]

PR #624, opened 2021-10-02
Ensure Cuda accelerator is bound before calling CuBlas methods.

MoFtZ:
Fixes https://github.com/m4rs-mt/ILGPU/issues/623.

Before calling a CuBlas method, ensure that the associated Accelerator is made the current accelerator.


m4rs-mt (approved):

Ruberik:
I this thread-safe? Like, could thread A call `EnsureAcceleratorBinding`, then thread B could call `EnsureAcceleratorBinding`, then thread A calls its cublas function and crashes?

Ruberik:
Oh, I see now, https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#device-selection indicates this binding is per-thread. Ignore me.

PR #627, opened 2021-10-03
Fixed race condition in WarpReductions.Reduce for CPU accelerators.

m4rs-mt:
This PR fixes a critical race condition affecting the CPU implementation of `WarpExtensions.Reduce` and `WarpExtensions.AllReduce`.

### DIFF
--- a/Src/ILGPU.Algorithms/IL/ILFunctions.cs
+++ b/Src/ILGPU.Algorithms/IL/ILFunctions.cs
@@ -47,6 +47,16 @@ namespace ILGPU.Algorithms.IL
             int ThreadDimension { get; }
 
             /// <summary>
+            /// The number of segments for reduce operations.
+            /// </summary>
+            int ReduceSegments { get; }
+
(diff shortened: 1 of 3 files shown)


MoFtZ (approved):

MoFtZ:
QUESTION: Is there any plan for fixing the hard-coded [WarpImplementation.MaxNumThreads](https://github.com/m4rs-mt/ILGPU/blob/v1.0.0-rc1/Src/ILGPU.Algorithms/IL/ILWarpExtensions.cs#L36)?

m4rs-mt:
> QUESTION: Is there any plan for fixing the hard-coded [WarpImplementation.MaxNumThreads](https://github.com/m4rs-mt/ILGPU/blob/v1.0.0-rc1/Src/ILGPU.Algorithms/IL/ILWarpExtensions.cs#L36)?

What we could do is to simply use the `PTX` implementations, as they will also work on our CPU accelerator 🤓

PR #625, opened 2021-10-02
Adjusted copyright headers (based on #598)

m4rs-mt:
This PR adjusts all copyright headers based on the discussion in #598. This is the first PR that adjusts all copyright headers.

PR #636, opened 2021-10-07
Implemented tool to update copyright year.

MoFtZ:
@m4rs-mt
I have implemented a powershell script that parses the ILGPU repository, and updates the copyright headers. It uses the existing copyright year found in the file, and extends the date range to include the current year. If there is no existing copyright header, it adds a new one.

It also handles the `.nuspec.targets` and `LICENSE.txt` files.

If you run the script, it will recurse over all files/folders from the script location.
`.\UpdateCopyrightYear.ps1`

This should help with PR #625.


MoFtZ:
@jgiannuzzi Last modification date makes sense.

Can we rely on `git log` for the starting date? I have previously seen cases where git can sometimes add/remove a file when it was moved or renamed. This either happens with a small file, or when changing many files.

jgiannuzzi:
> Can we rely on `git log` for the starting date? I have previously seen cases where git can sometimes add/remove a file when it was moved or renamed. This either happens with a small file, or when changing many files.

Good point. `--follow` should help us with that, but I am not 100% sure it will work. It's probably worth checking what it would give for some of our oldest files before we decide to use it or not.

jgiannuzzi:
@MoFtZ ah that's too bad 😞

I agree with your current logic for completing the missing information. It's a massive step forward!

Probably not for this PR, but what would you think of moving the various scripts we wrote for the CI into actual files like this one, and putting them all into a separate directory?

Another future idea would be to automate copyright headers through the CI (e.g. something that runs on pushes to master and adds a new commit with updated copyright if required)

MoFtZ:
> Probably not for this PR, but what would you think of moving the various scripts we wrote for the CI into actual files like this one, and putting them all into a separate directory?

Sounds like a good idea. Maybe `Build`, `Tools` or `Scripts` directory? Alternatively, it can live within the `.github` directory.

> Another future idea would be to automate copyright headers through the CI (e.g. something that runs on pushes to master and adds a new commit with updated copyright if required)

I have a work-in-progress branch [here](https://github.com/MoFtZ/ILGPU/tree/feature/copyright-schedule) 😄 It produces a PR similar to [this](https://github.com/MoFtZ/ILGPU/pull/17).

MoFtZ:
> 
> 
> > @jgiannuzzi I'm not sure we can use git for the starting date.
> > Using [2a41407](https://github.com/m4rs-mt/ILGPU/commit/2a41407167bbc75938be55fd413444884c996f18) as an example, `Src/ILGPU/TypeInformation.tt` was removed, and `Src/ILGPU/Static/TypeInformation.ttinclude` was added.
> > The commit log for `TypeInformation.ttinclude` has an initial copyright year of 2020. The commit log for `TypeInformation.tt` has an initial copyright year of 2017. The contents of `TypeInformation.tt` has an initial copyright year of 2016, but git only has commits back to the initial github commit in 2017.
> 
> The issue here is, that the affected files were (at least partially) written in 2016 and committed to the repository in 2017. I think we can "just change" the copyright year to 2017...

The issue is that if we use the git history to automate the starting copyright  year, `TypeInformation.ttinclude` would have a copyright year of `2020-2021`, instead of `2017-2021` or even `2016-2021`.

Git did not track the rename of  `TypeInformation.tt` to `TypeInformation.ttinclude`, because it does not deal with filenames, only file contents. It uses heuristics to identify renames, which seems to have not worked in [that commit](https://github.com/m4rs-mt/ILGPU/commit/2a41407167bbc75938be55fd413444884c996f18).

jgiannuzzi:
> > Probably not for this PR, but what would you think of moving the various scripts we wrote for the CI into actual files like this one, and putting them all into a separate directory?
> 
> Sounds like a good idea. Maybe `Build`, `Tools` or `Scripts` directory? Alternatively, it can live within the `.github` directory.

I guess it depends how much we want the scripts to be easy to execute for contributors. If we don't really expect anyone to run them locally, I'd go with `.github/workflows/scripts`.

> > Another future idea would be to automate copyright headers through the CI (e.g. something that runs on pushes to master and adds a new commit with updated copyright if required)
> 
> I have a work-in-progress branch [here](https://github.com/MoFtZ/ILGPU/tree/feature/copyright-schedule) 😄 It produces a PR similar to [this](https://github.com/MoFtZ/ILGPU/pull/17).

This looks great!

Maybe we do want it running after every successful push to `master` instead of the schedule? That way the copyright is almost always up-to-date. But I can see the counterpoint that it might just create more noise 🤷
A way to avoid the noise might be to simply create a new commit instead of a push, in which case we might want to run it after the CI was successful for a push to `master` so we avoid having too many concurrent runs of the CI. Something like this:
```yaml
on:
  workflow_run:
    workflows: ["CI"]
    branches: [master]
    types: [completed]
```

MoFtZ:
> On a different note, this script is an amazing piece of PowerShell. I had no idea you could do such sophisticated things in that language 🤯. It got me thinking however that maybe we want this to become an actual C# program in the future? Any contributor to ILGPU will have .NET installed, but they might not automatically have PS, and running a C# program can be as simple as a `dotnet run`. Thoughts?

Yes, I had considered writing this as a C# program. However, it adds complications to the CI pipeline. @jgiannuzzi How would this work with `dotnet run`? Would we have to create a tool, similar to `T4.Build`?

> That way there will never be a PR getting merged with the wrong headers, the check is automated (so no burden on reviewers), and the fix is automated in most cases (so almost no burden on authors). What do you think?

That's an interesting workflow. I was originally thinking more like dependabot, with occasional PRs raised to "fix" the repository. We might need to test if the CI pipeline has permission to update the branch of a forked repository.

m4rs-mt:
@jgiannuzzi Any news on this?

jgiannuzzi:
> Yes, I had considered writing this as a C# program. However, it adds complications to the CI pipeline. @jgiannuzzi How would this work with `dotnet run`? Would we have to create a tool, similar to `T4.Build`?

I think it could probably be added as an inline MSBuild task. If it's too big for readability, I would recommend using the `Source` attribute of the `Code` element to reference a separate file (see https://docs.microsoft.com/en-us/visualstudio/msbuild/msbuild-roslyncodetaskfactory?view=vs-2022#code-element).
 
> That's an interesting workflow. I was originally thinking more like dependabot, with occasional PRs raised to "fix" the repository. We might need to test if the CI pipeline has permission to update the branch of a forked repository.

I was thinking of having this run as part of the regular `push` CI pipeline on the forked repo itself. Anyways, this is a future improvement and should not block this PR. And the same goes for potentially migrating the code to C#!

LICENSE.txt
```diff
@@ -1,4 +1,4 @@
-********************************************************************************
+﻿********************************************************************************
```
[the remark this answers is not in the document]
MoFtZ: This is a side-effect of how we are writing to the file.

We use `File.WriteAllText` and specify the UTF-8 encoding. Should probably update this to check if the existing file has a BOM.

MoFtZ:
> I think it could probably be added as an inline MSBuild task. If it's too big for readability, I would recommend using the `Source` attribute of the `Code` element to reference a separate file (see https://docs.microsoft.com/en-us/visualstudio/msbuild/msbuild-roslyncodetaskfactory?view=vs-2022#code-element).

OK, sounds good to me - I will investigate this option.

MoFtZ:
OK, I have re-implemented the Powershell script as a C# console application.

By default, if you run the application, it will use the current directory. However, if you run the application in VS2022, the launch settings will default to the root of the Git repository. On my computer, it takes around 50 seconds to process all the files.

LICENSE.txt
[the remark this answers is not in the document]
MoFtZ: The replacement C# console application fixes this issue by checking for the UTF8 BOM.

.github/workflows/updateCopyrightYear.yml
```diff
@@ -0,0 +1,36 @@
...
+        with:
+          dotnet-version: 6.0.x
+
+      - name: Update copyright year
+        uses: technote-space/create-pr-action@v2
+        with:
+          EXECUTE_COMMANDS: |
+            dotnet run --configuration=Release -p:TreatWarningsAsErrors=true --project Tools/CopyrightUpdateTool
+          COMMIT_MESSAGE: 'Bump copyright year.'
+          COMMIT_NAME: 'Marcel Koester'
```
m4rs-mt, line 32: Question: should we change this to a more generic committer (like ILGPU CLI)?

m4rs-mt (approved):

.github/workflows/updateCopyrightYear.yml
MoFtZ: What should we use for the email address?

m4rs-mt: What about `contact@ilgpu.net`? I will definitely set up this address as the primary contact address for the new website anyway.

MoFtZ: Done.

PR #646, opened 2021-10-14
Added metapackage containing all ILGPU libraries.

MoFtZ:
Not sure if we want this meta-package or not, but might be useful.
Also not sure about the name.


m4rs-mt:
@MoFtZ I really like this idea 👍 

I have to admit that I prefer @jgiannuzzi's idea of creating a "new" `ILGPU` package instead of an `ILGPU.All` package.

MoFtZ:
@jgiannuzzi @m4rs-mt
I agree that the name `ILGPU.All` is not great. However, I was trying to avoid a breaking change. Also, it matched some existing names like `Microsoft.AspNetCore.All`.

I'll put this PR on hold until we decide on a timeframe (and name?).

MoFtZ:
Something else to consider, if we rename `ILGPU` to `ILGPU.Core`, are we only renaming the nuget package? Or all the files/folders?

That would include:
- `Src/ILGPU` to `Src/ILGPU.Core`
- `Src/ILGPU.Tests` to `Src/ILGPU.Core.Tests`
- etc

MoFtZ:
Renames the ILGPU nuget package to ILGPU.Core.
Adds a new ILGPU metapackage.

MoFtZ:
Abandoning this PR.

In the future e.g. v2, it will be easier to deprecate `ILGPU.Algorithms` and release the `ILGPU` package with `ILGPU.Algorithms` baked in.

Issue #652, opened 2021-10-19
[1.0.0-rc2] serious bug with loop unrolling


deng0:
Today I've stumbled upon a serious bug with the ILGPU code generation when it does loop unrolling.

Consider a simple kernel like this:
```c#
        public static void MyKernel(Index1D index, ArrayView<int> dataView)
        {
            for (int i = 0; i <= 2; i++)
            {
                dataView[index] = i;
            }
        }
```

In the 1.0.0-rc2 and (rc1) the dataView will be filled with 1 instead of 2 for CUDA and OpenCL Accelerators.
It's obviously fine for the CPUAccelerator.

And it also works when using the old 0.9.2 ILGPU Version.

It seems the loop unrolling of ILGPU 1.0.0-rc does not handle the "<=" correctly (it handles it like "<").

When doing "for (int i = 0; i <= 33; i++)" it starts to work again, probably because it's not unrolled (DefaultMaxUnrollFactor = 32).

PR #661, opened 2021-10-26
Fixed invalid value updates of dependent values in LoopUnrolling

m4rs-mt:
This PR is part of the solution for issue report #652. As also described in PR #657, the following kernel program raises an exception in the `Release` and `O2` pipelines using the latest `master` branch:
```c#
            int value = 3;
            int counter = 0;
            do
            {
                ++value;
            }
            while (counter++ < 3);
            data[index] = value;
```
This PR fixes the `LoopUnrolling` transformation by tracking value updates for values being referenced outside of the loop in the presence of `do-while` loops. It also adds an "additional" test cases (it is an old test case which accidentally did not cover this edge case.... 😞) to reproduce this problem.

However, this PR does not fix all the problems related to "do-while" loops (yet...), since we have to perform an additional unwinding step at the beginning (+1 additional "iteration" compared to "while" loops). This raises the question of how to distinguish between these two types of loops. One possibility would be to update PR #657 to adjust the trip count of these loops.

@MoFtZ @deng0: What do you think?

*Note that this PR is a draft and not ready to be used in production*

### DIFF
--- a/Src/ILGPU.Tests/BasicLoops.cs
+++ b/Src/ILGPU.Tests/BasicLoops.cs
@@ -214,7 +214,9 @@ namespace ILGPU.Tests
 
         internal static void DoWhileIncrementKernel(
             Index1D index,
-            ArrayView1D<int, Stride1D.Dense> data)
+            ArrayView1D<int, Stride1D.Dense> data,
+            ArrayView1D<int, Stride1D.Dense> data2,
+            ArrayView1D<int, Stride1D.Dense> data3)
         {
             int value = 3;
             int counter = 0;
@@ -224,6 +226,22 @@ namespace ILGPU.Tests
             }
             while (counter++ < 3);
             data[index] = value;
+
+            int value2 = 1;
+            do
+            {
+                value2 *= 3;
+            }
+            while (counter++ == 5);
+            data2[index] = value2;
+
+            int value3 = 13;
+            do
+            {
+                --value3;
+            }
+            while (counter++ != 8);
+            data3[index] = value3;
         }
 
         [Fact]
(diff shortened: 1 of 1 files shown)


deng0:
@m4rs-mt I made some more tests and could not find any additional problems.
Only the ones I mentioned in my initial pull request:

open issues:
- critical: do not unroll when iter value is modified inside the loop
- optional: add support for ushort, short, byte, sbyte iter types

jgiannuzzi:
@m4rs-mt I agree that the support for additional iter types should be part of `v1.1` and not `v1.0`. We are too late in the release process to introduce such changes and should only focus on fixing critical bugs.

Thanks a lot @deng0 for reporting these issues, and thanks a lot @m4rs-mt for fixing this!

PS: this is also how I feel 😅:
> @m4rs-mt Thank you for your work. I don't understand most of it.

deng0:
@m4rs-mt I agree, the other iter types can wait for v1.1.

I've just checked all my testkernels with your latest commit and now they all produce the correct results.
Great work!

Src/ILGPU/IR/Transformations/LoopUnrolling.cs
```diff
@@ -465,6 +519,10 @@ private struct LoopProcessor : Loops.ILoopProcessor
                 // Replace all phi values with their computed values
                 loopSpecializer.ReplacePhis();
 
+                // Replace all values with used outside of the loop with their computed
+                // values
```
MoFtZ, line 523: MINOR: Spelling.
```suggestion
                // Replace all values used outside of the loop with their computed values
```

MoFtZ (approved):

PR #671, opened 2021-11-05
Added on-the-fly optimization for nested SetField operations.

m4rs-mt:
This PR adds an additional on-the-fly optimization of `SetField` operations by automatically converting chained `SetField`s into explicit `StructureValues`. This improves the optimization potential of all transformations.


Src/ILGPU/IR/Construction/Structures.cs
```diff
@@ -522,6 +522,50 @@ internal ValueReference FinishStructureBuilder<TBuilder>(ref TBuilder builder)
                 return instance.Seal();
             }
 
+            // Optimize common cases in which this set field operation fills a whole
+            // structure instance
+            if (objectValue is SetField setField &&
+                !fieldSpan.HasSpan &&
+                fieldSpan.Index + 1 == structureType.NumFields &&
```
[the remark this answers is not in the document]
m4rs-mt: Nice catch, that should make the code more applicable to additional scenarios.

Src/ILGPU/IR/Construction/Structures.cs
```diff
@@ -522,6 +522,49 @@ internal ValueReference FinishStructureBuilder<TBuilder>(ref TBuilder builder)
...
+                        otherSetField.FieldSpan.HasSpan ||
+                        fieldValues[otherSetField.FieldSpan.Index] != null)
+                    {
+                        break;
+                    }
+
+                    // Register this set-field operation
+                    fieldValues[otherSetField.FieldSpan.Index] = otherSetField.Value;
+                    // Decrease the traversal length
+                    --traversalLength;
```
MoFtZ, line 552: ISSUE: This section also needs to deal with field spans greater than a single field.

MoFtZ (changes requested):

Src/ILGPU/IR/Construction/Structures.cs
m4rs-mt: You are right 👍. After changing the if condition, we also need to update this code.

Src/ILGPU/IR/Construction/Structures.cs
```diff
@@ -522,6 +522,49 @@ internal ValueReference FinishStructureBuilder<TBuilder>(ref TBuilder builder)
...
+
+                // Traverse the whole chain of all set-field operations
+                var parent = objectValue;
+                int traversalLength = structureType.NumFields - 1;
+                while (!(parent is NullValue) && traversalLength > 0)
+                {
+                    // Check whether we are still on the right track while traversing
+                    // set of (possibly unordered) set-field operations
+                    if (!(parent is SetField otherSetField) ||
+                        otherSetField.FieldSpan.HasSpan ||
```
m4rs-mt, line 543: (see comment below)

Src/ILGPU/IR/Construction/Structures.cs
m4rs-mt: After inspecting the code more closely, it turns out that this is not necessary, because we support single-element `SetField` operations only in the inner loop (see comment). The question is whether we should relax these restrictions.... I'm not sure if it's really worth the effort....

MoFtZ: yup, that's fine.
should we change `fieldSpan.Index + fieldSpan.Span` back to `fieldSpan.Index + 1`?

MoFtZ (approved):

PR #684, opened 2021-11-09
Enhanced alignment functionality

m4rs-mt:
This PR improves the existing `AlignTo` API by providing an additional `AsAligned` method to inform ILGPU about domain-specific alignment information that cannot be automatically resolved with any of the existing program analyses. Previously, this information could be expressed (at least partially) in ILGPU, but it could not be preserved until it reached one of the backends, resulting in the generation of suboptimal code in many cases.

Most importantly, this PR ensures that domain-specific alignment information is preserved through all phases of the compiler and is accessible from one of the backends. This greatly improves vectorization capabilities and allows us to generate specialized alignment code for different platforms/backends. Part of the changes include a new hand-tuned alignment implementation for NVIDIA GPUs, which is about 16% faster (on average) than the currently available version in the master branch.

Note that this PR is based on #668.


Src/ILGPU/Backends/OpenCL/CLCodeGenerator.Values.cs
```diff
@@ -398,6 +399,82 @@ public void GenerateCode(LoadFieldAddress value)
...
+                    statement.AppendCommand(')');
+                }
+            }
+
+            // if (baseOffset == 0) { 0 } else { alignment - baseOffset }
+            var adjustment = AllocateType(arithmeticBasicValueType);
+            using (var selectStatement = BeginStatement(adjustment))
+            {
+                selectStatement.AppendArgument(baseOffset);
+                selectStatement.AppendCommand(" == 0 ? 0 : ");
```
MoFtZ, line 445: ISSUE: The ternary operator may need brackets.
Currently, `baseOffset == 0 ? 0 : alignment - baseOffset` is ambiguous.
Should use `baseOffset == 0 ? 0 : (alignment - baseOffset)`.

Src/ILGPU/Backends/PTX/PTXCodeGenerator.Values.cs
```diff
@@ -628,6 +628,154 @@ public void GenerateCode(LoadFieldAddress value)
...
+            {
+                using var convert = BeginCommand(
+                    convertOperation,
+                    predicate: predicate.GetConfiguration(false));
+                convert.AppendArgument(alignmentOffsetRegister);
+                convert.AppendArgument(alignment);
+            }
+            else
+            {
+                // Move the alignment constant into the c
```
MoFtZ, line 732: QUESTION: Is this comment incomplete?

MoFtZ (changes requested):

Src/ILGPU/Backends/OpenCL/CLCodeGenerator.Values.cs
m4rs-mt: Nice catch! This will fail to compile on several OpenCL platforms.

MoFtZ (approved):

PR #693, opened 2021-11-20
Added AMD printf OpenCL extension.

MoFtZ:
Experiment to fix a user-reported issue regarding missing printf output on AMD.


MoFtZ:
Only #694 is needed.

PR #697, opened 2021-11-24
Simplified API for copying to/from page locked memory.

### LINKED ISSUE #691: PageLockScope issue
Not sure if it is a bug or a missing feature, but it appears to be impossible for multiple GPUs to use the same pinned memory on host.

Code:
```
            var startTime = DateTime.Now;
            double timeSpent;
            void LogTime(string s) { timeSpent = DateTime.Now.Subtract(startTime).TotalSeconds; Console.WriteLine($"{DateTime.Now}\tTime to {s}: {timeSpent:0.0000}s"); startTime = DateTime.Now; }

            var context = Context.Create().AllAccelerators().EnableAlgorithms().ToContext();
            LogTime("create context with " + context.GetCudaDevices().Count + " gpus");

            var accelerators = Enumerable.Range(0, context.GetCudaDevices().Count).Select(i => context.CreateCudaAccelerator(i)).ToArray();

            int size = 1024;
            var cpuShared = new float[size];
            var intPtr = GCHandle.Alloc(cpuShared, GCHandleType.Pinned).AddrOfPinnedObject();
            var scopes = new PageLockScope<float>[accelerators.Length];
            LogTime("create cpu array");

            for(int accId = 0; accId < accelerators.Length; accId++) {
                scopes[accId] = accelerators[accId].CreatePageLockFromPinned<float>(cpuShared);
                LogTime("create scope for accelerator " + accId);
            }
```

Output:
```
Unhandled exception. ILGPU.Runtime.Cuda.CudaException: part or all of the requested memory range is already mapped
   at ILGPU.Runtime.CudaPageLockScope`1..ctor(CudaAccelerator accelerator, IntPtr hostPtr, Int64 numElements)
   at ILGPU.Runtime.Cuda.CudaAccelerator.CreatePageLockFromPinnedInternal[T](IntPtr pinned, Int64 numElements)
   at ILGPU.Runtime.Accelerator.CreatePageLockFromPinned[T](IntPtr pinned, Int64 numElements)
   at ILGPU.Runtime.Accelerator.CreatePageLockFromPinned[T](T[] pinned)
```
I suspect that CreatePageLockFromPinned() is trying call cudaHostRegister a second time when it does not need to - it only needs to call cudaHostGetDevicePointer for subsequent accelerators.

MoFtZ:
Fixes #691.

Removes the various CopyFrom/CopyToPageLockedAsync methods in favour of the more generic CopyTo/CopyFrom ArrayView methods. This allows copying to/from a subset of the page locked array without adding a lot of extra to methods specifically support it.

`PageLockedScoped` now exposes an `ArrayView` property that can be used to interact with the existing CopyFrom/CopyTo ArrayView methods.

### DIFF
--- a/Src/ILGPU/Runtime/ArrayViewExtensions.Generated.tt
+++ b/Src/ILGPU/Runtime/ArrayViewExtensions.Generated.tt
@@ -712,6 +712,7 @@ namespace ILGPU.Runtime
         /// <param name="pageLockScope">The page locked memory.</param>
         /// <remarks>This method is not supported on accelerators.</remarks>
         [NotInsideKernel]
+        [Obsolete("Use PageLockScope.ArrayView instead")]
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public static void CopyToPageLockedAsync<T>(
             this <#= viewName #> source,
@@ -729,6 +730,7 @@ namespace ILGPU.Runtime
         /// <param name="pageLockScope">The page locked memory.</param>
         /// <remarks>This method is not supported on accelerators.</remarks>
         [NotInsideKernel]
+        [Obsolete("Use PageLockScope.ArrayView instead")]
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public static void CopyToPageLockedAsync<T>(
             this <#= viewName #> source,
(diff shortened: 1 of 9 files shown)


m4rs-mt (approved):

m4rs-mt:
@MoFtZ looks like we have to address some conflicts before merging this feature.

Src/ILGPU/Runtime/ArrayViewExtensions.Generated.tt
```diff
@@ -700,172 +700,6 @@ namespace ILGPU.Runtime
 
         #endregion
 
-        #region Copy to/from Page Locked
```
m4rs-mt, line 703: Question: Given our attempts to maintain backwards compatibility, I was reconsidering these changes. Should we just keep all functions and mark the as `obsolete`?

MoFtZ: Re-added methods and marked as `[Obsolete]`.

m4rs-mt (approved):

PR #703, opened 2021-12-02
Refined LICM transformation to make code movement smarter.

m4rs-mt:
This PR refines the current loop-invariant code motion transformation by making it "smarter". Smarter refers to the fact that constants are not moved out of loops, if all of their uses cannot be moved out of loops.


Src/ILGPU/IR/Transformations/LoopInvariantCodeMotion.cs
```diff
@@ -111,6 +113,66 @@ public LoopInvariance(Loop loop)
...
+            /// </summary>
+            /// <param name="value">The value to check.</param>
+            /// <returns>True, if the given value should be moved.</returns>
+            public readonly bool ShouldBeMoved(Value value)
+            {
+                // Check whether the given value is scheduled to be moved
+                if (!visited.Contains(value))
+                    return false;
+
+                // Check whether this value is a constant and should not be move out of
```
MoFtZ, line 160: MINOR: Spelling.
```suggestion
                // Check whether this value is a constant and should not be moved out of
```

MoFtZ (approved):

PR #708, opened 2021-12-03
Cancel CI test jobs early when one of them fails

jgiannuzzi:
This PR aims to improve the CI feedback loop by cancelling test jobs early when one of them fails (`fail-fast`). This means that e.g. the CPU tests failing would cancel the long-running Cuda tests and free some of our self-hosted runners to run other jobs instead.

Here is a demo of one of the test jobs failing and all the others being cancelled: https://github.com/jgiannuzzi/ILGPU/actions/runs/1535618199

This could be improved further by instructing `dotnet test` to stop after the first failure, but I couldn't find how to do it 😔


MoFtZ:
> This could be improved further by instructing `dotnet test` to stop after the first failure, but I couldn't find how to do it 😔

[Looks like it was added ](https://github.com/xunit/xunit/issues/856#issuecomment-221726401)to the xunit console runner (`dotnet xunit` - now deprecated), and the MSBuild runner, but is not available via `dotnet test`.

MoFtZ (approved):

PR #712, opened 2021-12-10
[CI] Disable parallelization when running clean tests to avoid OutOfMemory issues

ljubon:
Build test job should have default environment variables for master/tags and schedule events, otherwise use `ILGPU_CLEAN_TESTS=false` and `ILGPU_TEST_PARALLELIZATION_LEVEL=4` to debug and test parallelization.

Example:
 [here ](https://github.com/ljubon/ILGPU/runs/4831531492?check_suite_focus=true#step:7:3)is when the condition is `true` (p.s I've changed just `if condition` to match my branch name) and [here ](https://github.com/ljubon/ILGPU/runs/4831531659?check_suite_focus=true#step:8:7)confirmation that jobs `Build and test` afterward used configured values.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -140,10 +140,11 @@ jobs:
         id: env
         shell: bash
         run: |
-          echo "::set-output name=json::{}"
+          jsonValue="{}"
           if [[ ${{ github.event_name == 'schedule' || (github.event_name == 'push' && (github.ref == 'refs/heads/master' || startsWith(github.ref, 'refs/tags/v'))) }} ]]; then
-            echo "::set-output name=json::{\"ILGPU_CLEAN_TESTS\":\"true\", \"ILGPU_TEST_PARALLELIZATION_LEVEL\":\"1\"}"
+            jsonValue="{\"ILGPU_CLEAN_TESTS\":\"true\", \"ILGPU_TEST_PARALLELIZATION_LEVEL\":\"1\"}"
           fi
+          echo "::set-output name=json::$jsonValue"
 
       - name: Build and test
         run: dotnet test --configuration=Release --framework=${{ matrix.framework }} -p:TreatWarningsAsErrors=true --logger GitHubActions Src/${{ matrix.library }}.Tests.${{ steps.test-flavor.outputs.flavor }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -133,10 +133,18 @@ jobs:
         shell: bash
         run: echo ::set-output name=flavor::$([[ "${{ matrix.os }}" == "cuda" ]] && echo "Cuda" || echo "CPU")
 
+      - name: Set environment variables
+        id: env
+        shell: bash
+        run: |
+          echo "::set-output name=json::{\"ILGPU_CLEAN_TESTS\":\"false\", \"ILGPU_TEST_PARALLELIZATION_LEVEL\":\"1\"}"
+          if [[ ${{ github.event_name == 'schedule' || (github.event_name == 'push' && (github.ref == 'refs/heads/master' || startsWith(github.ref, 'refs/tags/v'))) }} ]]; then
+            echo "::set-output name=json::{\"ILGPU_CLEAN_TESTS\":\"true\", \"ILGPU_TEST_PARALLELIZATION_LEVEL\":\"4\"}"
+          fi
+
       - name: Build and test
         run: dotnet test --configuration=Release --framework=${{ matrix.framework }} -p:TreatWarningsAsErrors=true --logger GitHubActions Src/${{ matrix.library }}.Tests.${{ steps.test-flavor.outputs.flavor }}
-        env:
-          ILGPU_CLEAN_TESTS: ${{ github.event_name == 'schedule' || (github.event_name == 'push' && (github.ref == 'refs/heads/master' || startsWith(github.ref, 'refs/tags/v'))) }}
+        env: ${{ fromJson(steps.env.outputs.json) }}
 
   # Ensure that ILGPU libraries are built using the same version tag.
   check-version:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -133,13 +133,16 @@ jobs:
         shell: bash
         run: echo ::set-output name=flavor::$([[ "${{ matrix.os }}" == "cuda" ]] && echo "Cuda" || echo "CPU")
 
+      # Clean tests (kernel caching disabled) use more CPU and memory, so we only run them for
+      # scheduled runs, pushes to master, and pushes to tags.
+      # Parallelization also needs to be disabled to avoid Out of Memory errors.
       - name: Set environment variables
         id: env
         shell: bash
         run: |
-          echo "::set-output name=json::{\"ILGPU_CLEAN_TESTS\":\"false\", \"ILGPU_TEST_PARALLELIZATION_LEVEL\":\"1\"}"
+          echo "::set-output name=json::{}"
           if [[ ${{ github.event_name == 'schedule' || (github.event_name == 'push' && (github.ref == 'refs/heads/master' || startsWith(github.ref, 'refs/tags/v'))) }} ]]; then
-            echo "::set-output name=json::{\"ILGPU_CLEAN_TESTS\":\"true\", \"ILGPU_TEST_PARALLELIZATION_LEVEL\":\"4\"}"
+            echo "::set-output name=json::{\"ILGPU_CLEAN_TESTS\":\"true\", \"ILGPU_TEST_PARALLELIZATION_LEVEL\":\"1\"}"
           fi
 
       - name: Build and test
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -142,7 +142,7 @@ jobs:
         run: |
           jsonValue="{}"
           if [[ ${{ github.event_name == 'schedule' || (github.event_name == 'push' && (github.ref == 'refs/heads/master' || startsWith(github.ref, 'refs/tags/v'))) }} ]]; then
-            jsonValue="{\"ILGPU_CLEAN_TESTS\":\"true\", \"ILGPU_TEST_PARALLELIZATION_LEVEL\":\"1\"}"
+            jsonValue="{\"ILGPU_CLEAN_TESTS\":\"true\"}"
           fi
           echo "::set-output name=json::$jsonValue"
 


.github/workflows/ci.yml
```diff
@@ -155,10 +155,21 @@ jobs:
...
+          if [[ ${{ github.event_name == 'schedule' || (github.event_name == 'push' && (github.ref == 'refs/heads/master' || startsWith(github.ref, 'refs/tags/v'))) }} ]]; then
+            echo ::set-output name=testParallel::1
+          else
+            echo ::set-output name=testParallel::4
+          fi
+
       - name: Build and test
         run: dotnet test --configuration=Release --framework=${{ matrix.framework }} -p:TreatWarningsAsErrors=true --logger GitHubActions Src/${{ matrix.library }}.Tests.${{ steps.test-flavor.outputs.flavor }}
         env:
           ILGPU_CLEAN_TESTS: ${{ github.event_name == 'schedule' || (github.event_name == 'push' && (github.ref == 'refs/heads/master' || startsWith(github.ref, 'refs/tags/v'))) }}
```
jgiannuzzi, line 161: The value for this could be set in the same way that we set the parallelisation level

For example you could set both environment variables as a json string when the 'clean tests' conditions are reached, and then use that as the value for `env`. When the conditions are not reached, no values are set, which means we use the default.

ljubon: I didn't get it from first...but please check the latest version when you can, I think i got it 🙏 👀

.github/workflows/ci.yml
```diff
@@ -155,10 +155,21 @@ jobs:
         shell: bash
         run: echo ::set-output name=flavor::$([[ "${{ matrix.os }}" == "cuda" ]] && echo "Cuda" || echo "CPU")
 
+      - name: Set values
+        id: values
+        shell: bash
+        run: |
+          if [[ ${{ github.event_name == 'schedule' || (github.event_name == 'push' && (github.ref == 'refs/heads/master' || startsWith(github.ref, 'refs/tags/v'))) }} ]]; then
+            jsonValues="{"ILGPU_CLEAN_TESTS":"1", "ILGPU_TEST_PARALLELIZATION_LEVEL":"true"}"
```
MoFtZ, line 163: ISSUE: It looks like JSON may be incorrect. Clean tests should be a boolean, while parallelisation level should be a number.
```suggestion
            jsonValues="{"ILGPU_CLEAN_TESTS":"true", "ILGPU_TEST_PARALLELIZATION_LEVEL":"1"}"
```

ljubon: Thanks, @MoFtZ !

.github/workflows/ci.yml
```diff
@@ -155,10 +155,21 @@ jobs:
...
 
+      - name: Set values
+        id: values
+        shell: bash
+        run: |
+          if [[ ${{ github.event_name == 'schedule' || (github.event_name == 'push' && (github.ref == 'refs/heads/master' || startsWith(github.ref, 'refs/tags/v'))) }} ]]; then
+            jsonValues="{"ILGPU_CLEAN_TESTS":"true", "ILGPU_TEST_PARALLELIZATION_LEVEL":"1"}"
+            echo ::set-output name=json::$jsonValues
+          else
+            jsonValues="{"ILGPU_CLEAN_TESTS":"false", "ILGPU_TEST_PARALLELIZATION_LEVEL":"4"}"
```
jgiannuzzi, line 166: as these are the default values, I think we could afford to set an empty json object instead (`{}`)

.github/workflows/ci.yml
```diff
@@ -155,10 +155,21 @@ jobs:
...
+            echo ::set-output name=json::$jsonValues
```
jgiannuzzi, line 167: this `echo` and the one above should be moved outside of the `if` block

.github/workflows/ci.yml
```diff
@@ -134,10 +134,18 @@ jobs:
...
         run: echo ::set-output name=flavor::$([[ "${{ matrix.os }}" == "cuda" ]] && echo "Cuda" || echo "CPU")
 
+      - name: Set environment variables
+        id: env
+        shell: bash
+        run: |
+          echo "::set-output name=json::{\"ILGPU_CLEAN_TESTS\":\"false\", \"ILGPU_TEST_PARALLELIZATION_LEVEL\":\"1\"}"
+          if [[ ${{ github.event_name == 'schedule' || (github.event_name == 'push' && (github.ref == 'refs/heads/master' || startsWith(github.ref, 'refs/tags/v'))) }} ]]; then
+            echo "::set-output name=json::{\"ILGPU_CLEAN_TESTS\":\"true\", \"ILGPU_TEST_PARALLELIZATION_LEVEL\":\"4\"}"
+          fi
```
MoFtZ, line 144: ISSUE: These settings do not look correct. The initial conversation says to set clean tests to false, and parallelization level to 4.

We might also want to add some comments about why we are doing this, rather than just what we are doing.

```suggestion
          echo "::set-output name=json::{\"ILGPU_CLEAN_TESTS\":\"false\", \"ILGPU_TEST_PARALLELIZATION_LEVEL\":\"4\"}"
          if [[ ${{ github.event_name == 'schedule' || (github.event_name == 'push' && (github.ref == 'refs/heads/master' || startsWith(github.ref, 'refs/tags/v'))) }} ]]; then
            echo "::set-output name=json::{\"ILGPU_CLEAN_TESTS\":\"true\", \"ILGPU_TEST_PARALLELIZATION_LEVEL\":\"1\"}"
          fi
```

MoFtZ (changes requested):

.github/workflows/ci.yml
jgiannuzzi: agreed about both — and as I mentioned [in a previous review comment](https://github.com/m4rs-mt/ILGPU/pull/712#discussion_r768567426), I think the default value should be an empty json object instead, so that we don't need to change the defaults both in the code and in the CI.

```suggestion
          echo "::set-output name=json::{}"
          if [[ ${{ github.event_name == 'schedule' || (github.event_name == 'push' && (github.ref == 'refs/heads/master' || startsWith(github.ref, 'refs/tags/v'))) }} ]]; then
            echo "::set-output name=json::{\"ILGPU_CLEAN_TESTS\":\"true\", \"ILGPU_TEST_PARALLELIZATION_LEVEL\":\"1\"}"
          fi
```

.github/workflows/ci.yml
```diff
@@ -134,10 +134,18 @@ jobs:
         shell: bash
         run: echo ::set-output name=flavor::$([[ "${{ matrix.os }}" == "cuda" ]] && echo "Cuda" || echo "CPU")
 
+      - name: Set environment variables
```
jgiannuzzi, line 137:
```suggestion
      # Clean tests (kernel caching disabled) use more CPU and memory, so we only run them for
      # scheduled runs, pushes to master, and pushes to tags.
      # Parallelization also needs to be disabled to avoid Out of Memory errors.
      - name: Set environment variables
```

ljubon: Thank you both @MoFtZ and @jgiannuzzi for these comments, much appreciated! 🍀 
Since the initial idea happened over a call, my bad was that started immediately working on PR for Ivan's repo instead of writing down those requirements.
I'll update this PR once the change is confirmed and tested by our team.

ljubon: here comes the latest changes, as you suggested 🚀

pavlovic-ivan:
@jgiannuzzi @m4rs-mt how do we feel about this PR? Is it ok to be merged, or are there further changes that need to be made?

jgiannuzzi (commented):
*clicked too soon on the submit button* 🤦

.github/workflows/ci.yml
```diff
@@ -133,10 +133,21 @@ jobs:
...
+      # scheduled runs, pushes to master, and pushes to tags.
+      # Parallelization also needs to be disabled to avoid Out of Memory errors.
+      - name: Set environment variables
+        id: env
+        shell: bash
+        run: |
+          echo "::set-output name=json::{}"
+          if [[ ${{ github.event_name == 'schedule' || (github.event_name == 'push' && (github.ref == 'refs/heads/master' || startsWith(github.ref, 'refs/tags/v'))) }} ]]; then
+            echo "::set-output name=json::{\"ILGPU_CLEAN_TESTS\":\"true\", \"ILGPU_TEST_PARALLELIZATION_LEVEL\":\"1\"}"
+          fi
```
jgiannuzzi, line 146: one last nitpick: maybe we should set a `json` variable on lines 143 and 145, and then `echo` it after the condition?
the current code will `echo` twice when the condition is met, which GitHub Actions seems to be happy about, but it doesn't _feel_ ideal.

jgiannuzzi (commented):
This PR looks good to me and will help improve the CUDA test failures we sometimes encounter. I have one very last comment below, but it's not a blocker for me.

.github/workflows/ci.yml
pavlovic-ivan: @ljubon do you think you can refactor this piece of code and create a final commit? i think we did similar thing elsewhere as well

ljubon: I'm on it, currently verifying...once it's done, I'm pushing here ⚡

ljubon: you can find example workflow here (tested, it works) pushing in min ⚡

jgiannuzzi (approved):
LGTM 👍

Could you please change the PR name to something more descriptive, like “[CI] Disable parallelization when running clean tests to avoid OutOfMemory issues”?

.github/workflows/ci.yml
```diff
@@ -133,10 +133,22 @@ jobs:
...
+      # Clean tests (kernel caching disabled) use more CPU and memory, so we only run them for
+      # scheduled runs, pushes to master, and pushes to tags.
+      # Parallelization also needs to be disabled to avoid Out of Memory errors.
+      - name: Set environment variables
+        id: env
+        shell: bash
+        run: |
+          jsonValue="{}"
+          if [[ ${{ github.event_name == 'schedule' || (github.event_name == 'push' && (github.ref == 'refs/heads/master' || startsWith(github.ref, 'refs/tags/v'))) }} ]]; then
+            jsonValue="{\"ILGPU_CLEAN_TESTS\":\"true\", \"ILGPU_TEST_PARALLELIZATION_LEVEL\":\"1\"}"
```
pavlovic-ivan, line 145: @ljubon latest discussion on the topic is to _not set_ `ILGPU_TEST_PARALLELIZATION_LEVEL`. So this line should be changed to:
```
jsonValue="{\"ILGPU_CLEAN_TESTS\":\"true\"}"
```

pavlovic-ivan (changes requested):
Please check the latest comment in the code

.github/workflows/ci.yml
m4rs-mt: This should work after we merge #741.

jgiannuzzi: This change would essentially become a no-op compared to what we already have today:
https://github.com/m4rs-mt/ILGPU/blob/3e59b1a1c523eb501405e392dc5bd5058e4562ee/.github/workflows/ci.yml#L138-L139

Now that #741 has been merged, I would suggest to close this PR instead. Sorry @ljubon! 😞

m4rs-mt: @jgiannuzzi Do you think it's not worth keeping this PR and its changes for future CI pipeline changes/updates when you push to `master`?

jgiannuzzi: @m4rs-mt I'd suggest we reopen this PR if the need arises (and adapt if necessary)

ljubon: hey guys, as mentioned I've pushed the latest change which we originally wanted
if you all agree, I would proceed with the closure of this PR (as is) just in case we want in future to re-open or add additional work on this PR

PR #714, opened 2021-12-13
Prevent unintentional initialization of Accelerator.Current.

### LINKED ISSUE #713: Division By Zero in LoadAutoGroupedKernel
I just tried out this library for the first time, so there might well be issues with my code, but this produces a DivideByZeroException on my machine:
```cs
using System.Diagnostics.CodeAnalysis;
using ILGPU;
using ILGPU.Runtime;

namespace ILGpu_DivByZero;

class Class
{
    private static void MyKernel(Index1D ix, ArrayView<int> data, int c)
    {
        data[ix] = ix + c;
    }
    
    public static void Main(string[] _)
    {
        // tried the following, all of them produced division by zero:
        // = Accelerator.Current 
        // = CPUAccelerator.Current
        // = context.GetImplicitCPUAccelerator()
        using var accelerator = Accelerator.Current;

        // error on this line!
        var myKernel = accelerator.LoadAutoGroupedKernel<Index1D, ArrayView<int>, int>(MyKernel);

        using var buffer = accelerator.Allocate1D<int>(50);
        myKernel(accelerator.DefaultStream, (Index1D)buffer.Length, buffer.View, 10);
        
        accelerator.Synchronize();

        var data = buffer.GetAsArray1D();

        Console.WriteLine(string.Join('\n', data));
    }
}
```
, and here's the stacktrace: 
```
Unhandled exception. System.DivideByZeroException: Attempted to divide by zero.
   at ILGPU.Runtime.CPU.CPUAccelerator.LoadAutoGroupedKernelInternal(CompiledKernel kernel, KernelInfo& kernelInfo) in ILGPU.dll:token 0x6000f52+0x19
   at ILGPU.Runtime.Accelerator.LoadAutoGroupedKernel(CompiledKernel kernel, KernelInfo& kernelInfo) in ILGPU.dll:token 0x6000791+0x6
   at ILGPU.Runtime.Accelerator.AutoKernelLoader.LoadKernel(Accelerator accelerator, CompiledKernel compiledKernel, KernelInfo& kernelInfo) in ILGPU.dll:token 0x60020bf+0x0
   at ILGPU.Runtime.Accelerator.<>c__DisplayClass157_0`1.<LoadGenericKernel>b__0(TKernelLoader& loader, KernelInfo& info) in ILGPU.dll:token 0x6002101+0x0
   at ILGPU.Runtime.Accelerator.LoadAutoGroupedKernel[TDelegate](MethodInfo method, KernelInfo& kernelInfo) in ILGPU.dll:token 0x60007a0+0x0
   at ILGPU.Runtime.Accelerator.LoadAutoGroupedKernel[TDelegate](MethodInfo method) in ILGPU.dll:token 0x600079f+0x0
   at ILGPU.Runtime.KernelLoaders.LoadAutoGroupedKernel[TIndex,T1,T2](Accelerator accelerator, Action`3 action) in ILGPU.dll:token 0x6000a5d+0xe
   at ILGpu_DivByZero.Class.Main(String[] _) in /home/name/RiderProjects/Bugs/ILGpu_DivByZero/Program.cs:line 27

```

I am almost certain that the issue is `WarpSize` being 0 in [CPUAccelerator#LoadAutoGroupedKernelInternal(CompiledKernel, KernelInfo)](https://github.com/m4rs-mt/ILGPU/blob/ce4845f886a98c05acb8913d28796c8c5dcb9c00/Src/ILGPU/Runtime/CPU/CPUAccelerator.cs#L219) - Is this an issue with my setup or a bug in the library?

MoFtZ:
Fixes #713.

@m4rs-mt Do you have a better suggestion for how to fix this? Is feels a bit hacky..

### DIFF
--- a/Src/ILGPU/Runtime/CPU/CPUAccelerator.cs
+++ b/Src/ILGPU/Runtime/CPU/CPUAccelerator.cs
@@ -90,7 +90,7 @@ namespace ILGPU.Runtime.CPU
         {
             NativePtr = new IntPtr(1);
 
-            DefaultStream = CreateStream();
+            DefaultStream = CreateStreamInternal();
 
             NumThreads = description.NumThreads;
             Mode = mode;
@@ -108,7 +108,8 @@ namespace ILGPU.Runtime.CPU
                     UsesSequentialExecution);
             }
 
-            Bind();
+            if (!description.Equals(CPUDevice.Implicit))
+                Bind();
             Init(context.DefautltILBackend);
         }
 
--- a/Src/ILGPU/Context.cs
+++ b/Src/ILGPU/Context.cs
@@ -260,11 +260,16 @@ namespace ILGPU
                 Properties.InliningMode);
 
             // Initialize the default CPU device
+            // NB: Ensure that the current accelerator is not changed when creating the
+            // implicit CPU accelerator. Otherwise, we may unintentionally initialize
+            // Accelerator.Current before an accelerator has been created by the user.
+            var currentAccelerator = Accelerator.Current;
             CPUAccelerator = new CPUAccelerator(
                 this,
                 CPUDevice.Implicit,
                 CPUAcceleratorMode.Parallel,
                 ThreadPriority.Lowest);
+            Debug.Assert(Accelerator.Current == currentAccelerator);
 
             // Initialize all devices
             Devices = devices;
--- a/Src/ILGPU/Runtime/CPU/CPUDevice.cs
+++ b/Src/ILGPU/Runtime/CPU/CPUDevice.cs
@@ -85,7 +85,7 @@ namespace ILGPU.Runtime.CPU
         /// purposes. Instead, it acts as a placeholder accelerator for buffers that
         /// are implicitly associated with a parent CPU accelerator.
         /// </summary>
-        public static readonly CPUDevice Implicit =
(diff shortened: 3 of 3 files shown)


Src/ILGPU/Runtime/CPU/CPUAccelerator.cs
```diff
@@ -108,7 +108,8 @@ public sealed partial class CPUAccelerator : Accelerator
                     UsesSequentialExecution);
             }
 
-            Bind();
+            if (!description.Equals(CPUDevice.Implicit))
```
m4rs-mt, line 111: I know that the solution feels kind of hacky... However, I think we can get rid of the call to `Bind` anyway... right?

MoFtZ: This call to `Bind` is the one that initializes `Accelerator.Current` when the user calls `context.CreateCPUAccelerator`. There are similar calls in the constructor of the other accelerators.

We could potentially move it out of the constructor, and into the `CreateCPUAccelerator`, `CreateCudaAccelerator` and `CreateCLAccelerator` methods?

MoFtZ: OK 👍

m4rs-mt (approved):

Issue #422, opened 2021-02-18
Can ILGPU give access to the  low-level CUDA functions like nvJPEG ?


Aleksey1555:
here - https://docs.nvidia.com/cuda/nvjpeg/index.html

m4rs-mt:
@Aleksey1555 Thank you for your feature request. This might be an interesting feature for the `Algorithms` library, I think.

m4rs-mt:
@Aleksey1555 We have been discussing about adding new bindings to Algorithms library. @MoFtZ is currently working on great extensions to cover most API bindings from several well-known and prominent Cuda libraries. However, we came to the conclusion that supporting the `nvJPEG` library is too much work at the moment. Perhaps you can elaborate on a specific use case you have in mind?

PR #716, opened 2021-12-15
Added bindings for NvJpeg Single Image Decode.

MoFtZ:
Related to #422.

Initial bindings for NvJpeg with simple wrappers handles that need to be disposed.


Src/ILGPU.Algorithms/Runtime/Cuda/NvJpegLibrary.cs
```diff
@@ -0,0 +1,110 @@
...
+        /// Creates a new NvJpeg state instance.
+        /// </summary>
+        public NvJpegState CreateState()
+        {
+            NvJpegException.ThrowIfFailed(
+                API.JpegStateCreate(LibHandle, out IntPtr stateHandle));
+            return new NvJpegState(API, stateHandle);
+        }
+
+        /// <inheritdoc cref="NvJpegAPI.GetImageInfo(IntPtr, ReadOnlySpan{byte}, out int, out NvJpegChromaSubsampling, out int[], out int[])"/>
```
m4rs-mt, line 67: Question: should we break these comments to ensure that the max-line length will not be exceeded?

Samples/AlgorithmsNvJpeg/Program.cs
```diff
@@ -0,0 +1,93 @@
+﻿using ILGPU;
```
m4rs-mt, line 1: Question: is the copyright header missing?

m4rs-mt (changes requested):
Great work! The only thing I'm afraid of right now is the sample image. I bought a publication license for the website which is not suitable for further publications. Can we replace this image with another one using a Creative Commons license?

Src/ILGPU.Algorithms/Runtime/Cuda/NvJpegLibrary.cs
MoFtZ: Done - didn't realise I could wrap the cref.

Samples/AlgorithmsNvJpeg/Program.cs
MoFtZ: Added.

MoFtZ:
Fixed review comments.
Changed sample image to ILGPU logo 😄

m4rs-mt:
Amazing work! Thanks a lot @MoFtZ 🏅

m4rs-mt (approved):

PR #729, opened 2022-01-21
Require .NET 6 for building and enable package validation

jgiannuzzi:
This PR builds upon #720 to now require .NET 6 and/or Visual Studio 2022 to build the libraries, and enable [package validation](https://docs.microsoft.com/en-us/dotnet/fundamentals/package-validation/overview).

In order to speed up local development, the libraries are now built only for `net6.0` unless running in our CI. As the samples get published with the releases, those will instead be built for the target framework that matches the user's SDK (unless running in our CI, where we want to validate that all target frameworks can be built).

Our `global.json` file was malformed and did not enforce the spirit of #536, which was to require a given version of the .NET SDK. Fixing this means that the .NET 6 SDK has to be setup unconditionally in our CI to guarantee its presence.

Enabling package validation means that we will be able to catch any API incompatibility early and make a conscious decision about what to do. Here is how I think our process could go:
* when our CI fails ([example](https://github.com/jgiannuzzi/ILGPU/actions/runs/1729216659)) because of a package validation error, assess whether the API change is acceptable.
  * if acceptable, add a suppression entry to `CompatibilitySuppressions.xml` (running `dotnet pack` with the `/p:GenerateCompatibilitySuppressionFile=true` flag can generate it automatically).
  * if not acceptable, revert the API change.
* after a stable release, on top of bumping the version number, also bump the baseline version (it's in the [same file](https://github.com/m4rs-mt/ILGPU/blob/master/Src/Directory.Build.props)) and delete the suppressions list.

This PR introduces a suppressions list of 13 x 3 entries (because they are duplicated for all the target frameworks that we built in 1.0.0):
* 1 entry (member `ILGPU.Runtime.CPU.CPUDevice.Implicit` being removed) is a change introduced in #714. As we know that using this property in a program would result in a crash, this API change can be deemed acceptable.
* 9 entries are changes in `ILGPU.IR` which we consider as an internal namespace.
* 3 entries are changes in `ILGPU.Backends` which we also consider internal.


global.json
```diff
@@ -1,5 +1,6 @@
 {
   "sdk": {
-    "version": "6.0.x"
+    "version": "6.0.0",
+    "rollForward": "latestFeature"
```
MoFtZ, line 4: ISSUE: We should probably add `"allowPrerelease": false`.

The [documentation](https://docs.microsoft.com/en-us/dotnet/core/tools/global-json?tabs=netcore3x#globaljson-schema) says that it will default to `true`.


> If you don't set this value explicitly, the default value depends on whether you're running from Visual Studio:
If you're not in Visual Studio, the default value is true.

MoFtZ (changes requested):
This looks really good. Am liking the package validation, although we're going to have to suppress all the internal API changes every time.

global.json
jgiannuzzi: Thanks for catching this, @MoFtZ!
I have added it in the next commit.

m4rs-mt (approved):
Looks really good to me and thanks for working on this 👍 As @MoFtZ said, this means we have to suppress all warnings affecting internal classes. However, this allows us to catch bugs that break API/ABI compatibility early in the development process.

PR #760, opened 2022-02-23
Extended inline PTX to support multiple output and by-ref parameters.

MoFtZ:
Related to https://github.com/MoFtZ/ILGPU.OptiX/issues/2.

The current implementation of `CudaAsm` supports an `Emit` function that accepts up to 10 parameters, with the first parameter allowed to be an output variable. Unfortunately, this does not support more complex scenarios like `optixTrace` which has up to 44 argument, with up to 32 by-reference parameters.

This PR introduces the new `EmitRef` function that accepts up to 44 parameters, and passes all parameters by-reference. Each parameter can be one of `Input<T>`, `Output<T>` or `Ref<T>`, where `T` is `struct` (rather than `unmanaged`) in order to support `bool`.

This should then allow the implementation of `optixTrace`.


m4rs-mt (approved):

MoFtZ:
Resolved conflicts on `Src\ILGPU\CompatibilitySuppressions.xml`, and rebased.

PR #765, opened 2022-03-03
Uniform branches for NVIDIA GPUs

m4rs-mt:
This PR adds a performance-oriented optimization to differentiate between *uniform* and *divergent* branches. Uniform branches are jumps that will be taken by all *active* lanes in a warp in all cases. This PR adds this new performance-driven pass to the O2 pipeline and is designed to reduce register consumption and improve control-flow performance of kernels for NVIDIA GPUs.

### DIFF
--- a/Src/ILGPU/Backends/PTX/PTXBackend.cs
+++ b/Src/ILGPU/Backends/PTX/PTXBackend.cs
@@ -216,19 +216,29 @@ namespace ILGPU.Backends.PTX
 
             GenerateLibDeviceCode(backendContext, builder);
 
+            // Check whether we are running in the O1 or O2 pipeline
+            bool o1Enabled = Context.Properties.OptimizationLevel >= OptimizationLevel.O1;
+            bool o2Enabled = Context.Properties.OptimizationLevel > OptimizationLevel.O1;
+
             // Creates pointer alignment information in the context of O1 or higher
-            var alignments = Context.Properties.OptimizationLevel >= OptimizationLevel.O1
+            var alignments = o1Enabled
                 ? PointerAlignments.Apply(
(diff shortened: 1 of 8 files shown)


Src/ILGPU/IR/Analyses/Uniforms.cs
```diff
@@ -0,0 +1,292 @@
...
+            /// True, if the first and second instances are not the same.
+            /// </returns>
+            public static bool operator !=(ValueInfo first, ValueInfo second) =>
+                !first.Equals(second);
+
+            #endregion
+        }
+
+        /// <summary>
+        /// Stores alignment information of an alignment analysis run.
```
MoFtZ, line 133: MINOR: Is this comment about alignment correct?

Src/ILGPU/Backends/PTX/PTXBackend.cs
```diff
@@ -216,19 +216,28 @@ protected override void Dispose(bool disposing)
 
             GenerateLibDeviceCode(backendContext, builder);
 
+            // Check whether we are running in the O2 pipeline
+            bool o2Enabled = Context.Properties.OptimizationLevel >= OptimizationLevel.O1;
```
MoFtZ, line 220: QUESTION: The original code `Context.Properties.OptimizationLevel >= OptimizationLevel.O1` included `O1`. Are we intentionally moving the alignment from `O1` to `O2`?

m4rs-mt: I think we should distinguish between `O1` and `O2` here. `O2` will contain the newly added uniform branch generation, while `O1` will still include alignment information.

MoFtZ (approved):

PR #751, opened 2022-02-16
Add ephemeral dynamic runners workflow and config

pavlovic-ivan:
**Intention**

The intention of this PR is to provide the project the ability to deploy dynamic runners that will run CUDA jobs, by adding new runners workflow. In order to have the complete feature, Github repo needs to be configured. More about that bellow in the Guide.

**Guide**

This guide explains how to configure Github repo. First, the prerequisits. You need a protected environment. If none exist:

1. Go to `Settings` > `Environments` > `New environment`
2. Give the Environment a name `cuda-test-infra`
3. Select a branch for which this Environment will be available by selecting the option "Selected branches" in the drop down menu of `Deployment branches`.
4. Select "+ Add deployment branch rule"
5. In the popup window enter the name of the main branch (eg, main, master, whichever is the main for your repo)

Now, make sure you have a Personal Access Token (PAT) set. If not:

1. Open Your Profile > `Settings` > `Developer Settings` > `Personal access tokens`
2. Select `Generate new token`
3. Give the token a name
4. Set expiration to "No expiration"
5. Select scopes for the token: `repo: Full control of private repositories`
6. Select `Generate token`
7. Copy the token and save it, it is visible to you only once

After the Environment [exists/is created], add new Environment secrets (provided safely offline), but:

- Select "+ Add Secret" within selected Environment (Environment named `cuda-test-infra`)

**Working example**

A working example can be seen [here](https://github.com/pavlovic-ivan/ILGPU-fork2/actions/runs/1848580697). You will see that CUDA tests are green. Workflow failed on `publish-preview` job, as i don't have perms and secrets needed to make it successful.

When this workflow was created, it created an event that triggered a workflow that deploys the runners. Workflow available [here](https://github.com/pavlovic-ivan/ILGPU-fork2/actions/runs/1848580754). Once the main CI workflow is done, another event triggers a workflow that destroys the created runners (as seen [here](https://github.com/pavlovic-ivan/ILGPU-fork2/actions/runs/1848719253)).

**Further steps before accepting PR**

- [ ] Communicate securely with upstream maintainer. Provide secrets needed to deploy runners to GR GCP project

Co-authored-by: Jonathan Giannuzzi jonathan@giannuzzi.me

PR #770, opened 2022-03-14
Enable dynamic self-hosted CUDA runners

jgiannuzzi:
This PR is based on and supersedes #751 by @pavlovic-ivan.

All the secrets have already been setup offline with @m4rs-mt.

The key differences are:
* labels are created and assigned _per workflow run attempt_
* the repo used to create/destroy GCP instances is under @m4rs-mt's control
* some secrets have been renamed
* a GitHub App owned and setup by @m4rs-mt is used instead of a PAT (this app has `admin:write` access to `m4rs-mt/ILGPU` and nothing else)
* service account email has been removed
* skipped workflows should not create/destroy runners
* npm modules are installed according to `package-lock.json` (through `npm ci`)
* machine image has been updated to the current latest versions of Ubuntu, GitHub Actions Runner, and NVIDIA drivers v470 (see https://github.com/pavlovic-ivan/ephemeral-github-runner-image-gcp/pull/7 for details)
* offline runners are cleaned up after the GCP instances are destroyed (as a fallback)

More information about the changes can be found in https://github.com/pavlovic-ivan/ephemeral-github-runner-gcp/pull/1.

This PR won't be able to run any CUDA tests, as it depends on itself being merged. Once merged, all the CUDA tests will run on the new GCP infrastructure. In case of problem, rolling back this change would make the CI use the current infrastructure again.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,7 +65,7 @@ jobs:
             )
           ) && os+=("macos-latest")
 
-          [ "${{ steps.is-fork.outputs.fork }}" == "false" ] && os+=("cuda")
+          [ "${{ steps.is-fork.outputs.fork }}" == "false" ] && os+=("cuda-${{ github.run_id }}-${{ github.run_attempt }}")
 
           echo "::set-output name=os::$(jq -cn '$ARGS.positional' --args ${os[@]})"
     outputs:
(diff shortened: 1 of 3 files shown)


m4rs-mt (approved):
Great work @jgiannuzzi and @pavlovic-ivan 🚀. 
@jgiannuzzi @pavlovic-ivan Am I right that we can close #751 in favor of this PR?

jgiannuzzi:
> Am I right that we can close #751 in favor of this PR?
Yes, this one supersedes #751 👍

PR #774, opened 2022-03-20
Fixed issue with unsigned nested conversions.

### LINKED ISSUE #772: Arithmetic and conditional errors with (byte) casts
# Summary

Casting `uint`s to `byte`s sometimes causes unexpected behaviour in simple arithmetic and conditional operations. This appears to happen only when the uint in question has its `128` bit set. The output depends on whether it's being built in Debug or Release mode. This has been tested on at least a T4, and with both rc1 and older ILGPU code.

# Code

```
public class MiscClean {
    public static void KernelByteBugs(Index1D index, ArrayView1D<uint, Stride1D.Dense> input, ArrayView1D<uint, Stride1D.Dense> output) {
        for (int i = 0; i < input.Length; i++) {
            byte b = (byte)input[i];
            output[0] += b;
            output[1] += ((byte)input[i]) & 255u;
            output[2] += ((byte)input[i]);
            output[3] += input[i] & 255;
            byte b0 = (byte)(input[i]);
            if (b0 < 128) output[4]++;
            if (((input[i]) & 255) < 128) output[5]++;
        }
    }

    static void Main(string[] args) {
        using var context = Context.Create().Cuda().ToContext();
        using var acc = context.CreateCudaAccelerator(0);
        var input = new uint[1] { 130u };
        var output = new uint[6];
        var gpuInput = acc.Allocate1D(input);
        var gpuOutput = acc.Allocate1D(output);
        acc.LaunchAutoGrouped(KernelByteBugs, new Index1D(1), gpuInput.View, gpuOutput.View);
        gpuOutput.CopyToCPU(output);
        for (int i = 0; i < output.Length; i++) {
            Console.WriteLine($"output[{i}] = {output[i]}");
        }
    }
}
```

```
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <TargetFramework>net5.0</TargetFramework>
    <ApplicationIcon />
    <OutputType>Exe</OutputType>
    <StartupObject />
    <RootNamespace>FurrowUtilities</RootNamespace>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="ILGPU" Version="1.1.0-rc1" />
    <PackageReference Include="ILGPU.Algorithms" Version="1.1.0-rc1" />
    <PackageReference Include="System.Collections.Immutable" Version="6.0.0" />
  </ItemGroup>
</Project>
```

# Output

Expected:
```
output[0] = 130
output[1] = 130
output[2] = 130
output[3] = 130
output[4] = 0
output[5] = 0
```

Actual, in debug mode:
```
output[0] = 130
output[1] = 130
output[2] = 4294967170
output[3] = 130
output[4] = 0
output[5] = 0
```

Actual, in release mode:
```
output[0] = 130
output[1] = 130
output[2] = 4294967170
output[3] = 130
output[4] = 1
output[5] = 0
```
Note the difference on `output[4]`.

# Notes

1. 4294967170 = 0xFFFFFF82 (where 0x82 = 130).
2. The hard-coded value=130 is important: this issue apparently happens iff `(value & 128) != 0`. Behaviour is as expected if that bit isn't set.

MoFtZ:
Fixes #772.

### DIFF
--- a/Src/ILGPU.Tests/ConvertIntOperations.cs
+++ b/Src/ILGPU.Tests/ConvertIntOperations.cs
@@ -1,6 +1,6 @@
 ﻿// ---------------------------------------------------------------------------------------
 //                                        ILGPU
-//                           Copyright (c) 2021 ILGPU Project
+//                        Copyright (c) 2021-2022 ILGPU Project
 //                                    www.ilgpu.net
 //
 // File: ConvertIntOperations.cs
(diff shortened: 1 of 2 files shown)


m4rs-mt (approved):

m4rs-mt:
I can confirm that this problem does not occur when unsigned shorts are used. It looks like the problem is related to a different set of instructions issued by the C# compiler. In any case, this fix should cover those cases as well.

PR #784, opened 2022-04-02
Fixed issues with LibDevice integration.

### LINKED ISSUE #781: Meet an error in JN function.
@MoFtZ 
I have try the build-in function in cuda library.

Abs,Cos, Sin, Exp,J0,Y0 are all OK.

But JN is not OK.

The error message is shown in the below figure.

My code is in the attachment.

Can you help me to check my code?
 
image

[Program.zip](https://github.com/m4rs-mt/ILGPU/files/8395603/Program.zip)

MoFtZ:
Fixes #781.

The issue with JN is that we are generating the wrong NVVM/LLVM code, used for extracting the PTX. It does not support the .NET type names of `int`, `long`, `uint` and `ulong`. Instead, we use the types `i32` and `i64`.

Upon further testing, also discovered an issue with `Half` type. The generated PTX uses the `b32` register, and the NVVM code uses the `i16` type.

@m4rs-mt I have based this commit on the ILGPU v1.1.0 tag, in case we wanted to do create an ILGPU v1.1.1 patch release.

### DIFF
--- a/Src/ILGPU/Backends/PTX/PTXLibDeviceMethods.tt
+++ b/Src/ILGPU/Backends/PTX/PTXLibDeviceMethods.tt
@@ -1,6 +1,6 @@
 ﻿// ---------------------------------------------------------------------------------------
 //                                        ILGPU
-//                           Copyright (c) 2021 ILGPU Project
+//                        Copyright (c) 2021-2022 ILGPU Project
 //                                    www.ilgpu.net
 //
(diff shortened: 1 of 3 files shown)


m4rs-mt (approved):

m4rs-mt:
@MoFtZ Thanks for working on this issue and for creating this PR based on the `v1.1` tag. Since we are planning to release `v1.2-beta1` soon, I think it should be ok to include this fix in the next release.

PR #799, opened 2022-04-22
Improve CodeQL analysis speed

jgiannuzzi:
This PR refactors the CodeQL analysis workflow to speed it up from an average of 45 minutes to an average of less than 15 minutes.

The main change is to build all solutions 'manually' because CodeQL autobuild was the main sluggishness culprit. We also build for one TFM instead of all by unsetting `GITHUB_ACTIONS` to speed things up even further.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,14 +1,3 @@
-# For most projects, this workflow file will not need changing; you simply need
-# to commit it to your repository.
-#
-# You may wish to alter this file to override the set of languages analyzed,
-# or to provide custom queries or build logic.
-#
-# ******** NOTE ********
-# We have attempted to detect the languages in your repository. Please check
-# the `language` matrix defined below to confirm you have the correct set of
-# supported CodeQL languages.
-#
 name: "CodeQL"
 
 on:
@@ -31,14 +20,6 @@ jobs:
       contents: read
       security-events: write
 
-    strategy:
-      fail-fast: false
-      matrix:
-        language: [ 'csharp' ]
-        # CodeQL supports [ 'cpp', 'csharp', 'go', 'java', 'javascript', 'python' ]
-        # Learn more:
-        # https://docs.github.com/en/free-pro-team@latest/github/finding-security-vulnerabilities-and-errors-in-your-code/configuring-code-scanning#changing-the-languages-that-are-analyzed
-
     steps:
     - name: Checkout repository
       uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


MoFtZ:
SUGGESTION: Should we run CodeQL in parallel? One for each TFM.

jgiannuzzi:
> SUGGESTION: Should we run CodeQL in parallel? One for each TFM.

That's an interesting idea. I'm experimenting with CodeQL to see if we can actually parallelise it. My current guess is that it has to run once for the entire codebase. I'm also experimenting to see if executing CodeQL for one TFM vs for all TFMs makes a difference in terms of analysis completeness.

jgiannuzzi:
> SUGGESTION: Should we run CodeQL in parallel? One for each TFM.

Sadly CodeQL does not properly support parallelisation. Results seem to be uploaded per commit and are not additive, so with multiple jobs, only the last one to report the results will be taken into account.

The analysis for code branches that are only taken in a given TFM does result in differences between executing CodeQL for one TFM vs for all TFMs, but not by much. I think that we could strike a good balance between usability/speed and completeness by having CodeQL run only for the default TFM on pull requests, and for all TFMs on pushes to `master` and during the weekly scheduled run. Though maybe pushes to `master` should also aim for usability/speed rather than completeness, and we should only have the complete analysis done weekly?

MoFtZ:
The documentation for CodeQL mentions [building in parallel](https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/troubleshooting-the-codeql-workflow#use-matrix-builds-to-parallelize-the-analysis) to speed up the workflow.

Upon further reading, it looks like we should try specifying a custom category - the results may be overridden because we are using the same category for each TFM. This would be how the multiple languages are parallelized.

"[You can upload multiple SARIF files for the same commit, and display the data from each file as code scanning results. When you upload multiple SARIF files for a commit, you must indicate a "category" for each analysis](https://docs.github.com/en/code-security/code-scanning/integrating-with-code-scanning/sarif-support-for-code-scanning?learn=code_security_integration&learnProduct=code-security#about-sarif-support)".

jgiannuzzi:
Thanks for pointing this out @MoFtZ!

I experimented further, and parallelisation is indeed working fine, with one caveat which was the source of my confusion: the summary of the latest scan in the 'Code scanning' tab only shows the latest job to upload results, which are thus partial. See the example below where 3 alerts were purposefully introduced in different TFMs and detected in the same parallel run, but the summary only shows 1, and with the wrong number of lines scanned.

image

About the custom category, it turns out it's done automatically by `github/codeql-action/analyze` by appending the `matrix` content to the job name.

MoFtZ (approved):

PR #818, opened 2022-06-20
Docs: debugging .IOOperations() clarification

KosmosisDire:
Added a clarification to the `PrintF-Like Debugging` section. `Interop.Write()` and `Interop.WriteLine()` do nothing if the `Accelerator` is not built with the `.IOOperations()` builder function. This is not well clarified in the `SimpleWriteLine` sample nor the docs, so I think it's a worthwhile contribution.

I copied a bunch of the wording from some of the other similar parts of the page. But since this directly pertains to the last section I thought it would make most sense down there. Please make any changes to be more concise or accurate.

### DIFF
--- a/Docs/Debugging-and-Profiling.md
+++ b/Docs/Debugging-and-Profiling.md
@@ -29,5 +29,4 @@ This string can be formatted to `Test output 1.0 and test output -45` using:
 ```c#
 Interop.Write("Test output {0} and test output {1}", 1.0, -45);
 ```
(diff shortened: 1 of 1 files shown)


MoFtZ (commented):
Also @KosmosisDire, could you please read and acknowledge https://github.com/m4rs-mt/ILGPU/issues/598

Thanks.

KosmosisDire:
I think that does sound better. I added a new commit with that change.

PR #821, opened 2022-06-24
Added script to generate compatibility suppression files.

MoFtZ:
This script is to help me, since I can never remember the exact command and have to keep looking it up 😄


m4rs-mt (approved):
Great job! This really comes in handy!

PR #822, opened 2022-06-24
Fixed issue promoting unsigned values on the evaluation stack.

### LINKED ISSUE #819: Implicit conversion operation does not correctly handle unsigned values.
A member of the community has found an issue in ILGPU that appears related to the handling of unsigned values.

In particular, the line `var outputValue = inputView[index] > threshold ? inputView[index] : default;` will fail, but changing `default` to `0` will work.

The difference is that `default` is treated by the compiler as `byte` (`UInt8`), and `0` is treat as `int` (`Int32`) - affecting the data type of `outputValue`.

When `outputValue` is an `int`, no conversion is required.
When `outputValue` is a `byte`, the assignment `outputView[index] = lhs;` has an implied conversion from `byte` to `int`. ILGPU is incorrectly treating the `byte` as `Int8` rather than `UInt8`.

```CSharp
class Program
{
    static void TestKernel(Index1D index, byte threshold, ArrayView<byte> inputView, ArrayView<int> outputView)
    {
        // Changing 'default' to '0' will workaround the problem.
        var outputValue = inputView[index] > threshold ? inputView[index] : default;
        outputView[index] = outputValue
    }

    static void Main()
    {
        var values = Enumerable.Range(223, 1).Select(x => (byte)x).ToArray();

        // Create main context

        // For each available device...
        {
            // Create accelerator for the given device
            Console.WriteLine($"Performing operations on {accelerator.AcceleratorType}");

            var kernel = accelerator.LoadAutoGroupedStreamKernel<Index1D, byte, ArrayView<byte>, ArrayView<int>>(TestKernel);
            using var input = accelerator.Allocate1D(values);
            using var output = accelerator.Allocate1D<int>(values.Length);

            output.MemSetToZero();
            kernel((int)input.Length, 0, input.View, output.View);

            var result = output.GetAsArray1D();
            Console.WriteLine($"Expecting = 223, Actual = {result[0]}");
        }
    }
}
```
```
Performing operations on CPU
Expecting = 223, Actual = 223
Performing operations on OpenCL
Expecting = 223, Actual = -33
Performing operations on Cuda
Expecting = 223, Actual = -33
```

MoFtZ:
Fixes #819.

Also removed the unused/obsolete flags `ConvertFlags.Overflow` and `ConvertFlags.OverflowSourceUnsigned`.

This PR depends on #820 to reset the compatibility suppression files.


m4rs-mt (approved):
Thanks a lot for this patch 🥇

PR #841, opened 2022-09-03
Added Generated Files For SPIR-V Builder

Yey007:
Hello!

Sorry for the delay on this PR, but I think I'm finally done. There are, however, a few minor issues I want to ask about:

1. Naming conventions: because this is generated code, they're hard to follow. I can do my best, but I don't know if it's worth the added complexity in the generator.
2. There are a few performance concerns that we may want to address in the future depending on whether this is going to be fast enough for our needs. For example, in `SPIRVWord.cs` there is an extra allocation here that could be avoided by using `Marshal`
```diff
- return new SPIRVWord(BitConverter.ToUInt32(bytes.ToArray(), 0));
+ return new SPRIVWord(MemoryMarshal.Read<int>(bytes));
```
but there is a readability tradeoff. There are other minor things like this that can be improved if necessary.
3. Finally, the signature of `Merge` for builders currently looks like this:
```cs
void Merge(ISPIRVBuilder other);
```
which is not type safe, as the compiler will allow a string builder to be merged with a binary builder. Can anyone think of a way of making this type safe without having to resort to `ISPRIVBuilder` generic over the implementing type, which would require specifying the builder we want to use everywhere and throw away the advantages of using an interface?

Thanks for reviewing!


Yey007:
Oops, looks like some of the Site files got added to this commit somehow. I'll fix that.

Yey007:
Question: What should I do about the Rider files that got committed? I'm guessing we don't want those.

Yey007:
Hi! Sorry for the late reply.

To add SPIR-V as a backend, you would probably need to implement a [Backend](https://github.com/m4rs-mt/ILGPU/blob/c3af8f368445d8e6443f36c99488fc601adcd737/Src/ILGPU/Backends/Backend.cs#L93) using the SPIR-V builder that is introduced in this PR. The existing OpenCL and PTX backends may be good to look at for inspiration---they both may have relevant pieces. I imagine the SPIR-V backend might end up being more similar to the PTX backend just because it's also a bytecode, but I'm not sure which language SPIR-V has more in common with. I would also check out [the SPIR-V specification](https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html).

The other piece that would probably be needed is a Vulkan runtime. Again, the CUDA and OpenCL runtimes are probably a good reference, though I don't know much about Vulkan at all, so I can't say how similar the process would be.

m4rs-mt:
Closing this PR as we're shifting our priorities to focus on other aspects of the project. Thank you very much for your time
and effort on this, we really appreciate it! We'll keep an eye out for opportunities to revisit SPIR-V in the future, so don't lose interest - we might be reaching back out if things change.

Yey007:
Thanks for your help on this-I really appreciate the time you put into reviewing this PR and previous versions. I'd love to contribute more in the future!

PR #842, opened 2022-09-04
Adding new ILGPU site

naskio:
This PR includes:
- Refactoring `Docs/` folder in order to match the naming convention.
- Adding the Jekyll website inside `Site/` folder.
- Adding the workflow `deploy-site.yml` to deploy it to GitHub Pages.

### DIFF
--- a/Src/Directory.Build.props
+++ b/Src/Directory.Build.props
@@ -8,7 +8,7 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <LibraryVersionPrefix>1.3.0-beta1</LibraryVersionPrefix>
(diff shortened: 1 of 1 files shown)


Site/_config.yml
```diff
@@ -0,0 +1,58 @@
+url: "https://m4rs-mt.github.io"
+baseurl: "/ILGPU"
```
naskio, line 2: Should be updated in order to use the domain: http://ilgpu.net/

m4rs-mt: I think we should point to https://www.ilgpu.net/

naskio: Yes, but we will need the to configure the DNS as well (CNAME file)

PR #851, opened 2022-09-17
Fix formatting on code samples in docs

TortillaZHawaii:
www.ilgpu.net does not always correctly render C# code.  It looks like the issue is using `C#` instead of `c#` in parts of the docs. This is especially visible in `Docs/Tutorial_01.md`:
- correct coloring when using lowercase `c#`:
image
- only rendering monospace font without coloring when using uppercase `C#`: 
image


MoFtZ (approved):
hi @TortillaZHawaii, thanks for submitting this PR.
We should probably also update the website to support uppercase `C#`, but this will do for now.

PR #861, opened 2022-10-11
Fix small wiki error

TriceHelix:
The amount of memory allocated should now be equal like described. It also matches https://github.com/m4rs-mt/ILGPU/blob/b7ed379375cc72df6744cd311fab1cae4c4b1933/Samples/DynamicSharedMemory/Program.cs#L80


m4rs-mt (approved):
Nice catch!

PR #871, opened 2022-10-24
Skip build notifications on forks.

MoFtZ:
The CI pipeline for the build notifications is failing on forks, because the webhook secrets have not been configured.

Do we need build notifications on forks? Alternatively, we could change the condition to check whether secrets have been configured.


m4rs-mt:
I think that we do not need the build notifications for forks.

m4rs-mt (approved):

PR #894, opened 2022-12-02
Added support for net70.

MoFtZ:
This PR adds support for compiling ILGPU using the .NET 7.0 SDK.

The CI pipeline has been modified to use the newer SDK. The `dotnet` tool appears to have swapped the order of some arguments, so this has been adjusted.

Updated the C# language version to 9, to fix an issue with `IntPtr` - it is now treated as a `nint`, which only became available in C# 9.

Fixed various compiler warnings. In particular, our custom `DisposeAcceleratorObject` function is not recognised by the Dispose pattern.

Fixed issues with the `in` parameter modifier. The mean appears to have changed. Previously, it behaved more like "here is a read-only struct". We would copy the struct when it was used. Now, the "this is passed by reference" is being enforced. So we have modified ILGPU to remove the `in` parameter modified, and to copy the struct as the argument.

Fixed an issue with the `ref` parameter modifier. This appears to be more strictly enforced. We used to return a reference to a local variable, but now that is no longer permitted.

Finally, fixed an issue with `Unsafe.As`. This has become a .NET intrinsic, and no longer provides MSIL code. We have had to replicate it's behavior within ILGPU.

### DIFF
--- a/Samples/MonitorProgress/CudaProgress.cs
+++ b/Samples/MonitorProgress/CudaProgress.cs
@@ -13,6 +13,7 @@ using ILGPU;
 using ILGPU.Runtime;
 using ILGPU.Runtime.Cuda;
 using System;
+using System.Diagnostics.CodeAnalysis;
 
 #pragma warning disable CA2213
 
@@ -83,6 +84,10 @@ namespace MonitorProgress
         /// <summary>
         /// Holds the Cuda memory buffer that contains the progress value.
         /// </summary>
+        [SuppressMessage(
+            "Microsoft.Usage",
+            "CA2213: Disposable fields should be disposed",
+            Justification = "This is disposed in DisposeAcceleratorObject")]
         private readonly CudaProgressMemoryBuffer memoryBuffer;
 
         /// <summary>
--- a/Src/ILGPU/ILGPU.csproj
+++ b/Src/ILGPU/ILGPU.csproj
@@ -50,7 +50,6 @@
     </PackageReference>
     <PackageReference Include="System.Collections.Immutable" Version="7.0.0" />
     <PackageReference Include="System.Memory" Version="4.5.5" />
-    <PackageReference Include="System.Runtime.CompilerServices.Unsafe" Version="6.0.0" />
     <PackageReference Include="System.Reflection.Metadata" Version="7.0.0" />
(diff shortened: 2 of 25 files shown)


Samples/MonitorProgress/CudaProgress.cs
```diff
@@ -13,6 +13,7 @@
 using ILGPU.Runtime;
 using ILGPU.Runtime.Cuda;
 using System;
+using System.Diagnostics.CodeAnalysis;
 
 #pragma warning disable CA2213
```
m4rs-mt, line 18: I think then we can remove the global oppression, right?

Src/ILGPU/ILGPU.csproj
```diff
@@ -50,7 +50,6 @@
     </PackageReference>
     <PackageReference Include="System.Collections.Immutable" Version="7.0.0" />
     <PackageReference Include="System.Memory" Version="4.5.5" />
-    <PackageReference Include="System.Runtime.CompilerServices.Unsafe" Version="6.0.0" />
```
m4rs-mt, line 53: Is this really not required for legacy-builds using older SDKs?

MoFtZ: It is a transitive dependency of ` System.Collections.Immutable` and `System.Memory`.

In net70, `System.Collections.Immutable` no longer depends on `System.Runtime.CompilerServices.Unsafe` (there is no package for net70).

QUESTION: Why do we depend on these "System" nuget packages? And not just use the version provided in the .NET framework?

Samples/MonitorProgress/CudaProgress.cs
MoFtZ: Yes, it should be possible to remove the warning suppression.

Src/ILGPU/ILGPU.csproj
MoFtZ: It looks like `System.Collections.Immutable` and `System.Reflection.Metadata` are required for `net471` and `netstandard2.1`. For the others (`net50`, `net60`, `net70`), we can omit the nuget packages and compile without issue.

`System.Memory` and `System.Runtime.CompilerServices.Unsafe` are transitive dependencies of `System.Collections.Immutable`, and can be excluded.

m4rs-mt (approved):
Thank you for this brilliant work 🥇.

PR #897, opened 2022-12-05
Updated to conditional nuget packages.

MoFtZ:
Since updating the Nuget packges to the latest version, compiling ILGPU now generates the warning: `System.Collections.Immutable 7.0.0 doesn't support net5.0 and has not been tested with it. Consider upgrading your TargetFramework to net6.0 or later. You may also set <SuppressTfmSupportBuildWarnings>true</SuppressTfmSupportBuildWarnings> in the project file to ignore this warning and attempt to run in this unsupported configuration at your own risk.`

This PR removes the nuget packages on target frameworks where they are not required.


m4rs-mt (approved):
Nice work 👍 !

PR #902, opened 2022-12-14
Fixed compilation warnings from updated net7.0 SDK.

MoFtZ:
The .NET 7.0 SDK was updated from 7.0.100 to 7.0.101. This has introduced a number of new warnings that has caused our CI pipeline to fail.


m4rs-mt (approved):
Nice catch!

PR #924, opened 2023-02-06
Added new specialized sparse-matrix extensions.

m4rs-mt:
This PR adds new helper classes and methods to process sparse matrices on the GPU while supporting specific masking operations to avoid computation of intermediate values. It also adds add a newly created `ConcurrentStreamProcessor` to perform generic operations concurrently on multiple concurrent GPU streams.

A sample use case is shown below that multiplies masked sparse matrixes. There are two options available:

1. Create a single-stream-focused masked multiplier to have explicit control over matrix multiplications
2. Create a stream-based concurrent processor to perform batched multiplications

```c#
// Create a single-streamed sparse matrix buffer
var processor = accel.CreateSparseTransposedMatrixMultiplierMasked<
    float, // Matrix type
    FloatEpsPredicate<Stride2D.DenseY>, // A generic predicate - a float-eps-value predicate in this case
    Stride2D.DenseY, // Custom striding of the matrix to use
    FloatMaskedSparseMatrixProcessor>(); // A float-specific FMA operation to accumulate dot products

// Multiply a single masked sparse matrix
processor(
    accel.DefaultStream,
    new(denseMatrixBuffer.View, 2.0f),
    denseMatrixBuffer.View,
    sparseView, // Sparse matrix view, obtained by explicit construction or using one of the available sparsification methods; will be transposed during runtime (see below)
    outBuffer.View);

// Create a concurrent processor that supports batched multiplication
using var maskedProcessor = new MaskedMatrixProcessor<
    float,
    FloatEpsPredicate<Stride2D.DenseY>,
    Stride2D.DenseY,
    FloatMaskedSparseMatrixProcessor>(accel);
maskedProcessor.Predicate = /* custom predicate here */;
maskedProcessor.MultiplyBatchedTransposed(
    accel.DefaultStream, // The main accelerator stream to attach the next multiplication operations to
    aViewList, // A list of dense input matrices used to multiply the sparse matrices (bViewList) with
    bViewList, // A list of sparse input matrices (will be transposed)
    outViewList); //A list of output views to matrix data containing the results 
```

Sample for accelerated creation of huge sparse matrices on accelerators using the newly added extensions:
```c#
// Setup sparse 2D matrix
var matrix = new float[4, 4]
{
    { 1, 0, 0, 0 },
    { 2, 1, 0, 0 },
    { 3, 0, 1, 0 },
    { 4, 0, 0, 1 },
};

// Allocate basic matrix on the accelerator and transfer it to the device
using var matrixBuffer = accel.Allocate2DDenseY<float>(matrix.GetExtent());
matrixBuffer.View.CopyFromCPU(matrix);

// Allocate a temp buffer (or use existing memory from somewhere else)
using var tempBuffer = accel.Allocate1D<int>(1);

// Initialize the basic shape converter and data converters
var shapeConverter = accel.CreateSparseTransposedMatrixShapeConverter<
    float,
    FloatEpsPredicate<Stride2D.DenseY>,
    Stride2D.DenseY>(tempBuffer.View);
var converter = accel.CreateSparseMatrixConverter<float, Stride2D.DenseY>();

// Get basic shape of the sparse matrix living on the device
using var numNeighborsBuffer = accel.Allocate1D<int>(matrix.GetLength(1));
var shapeView = shapeConverter(
    accel.DefaultStream,
    matrixBuffer,
    new(matrixBuffer, 0.0f),
    numNeighborsBuffer.View,
    maxNumNeighbors =>
        // Allocate a shape-view buffer to store the neighbor lists
        accel.Allocate2DDenseY<int>((matrix.GetLength(1), maxNumNeighbors)));

// Allocate the actual sparse data buffer
using var dataBuffer = accel.Allocate2DDenseY<float>(
    (matrix.GetLength(1),
    shapeView.MaxNonZeroEntries));

// Convert data and fill our sparse matrix structure
var sparseView = converter(accel.DefaultStream, matrixBuffer, shapeView, dataBuffer);
```

*Note: This PR requires PR #989 to be merged.*

Co-authored by @corwinjoy who wrote the initial POC.

### DIFF
--- a/Src/ILGPU.Algorithms.Tests/MatrixTests.cs
+++ b/Src/ILGPU.Algorithms.Tests/MatrixTests.cs
@@ -1,0 +1,275 @@
+// ---------------------------------------------------------------------------------------
+//                                   ILGPU Algorithms
+//                           Copyright (c) 2023 ILGPU Project
+//                                    www.ilgpu.net
+//
+// File: MatrixTests.cs
+//
+// This file is part of ILGPU and is distributed under the University of Illinois Open
+// Source License. See LICENSE.txt for details.
+// ---------------------------------------------------------------------------------------
+
+using ILGPU.Algorithms.MatrixOperations;
+using ILGPU.Runtime;
+using ILGPU.Tests;
+using System;
+using Xunit;
+using Xunit.Abstractions;
+
+#pragma warning disable CA1814 // Jagged arrays
+#pragma warning disable CA5394 // Insecure RNG
+
+namespace ILGPU.Algorithms.Tests
+{
+    public abstract partial class MatrixTests : TestBase
+    {
+        protected MatrixTests(ITestOutputHelper output, TestContext testContext)
+            : base(output, testContext)
+        { }
+
+        #region MemberData
+
+        public static TheoryData<object, object, object, object, object> DimensionsData =>
+            new TheoryData<object, object, object, object, object>
+            {
+                { 39, 42, 17, 2918291, 0.8f },
+                { 39, 42, 17, 2918291, 0.5f },
+                { 39, 42, 17, 2918291, 0.05f },
+
+                { 42, 42, 31, 2918292, 0.8f },
+                { 42, 42, 31, 2918292, 0.5f },
+                { 42, 42, 31, 2918292, 0.05f },
+
+                { 376, 382, 288, 4132107, 0.8f },
+                { 376, 382, 288, 4132107, 0.5f },
+                { 376, 382, 288, 4132107, 0.05f },
+
+                { 829, 277, 928, 31821912, 0.8f },
+                { 829, 277, 928, 31821912, 0.5f },
+                { 829, 277, 928, 31821912, 0.05f },
+
+                { 829, 829, 1121, 31821913, 0.8f },
+                { 829, 829, 1121, 31821913, 0.5f },
+                { 829, 829, 1121, 31821913, 0.05f },
+            };
+
+        #endregion
+
+        #region Helpers
+
+        /// <summary>
+        /// Multiplies two dense matrices and returns the resultant matrix in 2D.
+        /// </summary>
+        /// <param name="left">A dense MxK matrix</param>
+        /// <param name="right">A dense KxN matrix</param>
+        /// <returns>A dense MxN matrix</returns>
+        private static float[,] MultiplyMatrix2D(float[,] left, float[,] right)
+        {
+            var leftRows = left.GetLength(0);
+            var leftColumns = left.GetLength(1);
+            var rightRows = right.GetLength(0);
+            var rightColumns = right.GetLength(1);
+
+            if (leftColumns != rightRows)
+            {
+                throw new ArgumentOutOfRangeException(
+                    nameof(right),
+                    $"Cannot multiply {leftRows}x{leftColumns} matrix by " +
+                    $"{rightColumns}x{rightRows} matrix");
+            }
+
+            var result = new float[leftRows, rightColumns];
+            for (var x = 0; x < leftRows; x++)
+            {
+                for (var y = 0; y < rightColumns; y++)
+                {
+                    for (var z = 0; z < leftColumns; z++)
+                        result[x, y] += left[x, z] * right[z, y];
+                }
+            }
+
+            return result;
+        }
+
+        /// <summary>
+        /// Compute the transpose of a matrix in 2D.
+        /// </summary>
+        /// <param name="matrix">A MxN matrix</param>
+        /// <returns>The transpose of A, a NxM matrix</returns>
+        private static float[,] TransposeMatrix2D(float[,] matrix)
+        {
+            int rows = matrix.GetLength(0);
+            int columns = matrix.GetLength(1);
+            var transposed = new float[columns, rows];
+            for (var i = 0; i < rows; i++)
+            {
+                for (var j = 0; j < columns; j++)
+                    transposed[j, i] = matrix[i, j];
+            }
+
+            return transposed;
+        }
+
+        /// <summary>
+        /// Compares two matrices for equality.
+        /// </summary>
+        /// <param name="left">A MxN matrix (the actual matrix we got)</param>
+        /// <param name="right">A MxN matrix (the matrix we expected) </param>
+        /// <returns>True if the matrices are equal</returns>
+        private static void AssertMatrixEqual2D(
+            float[,] left,
+            float[,] right,
+            float eps = 0.0015f)
+        {
+            var leftRows = left.GetLength(0);
+            var leftColumns = left.GetLength(1);
+            var rightRows = right.GetLength(0);
+            var rightColumns = right.GetLength(1);
+
+            if (leftRows != rightRows || leftColumns != rightColumns)
+            {
+                throw new ArgumentOutOfRangeException(
+                    nameof(right),
+                    $"Matrix dimensions {leftRows}x{leftColumns} and " +
+                    $"{rightRows}x{rightColumns} do not match");
+            }
+
+            for (var i = 0; i < leftRows; i++)
+            {
+                for (var j = 0; j < leftColumns; j++)
+                {
+                    Assert.True(
+                        Math.Abs(left[i, j] - right[i, j]) < eps,
+                        "Matrix result not equal");
+                }
+            }
+        }
+
+        #endregion
+
+        [SkippableTheory()]
+        [MemberData(nameof(DimensionsData))]
+        public void MatrixMultiplySparse(
+            int length,
+            int length2,
+            int baseLength,
+            int seed,
+            float sparsityProbability)
+        {
+            // CPU-performance related checks to avoid extremely long test runs
+            Skip.If(
+                Accelerator.AcceleratorType == AcceleratorType.CPU &&
+                (length > 512 || length2 > 512 || baseLength > 384));
+
+            using var stream = Accelerator.CreateStream();
+
+            // Our starting dense matrix
+            var denseMatrix = new float[baseLength, length];
+
+            // Setup sparse 2D matrix
+            var sparseMatrix = new float[length2, length];
+
+            var random = new System.Random(seed);
+            for (int i = 0; i < denseMatrix.GetLength(0); ++i)
+            {
+                for (int j = 0; j < denseMatrix.GetLength(1); ++j)
+                    denseMatrix[i, j] = (float)random.NextDouble();
+            }
+
+            for (int i = 0; i < sparseMatrix.GetLength(0); ++i)
+            {
+                for (int j = 0; j < sparseMatrix.GetLength(1); ++j)
+                {
+                    sparseMatrix[i, j] = random.NextDouble() > sparsityProbability
+                        ? 0.0f
+                        : (float)random.NextDouble() * 8.0f;
+                }
+            }
+
+            // Setup the output mask
+            var sparseMaskMatrix = new float[baseLength, length2];
+            for (int i = 0; i < sparseMaskMatrix.GetLength(0); ++i)
+            {
+                for (int j = 0; j < sparseMaskMatrix.GetLength(1); ++j)
+                    sparseMaskMatrix[i, j] = 1.0f;
+            }
+
+            // Initialize our sparse matrix
+            using var matrixBuffer = Accelerator.Allocate2DDenseY<float>(
+                sparseMatrix.GetExtent());
+            matrixBuffer.View.CopyFromCPU(stream, sparseMatrix);
+
+            // Allocate a temp buffer (or use existing memory from somewhere else)
+            using var tempBuffer = Accelerator.Allocate1D<int>(1);
+
+            // Initialize the basic shape converter and data converters
+            var shapeConverter = Accelerator.CreateSparseMatrixShapeConverter<
+                float,
+                FloatEpsPredicate<Stride2D.General>,
+                Stride2D.General>(tempBuffer.View);
+            var converter = Accelerator.CreateSparseMatrixConverter<
+                float,
+                Stride2D.General>();
+
+            // Get basic shape of the sparse matrix living on the device
+            using var numNeighborsBuffer = Accelerator.Allocate1D<int>(
+                sparseMatrix.GetLength(0));
+            var shapeView = shapeConverter(
+                stream,
+                matrixBuffer.View.AsGeneral(),
+                new(matrixBuffer.View.AsGeneral(), 0.0f),
+                numNeighborsBuffer.View,
+                maxNumNeighbors =>
+                    // Allocate a shape-view buffer to store the neighbor lists
+                    Accelerator.Allocate2DDenseY<int>(
+                            (sparseMatrix.GetLength(0), maxNumNeighbors))
+                        .View.AsGeneral());
+
+            // Allocate the actual sparse data buffer
+            using var dataBuffer = Accelerator.Allocate2DDenseY<float>(
+                (sparseMatrix.GetLength(0), shapeView.Neighbors.Extent.Y));
+
+            // Convert data and fill our sparse matrix structure
+            var sparseView = converter(
+                stream,
+                matrixBuffer.View.AsGeneral(),
+                shapeView,
+                dataBuffer.View.AsGeneral());
+
+            // Allocate dense output matrix
+            using var aMatrixBuffer =
+                Accelerator.Allocate2DDenseY<float>(denseMatrix.GetExtent());
+            using var pMatrixBuffer =
+                Accelerator.Allocate2DDenseY<float>(sparseMaskMatrix.GetExtent());
+            using var outBuffer =
+                Accelerator.Allocate2DDenseY<float>(
+                    (denseMatrix.GetLength(0),
+                        sparseMatrix.GetLength(0)));
+            aMatrixBuffer.View.CopyFromCPU(stream, denseMatrix);
+            pMatrixBuffer.View.CopyFromCPU(stream, sparseMaskMatrix);
+
+            // Create a single-streamed sparse matrix buffer
+            var processor = Accelerator.CreateSparseTransposedMatrixMultiplierMasked<
+                float,
+                FloatEpsPredicate<Stride2D.General>,
+                Stride2D.General,
+                FloatMaskedSparseMatrixProcessor>();
+
+            // Multiply a single masked sparse matrix
+            processor(
+                stream,
+                new(pMatrixBuffer.View.AsGeneral(), 0.0f),
+                aMatrixBuffer.View.AsGeneral(),
+                sparseView,
+                outBuffer.View.AsGeneral());
+
+            // Compute reference values
+            var transposed = TransposeMatrix2D(sparseMatrix);
+            var expected = MultiplyMatrix2D(denseMatrix, transposed);
+
+            // Load our resulting matrix and test for equality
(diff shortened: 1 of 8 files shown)


Src/ILGPU.Algorithms/MatrixOperations/CondensedProductRows.cs
```diff
@@ -0,0 +1,235 @@
...
+                    {
+                        int sourceColumn = sparseMatrixShape.GetNeighbor(sparseRow, j);
+                        data[dataRow, j] = a[row, sourceColumn];
+                    }
+
+                    if (indexLength > 0)
+                    {
+                        rowIndices[dataRow] = row;
+                        columnIndices[dataRow] = sparseRow;
+                        ++dataRow;
```
corwinjoy, line 104: Under a parallel for loop the dataRow increment may have problems. I think that maybe we need to acquire a lock, capture the current value, then increment at line 93?

Src/ILGPU.Algorithms/MatrixOperations/MaskedMatrixProcessor.cs
```diff
@@ -0,0 +1,262 @@
...
+            int col = columnIndices[row];
+
+            float sum = 0.0f;
+            int i = 0;
+
+            // Do fast vectorized IO operations for multiples of 4
+            if (rowLength / 4 > 0)
+            {
+                var v4Rows = rows.Cast<float, Float4>();
+                var v4Cols = rows.Cast<float, Float4>();
```
corwinjoy, line 46: Is this really right?  I think you want cols.Cast here?

Src/ILGPU.Algorithms/MatrixOperations/MaskedMatrixProcessor.cs
```diff
@@ -0,0 +1,262 @@
...
+                        intermediate.W;
+                }
+                i *= 4;
+            }
+
+            // Do fast vectorized IO operations for multiples of 2
+            if (rowLength / 2 > 0)
+            {
+                var v2Rows = rows.Cast<float, Float2>();
+                var v2Cols = rows.Cast<float, Float2>();
```
corwinjoy, line 60: cols.Cast?

Src/ILGPU.Algorithms/MatrixOperations/SparseMatrixShape.cs
```diff
@@ -0,0 +1,332 @@
...
+            Parallel.For(0, numRows, i =>
+            {
+                int numRowEntries = 0;
+                for (int j = 0; j < numColumns; ++j)
+                {
+                    if (predicate.Apply(new Index2D(i, j)))
+                        ++numRowEntries;
+                }
+                outputNumNeighbors[i] = numRowEntries;
+                Atomic.Max(ref maxNumEntries, numRowEntries);
```
corwinjoy, line 179: I assume this gives a thread safe max?

Src/ILGPU.Algorithms/MatrixOperations/CondensedProductRows.cs
m4rs-mt: We should not use a parallel for in this case anyway...

Src/ILGPU.Algorithms/ConcurrentStreamProcessor.cs
```diff
@@ -0,0 +1,161 @@
...
+        /// </summary>
+        /// <param name="stream">The current accelerator stream.</param>
+        /// <param name="numActions">The number of actions to submit.</param>
+        /// <param name="action">
+        /// The action to invoke on each stream to submit work.
+        /// </param>
+        public void ProcessConcurrently(
+            AcceleratorStream stream,
+            int numActions,
+            Action<AcceleratorStream, int> action)
```
corwinjoy, line 98: Apologies, but I feel like this design is probably not what you want. It looks to me like you have a fair amount of locks here and in the case where you have a few long running actions this will bottleneck and stall while it waits for an action to finish.  Basically, what I think you want to do is:
1. Not tie this to a particular individual stream.
2. Dispatch this to a waiting stream, if available.
3. If no stream available wait until one signals ready.

Src/ILGPU.Algorithms/ConcurrentStreamProcessor.cs
```diff
@@ -0,0 +1,161 @@
...
+
+            if (stream != null && numActions == 1 && MaxNumConcurrentStreams == 1)
+            {
+                // Use the given stream to process the action request
+                action(stream, 0);
+            }
+            else
+            {
+                // Wait for the current stream to finish processing
+                stream?.Synchronize();
```
corwinjoy, line 115: This may take a long time waiting for the specified stream, but other streams may be ready and waiting. Basically, my understanding is that locking is bad and you should never do it unless you have to.

Src/ILGPU.Algorithms/ConcurrentStreamProcessor.cs
```diff
@@ -0,0 +1,161 @@
...
+                    for (int j = 0; j < numActionsPerStream; ++j)
+                    {
+                        int actionIndex = actionStride.ComputeElementIndex((i, j));
+                        if (actionIndex >= numActions)
+                            break;
+                        action(currentStream, actionIndex);
+                    }
+
+                    // Wait for the result
+                    currentStream.Synchronize();
```
corwinjoy, line 136: Again, is a synchronization necessary?  Can you just queue jobs and only synchronize later if you need to?

Src/ILGPU.Algorithms/ConcurrentStreamProcessor.cs
corwinjoy: I'm a .NET beginner, but it feels to me like you would want to make this similar to the existing [system.Threading.ThreadPool].
https://learn.microsoft.com/en-us/dotnet/api/system.threading.threadpool?view=net-7.0
https://referencesource.microsoft.com/#mscorlib/system/threading/threadpool.cs

Src/ILGPU.Algorithms/MatrixOperations/MaskedMatrixProcessor.cs
```diff
@@ -0,0 +1,112 @@
...
+        /// </summary>
+        /// <param name="stream">The current accelerator stream to use.</param>
+        /// <param name="aViews">The dense input matrices a of shape MxK.</param>
+        /// <param name="bViews">The sparse matrices b of shape NxK.</param>
+        /// <param name="outViews">Dense output matrices of shape of aViews.</param>
+        public void MultiplyBatched(
+            AcceleratorStream stream,
+            IReadOnlyList<ArrayView2D<T, TStride>> aViews,
+            IReadOnlyList<SparseMatrixView<T, TStride>> bViews,
+            IReadOnlyList<ArrayView2D<T, TStride>> outViews)
```
corwinjoy, line 99: Is this unused?  Also, it says it uses the predicate but I don't see where that comes in.

Src/ILGPU.Algorithms/MatrixOperations/MaskedSparseMatrixExtensions.cs
```diff
@@ -0,0 +1,390 @@
...
+            // Get all predicate results and perform thread compaction
+            int index = Grid.GlobalLinearIndex;
+            var maskIndex2D = aView.Stride.ReconstructFromElementIndex(index);
+            bool apply = maskIndex2D.X < aView.Extent.X & maskIndex2D.Y < aView.Extent.Y;
+
+            // Check the predicate (if in range) and compute thread-compaction offsets
+            if (apply)
+                apply = maskPredicate.Apply(maskIndex2D);
+            int groupOffset = GroupExtensions.ExclusiveScan<int, AddInt32>(
+                Utilities.Select(apply, 1, 0));
```
corwinjoy, line 209: So, if I understand this correctly you are making some assumptions about the sparsity structure defined by your predicate. Basically, you are assuming the mask will have a number of contiguous "chunks" so that you don't get too many groups here.  Otherwise, if you have a mask pattern like 0, 1, 0, 1, 0, 1, ...  I imagine this would quickly degrade.
I would suggest that we 
1. Get information from the quants about what their mask pattern looks like.
2. Document the kind of mask structure / assumption that this code is optimized for.
3. When we add unit tests show both "good" and "adversarial" cases where the performance degrades.

Src/ILGPU.Algorithms/MatrixOperations/MaskedSparseMatrixExtensions.cs
```diff
@@ -0,0 +1,390 @@
...
+            // and loaded upon request (based on the structure of the algorithm) into the
+            // L1 parts of the processing unit
+
+            // Implicit transpose of B
+            for (var z = 0; z < numNeighbors; z++)
+            {
+                int columnIndex = bView.Neighbors[index2D.Y, z];
+                dotProduct = processor.MultiplyAdd(
+                    dotProduct,
+                    aView[index2D.X, columnIndex],
```
corwinjoy, line 238: I'm a bit confused by how the masking actually works here. I guess index2D.X will limit the X range we consider. But what about the Y values? That is, if columnIndex takes on a Y value that should be excluded because of the mask, I don't see how this gets excluded from the dotProduct.

Src/ILGPU.Algorithms/ConcurrentStreamProcessor.cs
m4rs-mt: Thanks for your comments. It will indeed distribute the workload by attaching them to the the different accelerator streams available. This way, we can launch multiple kernels in parallel on the device. One design decision could be to also use the incoming stream for further processing. However, this could lead to invalid results as all previous operations on the device need to be finished before using multiple async streams. The downside of this approach is clearly the fact that we have to wait for the input accelerator stream to finish. There is no other way to launch multiple kernels concurrently on the device.

Src/ILGPU.Algorithms/ConcurrentStreamProcessor.cs
m4rs-mt: This is one of the situations in which you have to wait as the async processing streams that will be used afterwards could potentially rely on data written by kernels associated with the input stream.

Src/ILGPU.Algorithms/ConcurrentStreamProcessor.cs
m4rs-mt: Unfortunately, not at the moment as ILGPU does not feature a way to attach streams to each other.

Src/ILGPU.Algorithms/MatrixOperations/MaskedMatrixProcessor.cs
m4rs-mt: Not quite sure what you mean... the `Predicate` is a property defined in line 66.

Src/ILGPU.Algorithms/MatrixOperations/MaskedSparseMatrixExtensions.cs
m4rs-mt: Currently, unstructured sparsity is expected to affect the entire matrix and thus should also affect the Y coordinates based on thread compaction.

corwinjoy: I guess I would need to trace it. Basically, I would want to investigate the case where columnIndex takes on a y value that is masked out in A via the predicate. So, I'm not clear what happens when aView is asked to index a value that should be masked out.

Src/ILGPU.Algorithms/MatrixOperations/MaskedMatrixProcessor.cs
corwinjoy: OK. I see now. Thanks.

Src/ILGPU.Algorithms/MatrixOperations/MaskedSparseMatrixExtensions.cs
```diff
@@ -0,0 +1,390 @@
...
+        {
+            var maskView = SharedMemory.GetDynamic<int>();
+            Trace.Assert(
+                Group.DimX <= maskView.IntLength,
+                "Invalid shared memory config");
+
+            // Get all predicate results and perform thread compaction
+            int index = Grid.GlobalLinearIndex;
+            var maskIndex2D = aView.Stride.ReconstructFromElementIndex(index);
+            bool apply = maskIndex2D.X < aView.Extent.X & maskIndex2D.Y < aView.Extent.Y;
```
corwinjoy, line 203: What happens when apply is false here?  That is, the predicate is out of range?  Shouldn't we throw an error?

m4rs-mt: It can happen that the index will be out of range based on the current grid/group configuration that we dispatch. Threads processing oor elements will just be skipped and terminate peacefully.

Src/ILGPU.Algorithms.Tests/MatrixTests.cs
```diff
@@ -0,0 +1,258 @@
...
+            var sparseMatrix = new float[length, length];
+            var sparseMaskMatrix = new float[length, length];
+
+
+            var random = new System.Random(seed);
+            for (int i = 0; i < denseMatrix.GetLength(0); ++i)
+            {
+                for (int j = 0; j < denseMatrix.GetLength(1); ++j)
+                {
+                    denseMatrix[i, j] = random.NextSingle();
```
corwinjoy, line 167: I don't suppose there is a way to suppress this error about NextSingle? The page of warnings from Github makes this hard to read.

Src/ILGPU.Algorithms/MatrixOperations/MaskedMatrixProcessor.cs
```diff
@@ -0,0 +1,114 @@
...
+
+        #region Methods
+
+        /// <summary>
+        /// Multiplies the given matrices using the currently assigned predicate.
+        /// </summary>
+        /// <param name="stream">The current accelerator stream to use.</param>
+        /// <param name="aView">The dense input matrix a of shape MxK.</param>
+        /// <param name="bView">The sparse matrix b of shape NxK (will transpose).</param>
+        /// <param name="outView">A dense output matrix of shape of aView.</param>
```
corwinjoy, line 76: I don't think the description of outView is correct. Doesn't outView contain aView * transpose(bView) which would have a shape of MxN.
I think it would be good to give a description of what this function is doing.
e.g. This function computes aView * bView_T  using the given accelerator stream.

Src/ILGPU.Algorithms/MatrixOperations/MaskedMatrixProcessor.cs
```diff
@@ -0,0 +1,114 @@
...
+
+        /// <summary>
+        /// Multiplies the given matrices using the currently assigned predicate.
+        /// </summary>
+        /// <param name="stream">The current accelerator stream to use.</param>
+        /// <param name="aViews">The dense input matrices a of shape MxK.</param>
+        /// <param name="bViews">
+        /// The sparse matrices b of shape NxK (will transpose).
+        /// </param>
+        /// <param name="outViews">Dense output matrices of shape of aViews.</param>
```
corwinjoy, line 96: Comment on outViews as above.

Src/ILGPU.Algorithms/MatrixOperations/MaskedSparseMatrixExtensions.cs
```diff
@@ -0,0 +1,448 @@
...
+using System;
+using System.Diagnostics;
+using System.Runtime.CompilerServices;
+
+namespace ILGPU.Algorithms.MatrixOperations
+{
+    #region Mask Predicates
+
+    /// <summary>
+    /// An internal predicate based on mask arrays.
```
corwinjoy, line 24: An internal predicate used to represent a masked array.

Src/ILGPU.Algorithms/MatrixOperations/MaskedSparseMatrixExtensions.cs
```diff
@@ -0,0 +1,448 @@
...
+
+            // Note that profiling yielded that explicit L1 shared-memory caching did not
+            // help to improve performance in this case. This is also due to the fact
+            // that most reused results automatically end up being fetching into L2 cache
+            // and loaded upon request (based on the structure of the algorithm) into the
+            // L1 parts of the processing unit
+
+            for (var neighborIndex = 0; neighborIndex < numNeighbors; ++neighborIndex)
+            {
+                // Load index and sparse edge weight
```
corwinjoy, line 272: // Note that the format for bView means we will read the transposed entry from bView

Src/ILGPU.Algorithms/MatrixOperations/MaskedSparseMatrixExtensions.cs
```diff
@@ -0,0 +1,448 @@
...
+                int columnIndex = bView.Neighbors[index2D.Y, neighborIndex];
+                T bValue = bView.EdgeWeights[index2D.Y, neighborIndex];
+
+                // Load transposed a value
```
corwinjoy, line 276: I don't think you mean this. bView is transposed but aView is not.

Src/ILGPU.Algorithms.Tests/MatrixTests.cs
```diff
@@ -0,0 +1,258 @@
...
+            Skip.If(length > 512);
+            
+            using var stream = Accelerator.CreateStream();
+            
+            // Our starting dense matrix
+            var denseMatrix = new float[baseLength, length];
+
+            // Setup sparse 2D matrix
+            var sparseMatrix = new float[length, length];
+            var sparseMaskMatrix = new float[length, length];
```
corwinjoy, line 159: I would recommend making the shape of sparseMatrix be [length2, length] to more explicitly test the transpose and dimension matching.

Src/ILGPU.Algorithms.Tests/MatrixTests.cs
```diff
@@ -0,0 +1,258 @@
...
+                { 39, 17, 2918291, 0.5f },
+                { 39, 17, 2918291, 0.05f },
+                
+                { 376, 288, 4132107, 0.8f },
+                { 376, 288, 4132107, 0.5f },
+                { 376, 288, 4132107, 0.05f },
+                
+                { 1829, 928, 31821912, 0.8f },
+                { 1829, 928, 31821912, 0.5f },
+                { 1829, 928, 31821912, 0.05f },
```
corwinjoy, line 45: Maybe note that this last set of tests is skipped for now?

corwinjoy (changes requested):
Overall looks pretty good. I suggested a few minor changes.
I think the big piece that is missing here is documentation as to the motivation of this class and what problem it solves. We know, but I think users will be mystified.

Src/ILGPU.Algorithms.Tests/MatrixTests.cs
m4rs-mt: These tests will be skipped on the CPU but will run in the cloud on our GPUs.

corwinjoy (approved):
Looks good. As mentioned, I think the only thing that is missing is a piece of documentation explaining why end-users would want to use this / what problem this solves.

m4rs-mt:
> Looks good. As mentioned, I think the only thing that is missing is a piece of documentation explaining why end-users would want to use this / what problem this solves.

Absolutely! I think we should create a sample in a separate PR and add further documentation to the wiki before the final release.

corwinjoy (approved):
OK. Updated tests look good.

MoFtZ (approved):

PR #926, opened 2023-02-09
Implemented tool to fetch Cuda release history.

MoFtZ:
The tool will download and parse the Cuda [release history](https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#release-notes) and generate an XML file with the extract details.

During compilation, ILGPU will use T4 templates to parse the XML file and generate appropriate constants.

Finally, a weekly CI job is scheduled to automatically check for new Cuda instruction sets and architectures, and the corresponding driver versions.


Tools/CudaVersionUpdateTool/CudaVersionUpdateTool/Abstractions/CudaVersionSet.cs
```diff
@@ -0,0 +1,33 @@
...
+//
+// This file is part of ILGPU and is distributed under the University of Illinois Open
+// Source License. See LICENSE.txt for details.
+// ---------------------------------------------------------------------------------------
+
+using System.Diagnostics;
+
+namespace CudaVersionUpdateTool.Abstractions
+{
+    [DebuggerDisplay("{InstructionSet}, {DriverVersion}, {Architecture}")]
```
m4rs-mt, line 16: Nice feature 👍

Tools/CudaVersionUpdateTool/CudaVersionUpdateTool.sln
```diff
@@ -0,0 +1,25 @@
+﻿
```
m4rs-mt, line 1: Should we create a parent solution for "all" tools (including future tools)?

m4rs-mt (dismissed):

Tools/CudaVersionUpdateTool/CudaVersionUpdateTool.sln
MoFtZ: Makes sense. I have updated the PR to create a consolidated Tools solution.

m4rs-mt (approved):

PR #933, opened 2023-02-15
Added generic constraints to fix failing unit tests on .NET 7.0.3 SDK.

### LINKED ISSUE #932: Unit tests fail when using .NET 7.0.3
On Feb 14th 2023, Microsoft released the .NET 7.0.3 ([SDK 7.0.200](https://github.com/dotnet/core/issues/8191)). This is causing unit tests in our CI pipeline to fail.

After installing SDK 7.0.200, I can reproduce the failures on Windows 10 using the command line `dotnet test`. However, I am unable to reproduce the test failures in VS2022 (v17.4.5).

MoFtZ:
Fixes #932.

Looks like the .NET 7.0.3 SDK identified that the C# extension method was potentially calling a class instance, and added IL instructions for boxing.

I have applied the `struct` generic type constraint. This removes the boxing in IL. ILGPU probably had issues converting the boxing into internal IR nodes.

I have also changed the call from `stride.ComputeElementIndex(...)` to `ComputeElementIndex(stride, ...)`. Without this change, the IL would use `callvirt`. Since we are using a `struct` constraint, it should not be possible for the call to `ComputeElementIndex` to be overridden.

### DIFF
--- a/Src/ILGPU.Algorithms/HistogramExtensions.cs
+++ b/Src/ILGPU.Algorithms/HistogramExtensions.cs
@@ -1,6 +1,6 @@
 ﻿// ---------------------------------------------------------------------------------------
 //                                   ILGPU Algorithms
-//                        Copyright (c) 2020-2022 ILGPU Project
+//                        Copyright (c) 2020-2023 ILGPU Project
 //                                    www.ilgpu.net
 //
 // File: HistogramExtensions.cs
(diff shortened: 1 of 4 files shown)


MoFtZ:
Fixed issues in ILGPU.Algorithms. Looks like all the unit tests are passing now.

m4rs-mt (approved):

PR #961, opened 2023-03-06
Allow automated PRs to trigger other workflows.

MoFtZ:


MoFtZ:
Waiting to determine which access token to use.

jgiannuzzi:
you may want to take a look at https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions

MoFtZ:
The issue is that Github [prevents a workflow from triggering another workflow](https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#triggering-a-workflow-from-a-workflow) when using the default token. The recommendation is to use a PAT.

The Github Action we are using also [specifies the use of a PAT](https://github.com/technote-space/create-pr-action#github_token).

I [found some other suggestions](https://github.com/peter-evans/create-pull-request/blob/main/docs/concepts-guidelines.md#workarounds-to-trigger-further-workflow-runs), such as using a fork to raise the automated PR. This would reduce the security concerns

MoFtZ:
Abandoning. Will revisit another time.

Issue #971, opened 2023-03-16
'context is destroyed' exception after using CuFFT.


ebprj21:
I am using your masterpiece very well.
If I had the ability,  I would be able to help with anything, but in reality, I only write issue like this.

ILGPU's CuFFT example code is as follows:
(\Samples\AlgorithmsCuFFT\Program.cs)

```C#
static void DoForwardPlan(
    CudaAccelerator accelerator,
    CuFFT cufft,
    Complex[] input,
    out Complex[] output)
{
    using (var stream = accelerator.CreateStream() as CudaStream)
    {
        using (var inputBuffer = accelerator.Allocate1D(input))
        {
            using (var outputBuffer = accelerator.Allocate1D<Complex>(input.Length))
            {
                CuFFTException.ThrowIfFailed(
                    cufft.Plan1D(
                        out var plan,
                        input.Length,
                        CuFFTType.CUFFT_Z2Z,
                        batch: 1));
                using (plan)
                {
                    plan.SetStream(stream);
                    CuFFTException.ThrowIfFailed(
                        plan.ExecZ2Z(
                            inputBuffer.View.BaseView,
                            outputBuffer.View.BaseView,
                            CuFFTDirection.FORWARD));

                    output = outputBuffer.GetAsArray1D(stream);
                }
                WorkaroundKnownIssue(accelerator, cufft.API);

                Console.WriteLine("Output Values:");
                for (var i = 0; i < output.Length; i++)
                    Console.WriteLine($"  [{i}] = {output[i]}");
            }
        }
    }
}
```

In the above code, when 'plan' is Disposed, Context is also destroyed.
Therefore, when outputBuffer is Disposing, 'context is destroyed' exception is thrown.

The issue essentially looks like this.
[https://github.com/inducer/pycuda/discussions/356](https://github.com/inducer/pycuda/discussions/356)

I hope you will consider this issue seriously. please.

ps) BTW, the new features in CUDA v12 are interesting.
[CUDA Context-Independent Module Loading](https://developer.nvidia.com/blog/cuda-context-independent-module-loading/)

MoFtZ:
hi @ebprj21. Thanks for reporting this issue.

This looks like an issue with the CUDA SDK. Notice the call to `WorkaroundKnownIssue`. The sample code limits the workaround to specific versions of the CUDA SDK. The workaround will restore the original CUDA context. If you force the workaround to always run, it should fix your issue.

[...]

PR #975, opened 2023-03-16
Added support for LibDevice using Cuda Release v12.

MoFtZ:
#971 mentioned that LibDevice was not working with Cuda Release v12. Upon further investigation, the [v12 release notes](https://docs.nvidia.com/cuda/archive/12.0.0/cuda-toolkit-release-notes/index.html) state:

> NVVM IR Update: with CUDA 12.0 we are releasing [NVVM IR 2.0](https://docs.nvidia.com/cuda/nvvm-ir-spec/index.html) which is incompatible with NVVM IR 1.x accepted by the libNVVM compiler in prior CUDA toolkit releases. Users of the libNVVM compiler in CUDA 12.0 toolkit must generate NVVM IR 2.0.

Added `!nvvmir.version` to the NVVM IR that is generated. When previously omitted, this defaulted to v1.0.

### DIFF
--- a/Src/ILGPU/Backends/PTX/PTXBackend.cs
+++ b/Src/ILGPU/Backends/PTX/PTXBackend.cs
@@ -1,6 +1,6 @@
 ﻿// ---------------------------------------------------------------------------------------
 //                                        ILGPU
-//                        Copyright (c) 2018-2022 ILGPU Project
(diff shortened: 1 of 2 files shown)


m4rs-mt (approved):

ebprj21:
Now, LibDevice works very well on CUDA v12.

Thanks to your masterpiece!

PR #1023, opened 2023-06-29
Added new Vector data types for .NET 7.

m4rs-mt:
This PR introduces new vectorized data types that are compatible with the `INumberBase` interfaces from `.NET 7`. These data types simplify the development of GPU-based algorithms in ILGPU by making it easier to generate vectorized code.

Note: This PR now also contains an additional `IAccumulationVectorType` interface to support more precise accumulation vectors used to aggregate information.

~~This PR depends on #1021 and #1022~~.

PR #1022, opened 2023-06-29
Extended RNG functionality.

m4rs-mt:
This PR enhances the current RNG functionality by introducing new creation methods and extending buffer and view features.

PR #1028, opened 2023-07-07
Added new generic optimization API

m4rs-mt:
This PR adds a newly created optimization API to build massively-parallel (particle based) optimizers. The current PR includes a set of classes to build custom optimizers while sharing a unified problem-focused API surface.

The optimization classes support vectorized IO and computation operations using the type parameters `TNumericType` and `TElementType`. `TElementType` refers to the underlying scalar type, like `float`. `TNumericType` has to implement the `IVectorizedType` interface to support combined vectorized types like `Float32x4`.

Users can implement their own objective function by implementing to `IOptimizationFunction` interface as shown below:
```c#
  public readonly struct Objective : IOptimizationFunction<Float32x2, float, float>
  {
      public float Evaluate(LongIndex1D index, Index1D dimension, SingleVectorView<Float32x2> positionView)
      {
          float result = 0;
          for (Index1D i = 0; i < dimension; ++i)
          {
              var vec = positionView[i];
              var dist = vec - new Float32x2(5.0f, 5.0f); // The closer we are to (5.0f, 5.0f, ...), the better!
              result += dist.X * dist.X + dist.Y * dist.Y;
          }
          return result / dimension;
      }
 
      public bool CurrentIsBetter(float current, float proposed) => current <= proposed;
 }
```
Note that this code uses a specialized implementation operating on `Float32x2` vectors.

The high level idea is to derive your custom optimizer from `OptimizationEngine<...>`. Such an engine is meant to manage all buffers and most of the reusable kernels. An engine is then used to create a specialized `Optimizer` instance to take a given objective function into account. Using this design allows to reuse the same memory allocations with different objective functions.

A sample use case of an imaginary optimization engine is shown here:
```c#
// Create the current optimization engine and a specialized optimizer using our previously shown objective function
using var myEngine = new CustomOptimizationEngine<Float32x2, float, float, ...>(...);
using var optimizer = myEngine.CreateOptimizer(stream, random, new Objective());

// Load all variable bounds and custom use-case and domain-specific parameters
optimizer.LoadBounds(stream, lowerBounds, upperBounds);
optimizer.LoadParameters(stream, /* Custom optimizer parameters */);

// Initialize best-known positions and our best-known distance for a 4-dimensional problem
var bestPosition = new float[] { 0.0f, 0.0f, 0.0f, 0.0f };
var run = optimizer.BeginOptimization(stream, bestPosition, float.MaxValue);

// Realize a user-defined optimization loop
for (int i = 0; i < 100; ++i)
    run.Step();

// a) Finish the optimization run by accumulating all results into GPU accessible buffers
var gpuResult = run.Finish();

// b) Finish the optimization run by accumulating all results and fetching them into CPU memory
var result = run.FinishToCPUAsync();
```

~~Note that this PR requires PR #1022, PR #1023, and PR #1027.~~

### DIFF
--- a/Src/ILGPU.Algorithms/Optimization/OptimizationResult.cs
+++ b/Src/ILGPU.Algorithms/Optimization/OptimizationResult.cs
@@ -1,0 +1,95 @@
+// ---------------------------------------------------------------------------------------
+//                                   ILGPU Algorithms
+//                           Copyright (c) 2023 ILGPU Project
+//                                    www.ilgpu.net
+//
+// File: OptimizationResult.cs
+//
+// This file is part of ILGPU and is distributed under the University of Illinois Open
+// Source License. See LICENSE.txt for details.
+// ---------------------------------------------------------------------------------------
+
+using System;
+
+namespace ILGPU.Algorithms.Optimization
+{
+    /// <summary>
+    /// An optimization result view pointing to regions in accelerator memory.
+    /// </summary>
+    /// <typeparam name="TElementType">
+    /// The numeric element type used for optimization.
+    /// </typeparam>
+    /// <typeparam name="TEvalType">
+    /// The evaluation type used for evaluation.
+    /// </typeparam>
+    public readonly struct OptimizationResultView<TElementType, TEvalType>
+        where TElementType : unmanaged
+        where TEvalType : unmanaged
+    {
+        internal OptimizationResultView(
+            VariableView<TEvalType> resultView,
+            ArrayView<TElementType> positionView,
+            double elapsedTime)
+        {
+            ResultView = resultView;
+            PositionView = positionView;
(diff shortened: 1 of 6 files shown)


Src/ILGPU.Algorithms/Optimization/OptimizationResult.cs
MoFtZ, line 1: QUESTION: Does this file need to use `#if NET7_0_OR_GREATER`?

MoFtZ (approved):

Src/ILGPU.Algorithms/Optimization/OptimizationResult.cs
m4rs-mt: Nice catch. I removed the `#if` checks from this file.

PR #1033, opened 2023-07-16
Removed .NetFramework471, .NetStandard2.1, .NetCore3.1, and .Net5.0 builds

m4rs-mt:
This PR removes Removed .NetFramework471, .NetStandard2.1, .NetCore3.1, and .Net5.0 builds from the project, the CI pipelines, and all CodeQL analyses. 

~~@MoFtZ @kilngod @NullandKale question: What about the .Net Standard 2.1 build?~~


MoFtZ:
> question: What about the .Net Standard 2.1 build?

Do we need it for Mono or Unity? Not sure if anyone actually got it to work on those targets.

MoFtZ:
Also, can we move from `net47` to `net472` or `net480`?

NullandKale:
> > question: What about the .Net Standard 2.1 build?
> 
> Do we need it for Mono or Unity? Not sure if anyone actually got it to work on those targets.

I have tested with unity and some of the system.reflection libraries that we use do not work. I have been completely unable to get ILGPU to work with unity.

m4rs-mt:
> @m4rs-mt Also, can we move from `net47` to `net472` or `net480`?

I think we can. We should remove this dependency for the 2.0 release. However, I think we should still keep support for at least one .Net Framework until we reach 2.0.

m4rs-mt:
@MoFtZ @jgiannuzzi I had to disable package validation as validation against .net4.7.1, net5.0 and netstandard2.1 is no longer supported. Given the fact the we updated the version number to v2.0, and will break parts of the API anyway, this will not be a show stopper until we release v2.0 stable.

.github/workflows/codeql-analysis.yml
```diff
@@ -64,8 +64,6 @@ jobs:
         $LibraryTargetFrameworks=$LibraryUnitTestTargetFrameworks=$LibrarySamplesTargetFrameworks=$LibrarySamplesTargetFrameworksWindows="${{ matrix.framework }}"
         Switch($LibrarySamplesTargetFrameworks)
         {
```
MoFtZ, line 66: MINOR: This switch statement can be simplified, since we just need to add `-windows`.

Samples/Directory.Build.props
```diff
@@ -7,21 +7,9 @@
     <LibrarySamplesTargetFrameworks>net6.0</LibrarySamplesTargetFrameworks>
```
MoFtZ, line 7: QUESTION: Should we relax the condition, so that `net6.0` becomes our fallback? Looks like the fallback is to not define a TFM.

MoFtZ (approved):

.github/workflows/codeql-analysis.yml
m4rs-mt: I was thinking the same. However, as we want to test Velocity more excessively on other platforms as well, I was thinking that this could come in handy while still making it easy to understand.

PR #1038, opened 2023-08-02
Random ranges

m4rs-mt:
This PR adds new `IRandomRange` and `IRandomRangeProvider` interfaces to define common methods to combine random number generators with `[min,.. max)` value ranges. It adds a text template to generate `RandomRangeXYZ` types representing tuples of `min` and `max` values with utility RNG functions. The template also generates provider implementations combining the tuples with actual RNG types like `RandomRangeXYZProvider`.

This PR depends on PRs ~~#1033, #1037, and #1057.~~

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -291,6 +291,7 @@ Src/ILGPU.Algorithms/IL/ILContext.Generated.cs
 Src/ILGPU.Algorithms/PTX/PTXContext.Generated.cs
 Src/ILGPU.Algorithms/RadixSortOperations.cs
 Src/ILGPU.Algorithms/Vectors/VectorTypes.cs
+Src/ILGPU.Algorithms/Random/RandomRanges.cs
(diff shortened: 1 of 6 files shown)


MoFtZ:
Unit tests are failing in Release mode only. Need to switch to Release configuration in Visual Studio when running the unit tests, in order to reproduce the issue.

Looks like it is failing become ILGPU does not support `BinaryArithmeticKind.Add` for `ArithmeticBasicValueType.Int1`.

m4rs-mt:
Given the fact that this PR should not change the behavior of the ILGPU tests affected at the moment, this is particularly interesting.

MoFtZ:
I think this is a change in the latest .NET runtimes - our nightly builds are failing too.

m4rs-mt:
This would actually make sense.

MoFtZ (approved):

PR #1040, opened 2023-08-07
New CPU-based Optimizer

m4rs-mt:
This PR adds a new meta optimizer to the Algorithms library which is designed for CPUs. It implements an optimization-performance and runtime-performance optimized version of the SGO algorithm:
_Squid Game Optimizer (SGO): a novel metaheuristic algorithm [doi: 10.1038/s41598-023-32465-z]_.

The primary use case of this optimizer is to solve sophisticated optimization problems without the need to transform the input problem description into any intermediate state. Due to its efficiency it can be directly used in any .Net program to get excellent approximations of optimization results in near real time.

The implemented version **does not implement the vanilla SGO algorithm from the paper**. Instead, it uses modified update functions and specially tweaked position update logic using multiple buffers and tuned SGO-winner lists. These modifications of the original algorithm make this implementation significantly better in terms of optimization quality and runtime performance. Moreover, this version is fully parallelized and has the ability to use SIMD vector instructions to improve runtime performance.

This PR also adds a set of new optimizer test cases to ensure the convergence of the implemented approach.

Sample use case creating a scalar version of the optimizer while operating on floats to represent position vectors and evaluation results.
```c#
var random = new System.Random();
using var optimizer = MetaOptimizer.CreateScalar<
    float,
    float,
    RandomRanges.RandomRangeFloatProvider<XorShift64Star>>(
    random,
    1024 // Number of particles,
    2 // Number of dimensions);

optimizer.LowerBounds = new float[] { -1.0f, -1.0f };
optimizer.UpperBounds = new float[] { 1.0f, 1.0f };

var result = optimizer.Optimize(
    // Custom objective function here,
    // Custom break logic here,
    float.MaxValue // Best known solution);
```

### DIFF
--- a/Src/ILGPU.Algorithms.Tests.CPU/CPUMetaOptimizerTests.cs
+++ b/Src/ILGPU.Algorithms.Tests.CPU/CPUMetaOptimizerTests.cs
@@ -1,0 +1,336 @@
+// ---------------------------------------------------------------------------------------
+//                                   ILGPU Algorithms
+//                           Copyright (c) 2023 ILGPU Project
+//                                    www.ilgpu.net
+//
+// File: CPUMetaOptimizerTests.cs
+//
+// This file is part of ILGPU and is distributed under the University of Illinois Open
+// Source License. See LICENSE.txt for details.
+// ---------------------------------------------------------------------------------------
+
+using ILGPU.Algorithms.Optimization.CPU;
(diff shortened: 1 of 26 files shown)


m4rs-mt:
This feature should actually live in a different repository as it is not directly ILGPU related in any way.

MoFtZ (approved):

PR #1042, opened 2023-08-09
Collapsable Sidebar for Website

Yey007:
I noticed the sidebar for documentation wasn't collapsable, making documentation difficult to read on mobile. This PR makes the sidebar collapsible. Though more work can be done for a better mobile experience, I believe this is at least a start. Let me know what you think!


Site/assets/js/sidebar-collapse.js
naskio, line 1: Currently, the collapse status is not being persisted. This means that when navigating to a new page, the collapse status will be open even if it was closed on the previous page. To address this issue, it may be a good idea to use LocalStorage to persist and retrieve the collapse status.

Site/_sass/extras-styles.scss
```diff
@@ -227,22 +227,39 @@
...
+
 .light header {
   box-shadow: 0 0 25px 0 rgba(0, 0, 0, 0.1);
 }
 
+#sidebar-collapse {
+  cursor: pointer;
+  @include sidebar-collapse-transition(transform);
+}
+
```
naskio, line 240: Currently, the icon is reversed, reflecting the current state instead of the intended action. I suggest adding a transformation here or changing the SVG being used to address this issue.

naskio (changes requested):
Hi @Yey007,
Thank you for your contribution. After testing it locally, I have a few comments that I would like to share with you.

Site/_sass/extras-styles.scss
Yey007: Done!

Site/assets/js/sidebar-collapse.js
Yey007: Done! Not the prettiest code in the world, but I think it does the job. We can discuss improvements if you have suggestions--I wasn't quite sure what to do myself.

naskio (changes requested):
Hi @Yey007 
Everything looks good to me except these 2 files.

Src/ILGPU.Algorithms/Random/RandomRanges.cs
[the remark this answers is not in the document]
Yey007: Huh, that's interesting. I wonder what happened here.

naskio (approved):
thanks @Yey007,
LGTM

Site/assets/js/sidebar-collapse.js
```diff
@@ -0,0 +1,40 @@
+// ---------------------------------------------------------------------------------------
+//                                        ILGPU
+//                           Copyright (c) 2022 ILGPU Project
```
m4rs-mt, line 3: Shouldn't this be 2023?

Yey007: Yes, I will fix this right away.

naskio (approved):

m4rs-mt (approved):

PR #1068, opened 2023-08-30
Type node extensions

m4rs-mt:
This PR refines the current `TypeNode` implementation by modularizing the managed type generation from ILGPU types. This allows to use custom type providers to rewrite internal type representations resulting in specialized .Net types being emitted by the ILGPU type system.

### DIFF
--- a/Src/ILGPU/Util/TypeExtensions.cs
+++ b/Src/ILGPU/Util/TypeExtensions.cs
@@ -255,6 +255,29 @@ namespace ILGPU.Util
             };
 
         /// <summary>
+        /// Resolves the managed type for the given basic-value type.
+        /// </summary>
+        /// <param name="type">The source type.</param>
+        /// <returns>The resolved managed type.</returns>
+        public static Type GetManagedType(this ArithmeticBasicValueType type) =>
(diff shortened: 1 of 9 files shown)


Src/ILGPU/Util/TypeExtensions.cs
```diff
@@ -380,6 +403,80 @@ type switch
...
+                    return type;
+            }
+        }
+
+        /// <summary>
+        /// Limits the given basic type to 64 bits.
+        /// </summary>
+        /// <param name="type">The source type.</param>
+        /// <returns>The limited arithmetic basic value.</returns>
+        public static ArithmeticBasicValueType LimitTo64Bit(
```
MoFtZ, line 458: QUESTION: Should this be called `ForceTo64Bit`?

m4rs-mt: This makes sense! Let me change that.

MoFtZ (changes requested):
Waiting on rename of `ForceTo64Bit`.

MoFtZ (approved):

PR #1088, opened 2023-09-08
Modified OpenCL generator to forward declare all variables.

MoFtZ:
Backported #1084 to ILGPU v1.5.1 branch.

Based on #1094


m4rs-mt (approved):

m4rs-mt:
@MoFtZ I think we also need to back port #1078 in order to run the latest CI.

PR #1104, opened 2023-10-06
removing Site/Gemfile.lock

naskio:
The `Gemfile.lock` is not necessary in our Jekyll setup because we are using GitHub Pages.

> During the build phase, GitHub Pages always utilizes its own dependencies and versions, which are listed [here](https://pages.github.com/versions/).

- This pull request should address all Dependabot alerts related to `Site/`.
- We retain the `Gemfile` file for setting up the local development environment.


m4rs-mt:
Nice catch 👍

m4rs-mt (approved):

PR #1105, opened 2023-10-07
Fix F# sample

kant2002:


kant2002:
Since I did not touch license headers, I would like to understand what's my future steps here.

MoFtZ:
hi @kant2002. Thank you for taking an interested in ILGPU.

Looking at your PR, what problem is it trying to fix in the F# sample code?

kant2002:
When I run sample code the GPU part return 0 instead of results

MoFtZ:
@kant2002 That actually sounds like a bug with ILGPU, not with the sample project.

The call to `GetAsArray1D` internally calls `Synchronize`. It should not be necessary to call `Synchronize` twice (once explicitly, and a second time within GetAsArray1D).

MoFtZ:
@kant2002 OK, I see the problem now.

The F# sample project is launching the kernel on the default stream, rather than the newly created stream. I have raised #1106 to fix this issue.

m4rs-mt:
@kant2002 welcome to the ILGPU community and thank you for raising this PR. Nice catch 👍!

PR #982, opened 2023-04-08
Added StructureOfArrays source generator.

MoFtZ:
This PR adds support for generating a [structure of arrays](https://en.wikipedia.org/wiki/AoS_and_SoA) from a `struct`.

When the `[GeneratedStructutreOfArrays]` attribute is added to a `partial struct`, the source generator will populate the field members.

```CSharp
// Example:
public struct MyPoint
{
    public int X;
    public int Y;
}

[GeneratedStructureOfArrays(typeof(MyPoint), 4)]
partial struct MyPoint4
{ }

// Will generate:
unsafe partial struct MyPoint4
{
    public fixed int X[4];
    public fixed int Y[4];
}
```

The source generator is also able to expand inner structures as necessary.
```CSharp
// Example:
public struct MyPoint
{
    public int X;
    public int Y;
    public InnerStruct Z;
}

public unsafe struct InnerStruct
{
    public fixed int A[4];
    public fixed int B[2];
}

[GeneratedStructureOfArrays(typeof(MyPoint), 4)]
partial struct MyPoint4
{ }

// Will generate:
unsafe partial struct MyPoint4
{
    public fixed int X[4];
    public fixed int Y[4];
    public fixed int Z_A[16];
    public fixed int Z_B[8];
}

PR #1120, opened 2023-11-11
(WIP) Add basic support for Roslyn Analyzers

Yey007:
Hello! This is a proof-of-concept for shipping Roslyn Analyzers with ILGPU. The intent is to build a framework to easily add analyzers in the future, and provide a basic one that covers one of the common beginner mistakes with ILGPU: using reference types in the kernel.

#### TODO:
- [x] A number of things to improve with the analysis itself, outlined in the code
- [ ] Add tests
- [x] Add more samples
- [ ] ~~Remove remaining extraneous files, like generated .Designer.cs files~~
- [ ] Add analyzers to CI/CD pipeline
- [x] Add analyses to AnalyzerReleases.Shipped.md
- [x] Improve README.md


Yey007:
A couple of questions for discussion that came up while implementing the reference type analyzer specifically:

1. Should errors trace back to the point reference types are instantiated? Or should they pop up anywhere they're used?
2. Similarly, should invoked methods that _use_ reference types be highlighted at the invocation as well as at the usage of the reference type? Currently this is not the behavior.
3. How much effort do we want to put into discovering methods that are kernels? Dataflow analysis can probably be used to achieve somewhat good coverage, depending on the current implementation of the API in Roslyn. I haven't looked much into how it works.
4. Does having `KernelAnalyzer` as a base class seem useful? I don't know how many of our analyzers will really require the behavior of going through every line of code potentially accessible from a kernel.

Yey007:
Closing in favor of rewriting changes on top of #982.

PR #1133, opened 2023-12-01
Added specialized implementations of permutation methods

dfki-jugr:
In this PR, we provide different implementations of a permutation algorithm that permutes a given dataset on GPUs.
There are specialized implementations for a warp-wide, group-wide and device-wide permutation.


m4rs-mt (approved):
Nice extension!

PR #1151, opened 2024-01-17
Optimized PTX IntrinsicMath implementation to use LibDevice.

MoFtZ:
Depends on #1148.

Restructured ILGPU and ILGPU.Algorithms, moving the `IntrinsicMath` implementations into ILGPU itself. `XMath` has other functions that are not part of `IntrisicMath`, so it will stay as-is for now.

`CLMath` in ILGPU.Algorithms only needed to support `Rcp` and `Log(x,y)`. These have been moved into ILGPU, and `CLMath` has been removed.

`PTXMath` in ILGPU.Algorithms provided  a number of math functions using Cordic implemenentations. Now that pre-generated LibDevice is available in ILGPU, switched all the IntrinsicMath functions to call LibDevice for Cuda GPUs. The pre-generated LibDevice PTX code only works on `>= SM_60`, so the Cordic functions in ILGPU.Algorithms have been modified to only register for `< SM_60`. Otherwise, they are no longer used.

Unit Tests for `IntrinsicMath` have not been implemented. These are currently running via the ILGPU.Algorithms unit tests.


MoFtZ:
Abandoned. To be combined with #1148.

PR #1164, opened 2024-02-03
Half support of INumber<T>

kilngod:
This makes ILGPU.Half compatible with INumber<T>.

Aside from ILGPU.Half updates and copyright year updates the following should be noted.

1) Backend.cs IntrinsicProvider.Dispose() was removed from Backend.Dispose(). Note most tests never complete if the 
        IntrinsicProvider was disposed with the backend. Marcel is reviewing this issue.
2) RuntimeTests.cs maxNumThreads was adjusted lower to not error on my MacBook Pro
3) KernelEntryPoints.cs was updated to not throw an assert exception when testing with JetBrains Rider
4) TestBase.cs added "using" infront of "backend" instance to manage lifecycle. Again, Marcel is reviewing lifecycle of 
        IntrinsicProvider within scope of a backend.


m4rs-mt:
Hi @kilngod, thanks for contribution this amazing piece of work. As we track different features in separate PRs, can you please separate the different fixes into distinct PRs? This also helps us reviewing and addressing comment separately. Thank you!

PR #1167, opened 2024-02-15
Lifecycle of IntrinsicProvider

kilngod:
Marcel, 

I am of the opinion that Backend is not on the same lifecycle as IntrinsicProvider and the IntrinsicProvider should not be dispose of the IntrinsicProvider. This said,  I do not know the original intension of the relationship between Backend and IntrinsicProvider.

### DIFF
--- a/Src/ILGPU/Backends/Backend.cs
+++ b/Src/ILGPU/Backends/Backend.cs
@@ -877,12 +877,14 @@ namespace ILGPU.Backends
         #region IDisposable
 
         /// <summary cref="DisposeBase.Dispose(bool)"/>
-        protected override void Dispose(bool disposing)
+        protected override void Dispose(bool disposing) => base.Dispose(disposing);
+        /*
         {
             if (disposing)
                 IntrinsicProvider.Dispose();
             base.Dispose(disposing);
         }
+        */
 
         #endregion
     }
--- a/Src/ILGPU/Backends/Backend.cs
+++ b/Src/ILGPU/Backends/Backend.cs
@@ -1,6 +1,6 @@
 ﻿// ---------------------------------------------------------------------------------------
 //                                        ILGPU
-//                        Copyright (c) 2017-2023 ILGPU Project
+//                        Copyright (c) 2017-2024 ILGPU Project
 //                                    www.ilgpu.net
 //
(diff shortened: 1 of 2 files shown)


MoFtZ:
The intrinsic provider is [created by the backend](https://github.com/m4rs-mt/ILGPU/blob/3c455a48281839f0b9888673352d6068266bf815/Src/ILGPU/Backends/Backend.cs#L831). And the backend will [dispose of it ](https://github.com/m4rs-mt/ILGPU/blob/3c455a48281839f0b9888673352d6068266bf815/Src/ILGPU/Backends/Backend.cs#L883).

The backend is [created by the accelerator](https://github.com/m4rs-mt/ILGPU/blob/3c455a48281839f0b9888673352d6068266bf815/Src/ILGPU/Runtime/Cuda/CudaAccelerator.cs#L193). However, the accelerator does not appear to [dispose of it](https://github.com/m4rs-mt/ILGPU/blob/3c455a48281839f0b9888673352d6068266bf815/Src/ILGPU/Runtime/Accelerator.cs#L838).

@kilngod what issue are you seeing?

One potential problem could be that the backend and intrinsic providers are accessible by other classes. And as they make use of these instances, the accelerator may get disposed. The backend is not explicitly disposed, so it will hang around via reference counts. And when the backend gets GC-ed, the intrinsic provider will also be disposed.

kilngod:
@MoFtZ  I went down the rabit hole. @m4rs-mt  confirmed the backend should never be used in a "using" clause which was the source as invalid conclusion on my part due to "LoadKernel" is in a using clause. A quick review shows "LoadKernel" is not used anywhere else in a using clause and this using should likely be removed to be consistent.

Src/ILGPU/Backends/Backend.cs
```diff
@@ -1,6 +1,6 @@
 ﻿// ---------------------------------------------------------------------------------------
 //                                        ILGPU
-//                        Copyright (c) 2017-2023 ILGPU Project
+//                        Copyright (c) 2017-2024 ILGPU Project
```
m4rs-mt, line 3: Nit: can you please undo changes in this file?

m4rs-mt (changes requested):

PR #1170, opened 2024-03-01
Adds BrainFloat, 8-bit MiniFloat with 5/2 and 4/3

kilngod:
Marcel,

I created new INumber<T> compliant structs BFloat16, Mini52Float8, Mini43Float8, these might be useful for the velocity engine as well as interfacing future AI engines. There are some architectural questions that you need to answer, then I can re-submit. 1) I put these in the main folder alongside the Half struct, which is also updated to INumber<T>. Do we want these structs located here or in a subfolder? 2) Do we want to use my proposed names or is there an industry standard name?

In terms of performance these are the time to run the single threaded Blazor Mandelbrot in each type in seconds on an M1 Pro CPU.

double 0.48
float 0.52
Half 4.6
BFloat16 2.5
Mini43Float8 3.5
Mini52Float8 4.1

Since Half, BFloat16, Mini43Float8 and Mini52Float8 all use 32-bit floats on the CPU, the overhead of converting back and forth from to float is significant but not bad in the scheme of things. Clearly, we need to wire in platform intrinsic methods when supported to avoid the translation penalty.


Samples/BlazorSampleApp/MandelbrotExplorer/MandelbrotExtensions.cs
```diff
@@ -61,16 +64,181 @@ public static class MandelbrotExtensions
...
+
         /// <summary>
         /// Calculate the Mandelbrot set single threaded on the CPU.
         /// </summary>
         /// <param name="buffer"></param>
         /// <param name="width"></param>
         /// <param name="height"></param>
         /// <param name="max_iterations"></param>
-        public static void CalcCPUSingle(int[] buffer, int[] display, float[] view, int max_iterations)
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
m4rs-mt, line 75: Based on our latest discussions, I think we should remove these functions to avoid people trying to use newly added types for general purpose compute (see below).

Src/ILGPU.Algorithms.Tests/HistogramTests.tt
```diff
@@ -25,8 +25,8 @@ using Xunit.Abstractions;
 // disable: max_line_length
 
 <#
-var incrementTypes = AtomicNumericTypes;
-var inputTypes = NumericTypes;
+    var incrementTypes = AtomicNumericTypes;
```
m4rs-mt, line 28: Nit: can you please undo changes to non-modified lines?

Src/ILGPU.Algorithms.Tests/HistogramTests.tt
```diff
@@ -79,15 +84,36 @@ namespace ILGPU.Algorithms.Tests
...
+
+    (var inputType in inputTypes) { #>
         internal readonly struct ModuloBin<#= inputType.Name #>Operation
             : IComputeSingleBinOperation<<#= inputType.Type #>, Index1D>
         {
             public readonly Index1D ComputeHistogramBin(
                 <#= inputType.Type #> value,
                 Index1D numBins) =>
-<#      if (inputType.IsFloat) { #>
+<#
```
m4rs-mt, line 96: Same here: I think we should restrict uses of these types to specialized TensorCore operations rather than pushing for general purpose compute at the moment. Otherwise, this can have unintended side effects in terms of getting unexpected slow downs.
@MoFtZ @NullandKale

Src/ILGPU.Tests/CompareFloatOperations.tt
```diff
@@ -59,8 +59,10 @@ namespace ILGPU.Tests
 
 <# foreach (var (operationName, infix) in operationConfigurations) { #>
 <#      foreach (var type in FloatTypes) { #>
-<#         bool isFP16 = type == FloatTypes[0]; #>
-<#         var floatRanges = GetFloatRanges(type, true, true, true, false); #>
+<#         bool isFP8 = ( type == FloatTypes[0]) || (type == FloatTypes[1]); #>
```
m4rs-mt, line 62: Question: should we allow comparison operations on these types? @MoFtZ @NullandKale 
I would argue that this may lead to unexpected performance characteristics. People should probably cast explicitly into a FP16 value to be aware of the performance considerations. What do you think?

m4rs-mt:
@kilngod Thank you for your great extension proposal 👍. We really like your idea adding specialized numeric types to be mapped to TensorCore hardware. This extension is the corner stone of all upcoming changes to make used of tensor arithmetic acceleration in ILGPU 🥇!
Regarding names: I think "Bfloat16", "FP8_E4M3" and "FP8_E5M2" may be good candidates actually based on the Nvidia math specification and Intel's specific extension proposals.
@MoFtZ @NullandKale @MPSQUARK @12345swordy @lostmsu

PR #1176, opened 2024-03-19
Fixed alignment when marshalling Cuda kernel parameters.

### LINKED ISSUE #1145: Passing Int128 as kernel parameter is not working
I expected the new `Int128` data type to just work, even if not necessarily performant. However, I have found an unexpected issue.

The following kernel does not work on Cuda, and generates the wrong output:
```C#
class Program
{
    static void MyKernel(Index1D index, ArrayView<Int128> dataView, Int128 constant)
    {
        dataView[index] = index.X + constant;
    }

    static void Main()
    {
        {
            var kernel = accelerator.LoadAutoGroupedStreamKernel<
                Index1D, ArrayView<Int128>, Int128>(MyKernel);

            using var buffer = accelerator.Allocate1D<Int128>(1024);
            kernel((int)buffer.Length, buffer.View, 42);

            var data = buffer.GetAsArray1D();
            for (int i = 0, e = data.Length; i < e; ++i)
            {
                if (data[i] != 42 + i)
                    Console.WriteLine($"Error at element location {i}: {data[i]} found");
            }
        }
    }
}
```

Expected Output:
```
data[0] = { Lower = 42, Upper = 0 }
data[1] = { Lower = 43, Upper = 0 }
data[2] = { Lower = 44, Upper = 0 }
etc
```

Actual Output on Cuda:
```
data[0] = { Lower = 0, Upper = 42 }
data[1] = { Lower = 1, Upper = 42 }
data[2] = { Lower = 2, Upper = 42 }
etc
```

MoFtZ:
Fixes #1145.

When launching a Cuda kernel, ILGPU will create a runtime struct that represents all the kernel parameters. These kernel parameters are then passed to `cuLaunchKernel` as a single buffer. It is up to the caller of `cuLaunchKernel` (i.e. ILGPU) to make sure the alignment/padding of this buffer is correct.

This runtime struct is created using the alignment/padding rules of the .NET runtime. Using the example of:
`Kernel(Index1D index, ArrayView1D<Int128, Stride1D.Dense> output, Int128 constant)`

.NET would create a struct as:
- Offset 0 = Index1D (4 bytes)
- Offset 4 = Padding (4 bytes)
- Offset 8 = ArrayView (16 bytes)
- Offset 24 = Padding (24 bytes)
- Offset 48 = Int128 (16 bytes)

The additional padding at Offset 24 does not match the alignment rules of Cuda - it is expecting to find the Int128 at Offset 24.

Changing the order:
`Kernel(Index1D index, Int128 constant, ArrayView1D<Int128, Stride1D.Dense> output)`

.NET would create a struct as:
- Offset 0 = Index1D (4 bytes)
- Offset 4 = Padding (12 bytes)
- Offset 16 = Int128 (16 bytes)
- Offset 32 = ArrayView (16 bytes)

Again, the additional padding at Offset 4 does not match the alignment rules of Cuda - it is expecting to find the Int128 at Offset 8.

---------------
Attempt 1
- ~~This PR modifies how ILGPU provides the kernel parameters to Cuda. Instead of a single buffer that needs to be manually aligned, we now use an array of pointers, to the kernel parameters.~~

~~Cuda will handle the alignment/padding for us.
ILGPU continues to use the single buffer as the placeholder/container of all the kernel parameters.~~

Attempt 2
- ~~Changed the PTX Argument Mapper to manually align the kernel parameter fields.~~

Attempt 3
Looks like the issue is specific to `Int128`. It is not treated as a regularly struct. The .NET Runtime considers it an intrinsic type, and aligns it to 16 bytes. Modified ILGPU to pre-register `Int128`, and force a 16 byte alignment.


MoFtZ:
@m4rs-mt just a thought... this works for the top-level structure. what about the nested structures? I dont think it will work there.

MoFtZ:
@m4rs-mt OK, this is definitely a problem.

Placing `Int128` inside another struct, and passing `MyStruct` as a kernel parameter will also cause alignment/padding issues.
```
public struct MyStruct
{
    public byte X;
    public Int128 Y;
}
```
`Y` is aligned at Offset 16 in .NET, but Cuda expects it at Offset 8.

MoFtZ:
@m4rs-mt found another issue - if the kernel parameter is a structure containing a nested structure, ILGPU will flatten the all the fields. This process of flattening can potentially change the byte offset of fields, which means that we can no longer use the .NET representation of the variable when using `fixed`.

PR #1182, opened 2024-03-28
Fixed OpenCL printf newline output.

MoFtZ:
The `Interop.WriteLine` output on Windows OpenCL was always less than desirable - showing "\r\n" in the output instead of a newline. We suspected this was an OpenCL driver issue.

image

This PR fixes the output by addressing two issues.

The first issue is that escaping `\` was the last step, which incorrectly double-escaped all previous escapes.

The second issue is that `\r\n` is a Windows convention, but C-printf normally uses just `\n`. The underlying C-printf implementation then converts `\n` to `\r\n` on Windows devices. To handle this, we use `Environment.NewLine` to deal with whether we are running on Windows or Unix. Without converting `\r\n` as a single translation, my OpenCL implementation would translate `\r` to "r", and then `\n` to a newline.

The fixed output now looks correct:
image

### DIFF
--- a/Src/ILGPU/IR/Values/IOValues.cs
+++ b/Src/ILGPU/IR/Values/IOValues.cs
@@ -310,11 +310,16 @@ namespace ILGPU.IR.Values
         /// </summary>
         public string ToEscapedPrintFExpression() =>
             ToPrintFExpression()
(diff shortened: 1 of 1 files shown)


m4rs-mt (approved):
Nice catch 👍

PR #1185, opened 2024-04-04
Moved OpenCL IntrinsicMath implementations.

MoFtZ:
Extracted from #1148.

Currently, the Math functionality of OpenCL is missing an implementation for `Rcp` and `Log(x,y)`. To add the missing functionality, it is necessary to call `EnableAlgorithms`, provided by ILGPU.Algorithms.

Moved the functionality into ILGPU, so that it is no longer necessary to call `EnableAlgorithms`.


m4rs-mt (approved):
Nice catch and great extension!

PR #1199, opened 2024-04-19
Fixed marshalling ArrayView.Empty as kernel parameter.

MoFtZ:
Backported #1131 for v1.5.2 milestone.


m4rs-mt (approved):

m4rs-mt (changes requested):
It looks like a kernel attribute is missing on some .Net test to make it work with our .Net Framework build pipeline.

MoFtZ:
> It looks like a kernel attribute is missing on some .Net test to make it work with our .Net Framework build pipeline.

Added `[MethodImpl(MethodImplOptions.NoInlining)]` because `KernelMethodAttribute` was being optimized out.

PR #1214, opened 2024-04-26
[NativeAOT] Implement IRExporter API

IsaMorphic:
As the first part of the long deliberated NativeAOT feature implementation, I have created the following set of internal APIs that ILGPU consumes in order to provide a uniform, flattened representation of the IR graph to the end user:

- `ILGPU.IR.IRExporter`
- `ILGPU.IR.IRContainer`

Additionally, the following _public_ APIs are exposed to end users of the ILGPU compiler for this purpose:

- `ILGPU.IR.IRContainer`
- `ILGPU.IR.IRValue`
- `ILGPU.IR.IRType`
- `ILGPU.Backends.IR.IRExporterHook`

The intended use-case for this API is for the end user to inject `IRExporterHook` into their desired backend via the provided DI mechanism, and from that extract `IRExporterHook.CurrentContext.ExportContainer` once the initial code generation / optimization phase is completed. 

Runtime recovery of an IRContext from this exported representation will be written in the next PR (link pending!)

TODOs:
- [x] In-source documentation
- [x] Review by code owner


Src/ILGPU/IR/IRType.cs
```diff
@@ -0,0 +1,66 @@
...
+{
+    /// <summary>
+    /// A uniform value type representing an exported <see cref="Types.TypeNode"/> from an <see cref="IRContext"/>.
+    /// </summary>
+    /// <param name="Id">Corresponds to <see cref="Node.Id"/></param>
+    /// <param name="Class">A <see cref="Classifier"/> value denoting which kind of <see cref="Types.TypeNode"/> this instance represents</param>
+    /// <param name="Nodes">A list of <see cref="NodeId"/> values corresponding to those of related instances; important for later reconstruction</param>
+    /// <param name="BasicValueType">Corresponds to <see cref="Types.TypeNode.BasicValueType"/></param>
+    /// <param name="Data">Extra data specific to this type's kind or instance</param>
+    public record struct IRType(
```
En3Tho, line 13: In runtime repo people avoid records because records tend to bloat resulting binaries. Not sure if this is a concern here though. Just mentioning.

Src/ILGPU/IR/Construction/BasicBlock.Builder.cs
```diff
@@ -511,13 +511,29 @@ public void ReplaceWithCall(Value value, Method implementationMethod)
             protected override T CreateTerminator<T>(T node)
             {
                 Terminator?.Replace(node);
-                return (Terminator = node).AsNotNullCast<T>();
+
+                Terminator = node;
+
+                switch (node.GetType().FullName)
```
En3Tho, line 517: Is type of this concrete value important here or typeof(T) in general? If second, then I think it's better to use typeof(T)

IsaMorphic: `System.Type` can't be used in `switch(X)`, so this would be the reason why. I agree that what you're suggesting would be more semantic, however, if it were allowed by the compiler.

Src/ILGPU/IR/IRType.cs
IsaMorphic: In this case I'm purposefully using `record` so that it auto-implements `IEquatable` and all that.

Src/ILGPU/IR/Construction/BasicBlock.Builder.cs
En3Tho: `typeof(T).FullName` can be used in a similar way too. 

It's just that node's type can be different from that of `typeof(T)` e.g.
```csharp
interface QQ {}
class WW : QQ {}

static class Test
{
    public static bool Z<T>(T value)
    {
        return typeof(T) == value.GetType();
    }
}

QQ ww = new WW();
Test.Z(ww) // false
```

This might be not a problem at all though

IsaMorphic: I do in fact need to use `object.GetType()` in order for this type check to work because `T` will always be `IR.Values.TerminatorValue` for the erroneous calls in question.

IsaMorphic:
I believe this should all build now, I even went in and substituted the C# 12.0 specific "Collection initializers" feature for proper array initializers so that everything backports OK to `net6.0`. I will also be opening a draft PR containing the `IRImporter` class.

PR #1221, opened 2024-05-10
Support for AI floats BF16, FP8E4M3, FP8E5M2

kilngod:
Added classes to support and emulate AI data types found in 2023 and 2024 hardware.


hez2010:
As for bfloat16, better wait a bit for https://github.com/dotnet/runtime/pull/98643.
While I asked the dotnet team and they showed no interest in fp8 in near future, so I the fp8e4m3 and fp8e5m2 can continue to use the custom types in this PR.

PR #1225, opened 2024-05-17
[NativeAOT] Implement IRImporter API

IsaMorphic:
This is a continuation of #1214. Changes include adding the `IRImporter` class (consumed internally by `IRContext.Import(IRContainer.Exported)`) and modifying the `IRExporter` and `IRContainer` classes to integrate properly with the entire export/import pipeline. 

A potential memory leak issue has also been brought to light by the relative abuses of the ILGPU API delegated by this feature. I will investigate and open an issue if it does indeed come to fruition. 

I will rebase this branch on top of `master` once #1214 gets merged.


Src/ILGPU/IR/IImportable.cs
```diff
@@ -0,0 +1,30 @@
...
+
+namespace ILGPU.IR
+{
+    /// <summary>
+    /// Describes an entity that can be imported from a specific other type.
+    /// </summary>
+    /// <typeparam name="TFrom">
+    /// The type that this instance can import from.
+    /// </typeparam>
+    public interface IImportable<TFrom>
```
m4rs-mt, line 20: Info: this interface may no longer be required in the future.

Src/ILGPU/IR/IExportable.cs
```diff
@@ -0,0 +1,30 @@
...
+
+namespace ILGPU.IR
+{
+    /// <summary>
+    /// Describes an entity that can be exported to a specific other type.
+    /// </summary>
+    /// <typeparam name="TTo">
+    /// The type that this instance can export to. 
+    /// </typeparam>
+    public interface IExportable<TTo>
```
m4rs-mt, line 20: Info: this interface may no longer be required in the future.

m4rs-mt (commented):
Hey @IsaMorphic, thank you for raising this draft PR. As we discussed on Discord, I left some comments here to give you feedback on your early work. Some general remarks:

- This PR seems trying to add a new optimization level called `AOT`. This should not be required for realizing AOT compilation, as optimization levels will remain the same.
- The serialization approach being used creates an unsafe version of our IR first and then performing the serialization of this newly created IR graph. A direct approach should be much more elegant and considerably easier to maintain. Please see my inline comments for more information.
- Serializing `NodeId` values may not be lowered to raw `long` values to maintain type safety.

Src/ILGPU/IR/Construction/Memory.cs
```diff
@@ -139,8 +139,11 @@ partial class IRBuilder
             Value value)
         {
             location.Assert(
-                target.Type is PointerType pointerType &&
-                pointerType.ElementType == value.Type);
+                target.Type is AddressSpaceType targetAddrType &&
+                targetAddrType.ElementType == value.Type | (
+                value.Type is AddressSpaceType valueAddrType &&
```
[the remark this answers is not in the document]
IsaMorphic: Yes. After preliminary optimization steps on the IR graph, we can't always assume that parameter `value` will consist only of non-address types. For example, what is initially built from IL as a guaranteed Store(Pointer Expression <- Value Expression) might yet be optimized/raised to Store(View Element Addr <- Implicitly Indirect Reference to Value) or the like. 

In this event, we should still be able to reconstruct the IR using these semantics, even though the original use case was not designed with this intention. Hence why the assertion should be updated.

IsaMorphic:
Going to close this and open a new PR starting from scratch with this feature with the `IRSerializer` approach in mind. 
Importantly I will backlink to this PR for posterity.

PR #1228, opened 2024-05-21
[NativeAOT] Implement `IIRWriter` API

IsaMorphic:
Based on the research and precedent consisting of #1225 and #1214, I am starting from scratch to implement a bitstream writer for the IR graph.

PR #1232, opened 2024-05-25
[NativeAOT] Implementation Proposal for AOT-friendly `IIRReader` API

IsaMorphic:
Based on #1228, this is a PROPOSAL for how one could implement a reflection-free deserialization API for ILGPU, designed to mirror/complement its sister `IIRWriter` API.


Src/ILGPU/IR/Values/Arithmetic.cs
IsaMorphic, line 1: In this diff we see an example of how this API gets implemented in action. 
The `IIRReader` instance provides enough information via its `Context` property to call the appropriate `IRBuilder` API method for each `ValueKind`.

Src/ILGPU/IR/Values/ValueKind.cs
```diff
@@ -392,8 +394,21 @@ public ValueKindAttribute(ValueKind kind)
     /// <summary>
     /// Utility methods for <see cref="ValueKind"/> enumeration values.
     /// </summary>
-    public static class ValueKinds
+    public static partial class ValueKinds
     {
+        // TODO: Source generator writes static type
+        // constructor to initialize this value, reflection free!
+        private readonly static Dictionary<ValueKind,
```
IsaMorphic, line 401: As per the comment lines directly above, ILGPU would implement its own internal source generator to initialize a call table which maps `ValueKind` keys to their corresponding `GenericValueReader` delegates, those ofc being that which are implemented via `IValueReader` for each concrete `Value` type!

Issue #1209, opened 2024-04-19
[BUG]: CopyToCpu is using refs in unsafe way but there is no indication of that.


En3Tho:
### Describe the bug

This routine calls `CopyToCPUUnsafeAsync` which does not pin/fix refs and simply calls Unsafe.AsPointer on them. This could potentially lead to situations where pointer could end up pointing at invalid memory region if GC kicks in while copy operation is ongoing.

I guess there is an implicit assumtion that all refs should be pinned before use. But I personally would have expected ref taking api to pin those manually, e.g. `fixed p = ref ...`

I think there are few ways to deal with that: 
1. Simply tell people that there are NO ref safety guarantees so it becomes a global implicit rule
2. Explicitly state this in the comment
3. Find such apis and actually use `fixed` then route to pointer-taking apis
4. (I guess not possible) rename api to CopyToCPUUnsafe

```csharp
public static void CopyToCPU<T, TView>(
      this TView source,
      AcceleratorStream stream,
      ref T cpuData,
      long length)
      where T : unmanaged
      where TView : IContiguousArrayView<T>
    {
      source.CopyToCPUUnsafeAsync<T, TView>(stream, ref cpuData, length);
      stream.Synchronize();
    }
```

### Environment

- ILGPU version: main

### Steps to reproduce

In the description

### Expected behavior

In the description

m4rs-mt:
Hi @En3Tho, thank you for raising this potential bug report and welcome to the ILGPU community! Indeed, these functions are considering access to either pinned managed/or even unmanaged pointers to be targeted directly from GPU drivers at the moment. I agree that we should improve documentation of these functions to help people understand potential side effects (as in undefined behavior...) and potentially enhance this function with `fixed p = ...` while providing a native overload using `IntPtr` or `void*` to avoid additional pinning attempts. What do you think?

En3Tho:
I think it would be good to have different `ref T` and `T*` overloads, former being "safer oner". I understand that basically once you go into gpu land there is no such thing as "safety" but still it would be good to eliminate a possible cause of clr crash. 

My main concern is that someone might try using this on web server for example where some tasks are a perfect fit for gpu. Active memory hogs like webservers are triggering gc quite often and it could really lead to the whole app crashing. Usually using `ref T` means that operation is guaranteed to be `safe` at least memory wise.

[...]

PR #1296, opened 2024-10-30
Remove some Gc holes

benaadams:
Going via Unsafe.AsPointer risks creating Gc holes where the GC has moved the memory but the pointer has not been updates

Use safer alternatives via `ref`

### DIFF
--- a/Src/ILGPU/Runtime/CPU/CPUMemoryBuffer.cs
+++ b/Src/ILGPU/Runtime/CPU/CPUMemoryBuffer.cs
@@ -72,12 +72,14 @@ namespace ILGPU.Runtime.CPU
             ref byte sourcePtr,
             ref byte targetPtr,
             long sourceLengthInBytes,
-            long targetLengthInBytes) =>
-            Buffer.MemoryCopy(
-                Unsafe.AsPointer(ref sourcePtr),
-                Unsafe.AsPointer(ref targetPtr),
-                sourceLengthInBytes,
-                targetLengthInBytes);
+            long targetLengthInBytes)
+        {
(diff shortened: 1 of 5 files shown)


Src/ILGPU/Runtime/CPU/CPUMemoryBuffer.cs
```diff
@@ -72,12 +72,14 @@ public static unsafe void CPUCopyToCPU(
...
-                Unsafe.AsPointer(ref targetPtr),
-                sourceLengthInBytes,
-                targetLengthInBytes);
+            long targetLengthInBytes)
+        {
+            ArgumentOutOfRangeException.ThrowIfLessThan(targetLengthInBytes, sourceLengthInBytes);
+            Unsafe.CopyBlock(
+                ref targetPtr,
+                ref sourcePtr,
+                (uint)sourceLengthInBytes);
```
neon-sunset, line 81: Cast to `uint` here will truncate the chunk of memory being copied to 4GiB. Instead, this could be done with pinning both byrefs and calling `Buffer.MemoryCopy` which does the right thing. In any case, happy to see more people discovering an amazing project that ILGPU is.

En3Tho:
Related: https://github.com/m4rs-mt/ILGPU/issues/1209

m4rs-mt (changes requested):
Thanks for submitting this PR and addressing this issue. Before we merge it, could you also take a look at resolving the build issues we're seeing? Once those are addressed, this PR will be ready to move forward.

En3Tho:
@m4rs-mt Some build failures are related to .Net 6 runtime not having support for `ref field`. .Net 6 is out of support currently. Do you think ILGpu should support it moving forward?

PR #1310, opened 2024-12-09
Fixed disassembly of Conv_R_Un opcode.

### LINKED ISSUE #1309: [BUG]: <title>Type cast uint to float
### Describe the bug

When cast uint variable to float generated IL code:
```
  IL_0020:  conv.r.un
  IL_0021:  conv.r4
``` 
Then ILGPU comliller interprit `conv.r.un` as cast to float64 (double). And if host (OpenCL) not support double then it's lead to Exception and not work.

### Environment

- ILGPU version: [e.g., 1.5.1]
- .NET version: [e.g., .NET 8]
- Operating system: [e.g., Windows 10]
- Hardware (if GPU-related): [e.g., NVIDIA GeForce GTX 1080]

### Steps to reproduce

```uint a = ..
...
float b = a * 6.18;
...
```

### Expected behavior

I think we should consider these two lines as one atomic command

### Additional context

_No response_

MoFtZ:
Fixes #1309.

Looking at the documentation, it should be a `float` operation rather than `double`.

However @m4rs-mt, I do not understand the last section of the implementation notes. Do we need to do anything special to handle the `float64` case?

`If overflow occurs converting a floating-point type to an integer the result returned is unspecified. The conv.r.un operation takes an integer off the stack, interprets it as unsigned, and replaces it with a floating-point number to represent the integer: either a float32, if this is wide enough to represent the integer without loss of precision, or else a float64.`

Source: https://learn.microsoft.com/en-us/dotnet/api/system.reflection.emit.opcodes.conv_r_un?view=net-8.0

### DIFF
--- a/Src/ILGPU/Frontend/DisassemblerDriver.cs
+++ b/Src/ILGPU/Frontend/DisassemblerDriver.cs
@@ -1,6 +1,6 @@
 ﻿// ---------------------------------------------------------------------------------------
 //                                        ILGPU
-//                        Copyright (c) 2018-2021 ILGPU Project
+//                        Copyright (c) 2018-2024 ILGPU Project
 //                                    www.ilgpu.net
 //
 // File: DisassemblerDriver.cs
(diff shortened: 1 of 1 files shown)


m4rs-mt:
Nice catch, thank you very much for looking into this. I think the documentation may be interpreted as:
`if IsInt(top of stack) then castFP32(top of stack) else castFP64(top of stack)`

m4rs-mt:
Superseded by #1356.

PR #1319, opened 2025-02-06
Bump Cuda versions.

MoFtZ:
Backported #1314 for v1.5.2 milestone.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -99,7 +99,7 @@ jobs:
       - name: Setup OS matrix
         id: setup-os-matrix
         run: |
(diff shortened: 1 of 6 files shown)


MoFtZ:
@m4rs-mt
`ubuntu-latest` is now using v24.04, which is not compatible with `netcoreapp3.1` and `net5.0` - looks like some dependencies are no longer included. Have downgraded for this branch.

Also needed to update the upload and download actions, as the version we are using is no longer supported:
https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/

m4rs-mt (approved):

PR #1324, opened 2025-03-23
Enable default interface members

HyperSphereStudio:
Added in ability to resolve default interface members.

For instance, the following would not compile:

```csharp
public interface TestInterface {
    public int Min { get; }
    public int Max { get; }
    public int Center => (Min + Max)/2;
}

public record struct TestStruct(int Min, int Max) : TestInterface {
}
```

because "get_Center"'s definition is in TestInterface. When ILGPU attempts to resolve the underlying datatype, it returns with TestInterface.

With .NET's latest constraint type for struct members this type needs to be forwarded from the call site down to every reference of "this" in the instance definition. To do this, I created a "SpecializedMethod" object that forwards the constrained type.

Nested example where forwarded is required (If get_NestedCenterCall is invoked with constrained type, this type has to be forwarded to get_Center):

```csharp

 public interface ITestInterface {
            public float Min { get; }
            public float Max { get; }
            public float Center => (Min + Max)/2;
            public float NestedCenterCall {
                get => Center*4;
            }
        }
```

Essentially, this PR replaces most usages of MethodBase to SpecializedMethod containing the specialized call type instance and the MethodBase.

### DIFF
--- a/Src/ILGPU.Tests/DisassemblerTests.cs
+++ b/Src/ILGPU.Tests/DisassemblerTests.cs
@@ -107,7 +107,7 @@ namespace ILGPU.Tests
                 .ThrowIfNull();
             var method = methodInfo.MakeGenericMethod(genericType);
 
-            var disassembler = new Disassembler(method, SequencePointEnumerator.Empty);
+            var disassembler = new Disassembler(new(null, method), SequencePointEnumerator.Empty);
             disassembler.Disassemble();
         }
     }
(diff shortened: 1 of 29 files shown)


m4rs-mt:
Welcome @HyperSphereStudio to the ILGPU community. Thank you very much for your contribution and adding this feature to ILGPU. We will review your PR shortly.

HyperSphereStudio:
Hmm Ill take a look into why its not compiling, half of it seems to not like the nullable types

m4rs-mt:
Looks like we are running into compilation issues which were addressed on branch 1.5.X. Considering our current plans
for version 2.0, I recommend changing the target branch of this PR to 1.5.X to include the feature in the upcoming
1.5.3 release.

